PR #132, opened 2021-01-22
Incorporate NREL EFS Flexibility Data

lanesmith218:
### Purpose
The added modules allow the flexibility data sets from NREL's Electrification Futures Study (EFS) to be downloaded and reformatted for use in the grid model. The flexibility data reveals the amount of base demand data that can be shifted for each time step and each state. This data (and likely data of a similar kind/format from our other external collaborators) will be used to help set bounds on the curtailment amount (via constraints) in the forthcoming demand flexibility optimization formulation.

### What the code is doing
The following functions within `get_efs_data.py` are created to execute this purpose:
* The `download_flexibility_data()` function is responsible for downloading and extracting the flexibility data. The flexibility data is provided slightly differently from the base demand data, which was the focus of the previous PR. All of the flexibility data is provided in three different .csv files (each of which is contained within its own .zip file), whereas the base demand data was provided in nine different .csv files. The difference is that the flexibility data is only separated by electrification scenario, meaning that each .csv file contains flexibility data for each state, time step, technology advancement, sector (not down to the subsector, as was seen with the base demand data), and flexibility scenario. As was seen in the last PR with the base demand data, the .zip files use a compression style (Deflate64) that is unsupported by the popular Python package `zipfile`, necessitating the use of the Command Line or Terminal to extract the .csv file.
* The `partition_flexibility_by_sector()` function filters the flexibility data for a specific year, technology advancement, and flexibility scenario and splits the flexibility by sector. `partition_flexibility_by_sector()` is also responsible for formatting the flexibility data to be similar to our current demand data formatting (time steps x states).

These modules integrate well with the other EFS-related modules from the last PR. Currently, I don't anticipate aggregating the different sectoral flexibility like I did with the sectoral demand. I anticipate different sectoral flexibility possessing different operating characteristics (e.g., electric vehicles may be able to shift in different directions for different durations than a hot water heater), though I do need to review how NREL baked operating characteristics into their flexibility data sets. All of that being said, `partition_flexibility_by_sector()` produces an output with the same structure as `partition_demand_by_sector()`, indicating that it would work similarly with `combine_efs_demand()` (located in `aggregate_demand.py`) if we decide to aggregate flexibility. The separate sectoral flexibility data also works with `decompose_demand_profile_by_state_to_loadzone()` (located in `map_states.py`), which is responsible for mapping state-level data to the different load zones. The only difference is that sectoral flexibility must be passed into `decompose_demand_profile_by_state_to_loadzone()` one sector at a time (i.e., using a loop); this is contrasted against the process for the base demand data, which is aggregated into one profile, that can be passed into `decompose_demand_profile_by_state_to_loadzone()` in one go. For an example of how this might work, please see the demo notebook (`efs_base_flexibility_reference_slow_2030_demo.ipynb`) or the `README`.

I will also note that `download_flexibility_data()` and `partition_flexibility_by_sector()` are similar in nature to `download_demand_data()` and `partition_demand_by_sector()`, respectively. I ended up creating separate functions (as opposed to combining the functionalities in one function) due to the differing structures of the flexibility and base demand data sets. If desired, I could combine the similar functions into one function.

### Testing
All tests pass. I created tests for both of the functions that I created in `get_efs_data.py`.  I have also created an additional demo notebook to demonstrate the functionality of these new modules (`efs_base_flexibility_reference_slow_2030_demo.ipynb`).

### Where to look
This code lives in the `nrel_efs` folder within `PreREISE/prereise/gather/demanddata/`. The two new modules are located in `PreREISE/prereise/gather/demanddata/nrel_efs/get_efs_data.py`.

The accompanying tests are located in `PreREISE/prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py`.

The demo notebook is located in `PreREISE/prereise/gather/demanddata/nrel_efs/demo/efs_base_flexibility_reference_slow_2030_demo.ipynb`.

I also updated the following documentation:

`PreREISE/README.md`
`PreREISE/ATTRIBUTION.md`

### Usage Example
For an example of how the code can be used, please refer to the specified demo notebook (`efs_base_flexibility_reference_slow_2030_demo.ipynb`) or the `README`.

### Time estimate
Since this code is similar to my recent PR, this should hopefully be a pretty quick PR to review.


README.md
```diff
@@ -559,33 +601,52 @@ a particular load zone. This mapping can be accomplished as follows:
...
+
+```python
+from prereise.gather.demanddata.nrel_efs.get_efs_data import partition_flexibility_by_sector
+from prereise.gather.demanddata.nrel_efs.map_states import decompose_demand_profile_by_state_to_loadzone
+
+sect_flex = partition_flexibility_by_sector(es="Reference", ta="Slow", flex="Base", year=2030, fpath="")
+sect_flex_lz = {}
+for i in sect_flex.keys():
+    sect_flex_lz[i] = decompose_demand_profile_by_state_to_loadzone(df=sect_flex[i])
 ```
```
rouille, line 650:
```
from prereise.gather.demanddata.nrel_efs.get_efs_data import partition_flexibility_by_sector
from prereise.gather.demanddata.nrel_efs.map_states import decompose_demand_profile_by_state_to_loadzone

sect_flex = partition_flexibility_by_sector(es="Reference", ta="Slow", flex="Base", year=2030, fpath="")
sect_flex_lz = {k: decompose_demand_profile_by_state_to_loadzone(df=v) for k, v in sect_flex.items()}
```

README.md
```diff
@@ -559,33 +601,52 @@ a particular load zone. This mapping can be accomplished as follows:
...
 
 [RAP]: https://www.ncdc.noaa.gov/data-access/model-data/model-datasets/rapid-refresh-rap
```
rouille, line 652: Whenever you quote a function in the README, I would omit the parentheses, e.g., `foo()` returns --> `foo` returns

README.md
lanesmith218: Thanks for the suggestion! I'll make that change.

README.md
lanesmith218: Thank you for letting me know. I'll fix that.

README.md
lanesmith218: Done.

README.md
lanesmith218: Done.

rouille:
I notice a lot of similarity between the `download_demand_data` and `download_flexibility_data` functions. First, they share the same `es` and `fpath` parameters and hence we could have a couple of private functions (`_check_electrification_scenarios` and `_check_path`) that check that what is passed is valid. Second, the download is almost identical and consequently it could be factored out too with a `_download_data` that would take as arguments the url and other things that differ between demand and flexibility.

lanesmith218:
> I notice a lot of similarity between the `download_demand_data` and `download_flexibility_data` functions. First, they share the same `es` and `fpath` parameters and hence we could have a couple of private functions (`_check_electrification_scenarios` and `_check_path`) that check that what is passed is valid. Second, the download is almost identical and consequently it could be factored out too with a `_download_data` that would take as arguments the url and other things that differ between demand and flexibility.

@rouille great, thank you for that feedback! I knew there were some redundancies, but I didn't know how best to clean that up. Is it better to create the private functions (e.g., `_download_data`, `_check_electrification_scenarios`) than to combine `download_demand_data` and `download_flexibility_data` into one function (e.g., one `download_data`)?

lanesmith218:
> > > I notice a lot of similarity between the `download_demand_data` and `download_flexibility_data` functions. First, they share the same `es` and `fpath` parameters and hence we could have a couple of private functions (`_check_electrification_scenarios` and `_check_path`) that check that what is passed is valid. Second, the download is almost identical and consequently it could be factored out too with a `_download_data` that would take as arguments the url and other things that differ between demand and flexibility.
> > 
> > 
> > @rouille great, thank you for that feedback! I knew there were some redundancies, but I didn't know how best to clean that up. Is it better to create the private functions (e.g., `_download_data`, `_check_electrification_scenarios`) than to combine `download_demand_data` and `download_flexibility_data` into one function (e.g., one `download_data`)?
> 
> I tend to prefer smaller functions. Those can be easily tested. For example the _check function can easily be tested because these would be de-correlated from the download that requires an ssh connection. The two main functions (`download_demand_data` and `download_flexibility_data`) will be very slim in this scenario.

Great, that all makes sense! I'll start implementing this for the different functions.

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -9,33 +9,33 @@
...
+    if not isinstance(df, pd.DataFrame):
+        raise TypeError("df must be input as a pandas.DataFrame.")
 
-    # Check the aggregate demand DataFrame time stamps and column headers
-    if not agg_dem.index.equals(
+    # Check the demand DataFrame time stamps and column headers
+    if not df.index.equals(
         pd.date_range("2016-01-01", "2017-01-01", freq="H", closed="left")
     ):
         raise ValueError("This data does not have the proper time stamps.")
```
rouille, line 37: We usually use `timestamp`, i.e., one word

lanesmith218: Okay, I'll fix that

lanesmith218: Done.

lanesmith218:
@rouille I added the private functions as we discussed earlier. However, I haven't created extra tests yet. Should I create a test for every function or does the presence of the private functions mean that some functions don't need to be tested? For instance, with both `download_demand_data` and `download_flexibility_data` relying on `_check_electrification_scenarios_for_download`, `_check_path`, `_download_data`, and `_extract_data`, should I create tests for each function, only `download_demand_data` and `download_flexibility_data` (since they encompass the private functions), or only the private functions (since they essentially make up `download_demand_data` and `download_flexibility_data`)?

rouille:
> @rouille I added the private functions as we discussed earlier. However, I haven't created extra tests yet. Should I create a test for every function or does the presence of the private functions mean that some functions don't need to be tested? For instance, with both `download_demand_data` and `download_flexibility_data` relying on `_check_electrification_scenarios_for_download`, `_check_path`, `_download_data`, and `_extract_data`, should I create tests for each function, only `download_demand_data` and `download_flexibility_data` (since they encompass the private functions), or only the private functions (since they essentially make up `download_demand_data` and `download_flexibility_data`)?

No need to refactor the tests since, as you pointed out, you implicitly test the private functions when you test `download_demand_data` and `download_flexibility_data`. I will take a look tomorrow at the changes you made. Thanks for the extra work.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
```diff
@@ -36,129 +33,273 @@ def download_data(es=None, ta=None, fpath=""):
...
+    ta = _check_technology_advancements_for_download(ta)
+    fpath = _check_path(fpath)
+
+    # Download each of the specified load profiles
+    z = {}
+    for i in es:
+        z[i] = {}
+        for j in ta:
+            # Assign path and file names
+            zip_name = "EFSLoadProfile_" + i + "_" + j + ".zip"
```
danielolsen, line 47: We are only testing against Python 3.7+, so we can make use of f-strings to simplify the construction of these filenames (here and elsewhere). E.g.: `"EFSLoadProfile_" + i + "_" + j + ".zip"` -> `f"EFSLoadProfile_{i}_{j}.zip"`.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
```diff
@@ -36,129 +33,273 @@ def download_data(es=None, ta=None, fpath=""):
...
+    :param str fpath: The input file path.
+    :return: (*zipfile.ZipFile*) -- The .zip file stored in memory for attempted
+        extraction using Python's zipfile module.
+    """
+
+    # Save a local copy of the .zip file for extraction
+    r = requests.get(url, stream=True)
+    if r.status_code != requests.codes.ok:
+        r.raise_for_status()
+    open(zip_name, "wb").write(r.content)
```
danielolsen, line 209: I think it is generally cleaner to do `with open(...) as f: f.write(...)`

prereise/gather/demanddata/nrel_efs/get_efs_data.py
```diff
@@ -36,129 +33,273 @@ def download_data(es=None, ta=None, fpath=""):
...
-                    z[i][j].extractall(fpath)
-                    print(csv_name + " successfully extracted!")
-                else:
-                    # Bypass the zipfile module if it does not work on the first file
-                    raise NotImplementedError
-            except NotImplementedError:
-                if zf_works:
+                # Windows Command Line does not support this type of compression
+                # Try using 7zip, if it is installed in the default location
+                sz_path = "C:/Program Files/7-Zip/7z.exe"
```
danielolsen, line 263: Should the user be able to specify a path to an alternate location at which they've installed 7zip?

lanesmith218: Yeah I can definitely add that. I think I originally left it out because I wasn't sure how often people install it outside of the default location and if it was worth it to include an additional input. I can make it so the current location is the default input for an additional `sz_path` input.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
```diff
@@ -288,6 +392,259 @@ def partition_by_sector(es, ta, year, sect=None, fpath="", save=False):
...
+                + "_"
+                + flex
+                + "_Flexibility_"
+                + es
+                + "_"
+                + ta
+                + "_"
+                + str(year)
+                + ".csv"
+            )
```
danielolsen, line 502: `new_csv_name = f"{i}_{flex}_Flexibility_{es}_{ta}_{year}.csv"`

lanesmith218: Yeah, this is much cleaner! I just finished updating all file names to use the f-strings, as suggested above.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
lanesmith218: I can certainly change `open(zip_name, "wb").write(r.content)` if that is preferred. However, I don't really understand how this is cleaner. @danielolsen, can you explain?

BainanXia: @lanesmith218 Using `with` statement will automatically close the file and clean the cache after interactions. It is a fancier way to do `try: open...finally: close` provided by Python3.

lanesmith218: Thanks for clarifying! I'll make that change.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
```diff
@@ -288,6 +392,259 @@ def partition_by_sector(es, ta, year, sect=None, fpath="", save=False):
...
+    sect_flex = {
+        i: df[df["Sector"] == i]
+        .drop(columns=["Sector"])
+        .groupby(["LocalHourID", "State"])
+        .sum()
+        .unstack()
+        for i in sect
+    }
+    sect_flex = {
+        i: sect_flex[i].set_axis(sorted(set(abv2state) - {"AK", "HI"}), axis="columns")
```
BainanXia, line 471: Just to confirm the original order of the columns here matchers the new label order, i.e. `sorted(set(abv2state_- {"AK", "HI"})`?

lanesmith218: Yes. The reason we need to set the columns is because `unstack` creates double-valued columns (i.e., each column is a pair of "LoadMW" and the state). However, the pairs are already ordered according to the sorted states (if I remember correctly, I think `groupby` orders the columns by default). So, setting the columns in effect eliminates the additional "LoadMW" label on every column.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
danielolsen: In general, classes can have `__enter__` and `__exit__` methods, and the `with` syntax ensures that `__enter__` is done before anything in the block, and `__exit__` is done as a `finally` step (i.e. always executed). For `open` specifically, this ensures that the file is closed afterwards.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
BainanXia: I see. Then that's okay. I'm concerned about this simply because `set_axis` only change the column labels instead of moving corresponding data and if it happens to be different order, the table is messed up.

BainanXia:
I've gone through the code. It is consistent with what @lanesmith218 and I have discussed before this PR is up and looks good to me. Once @rouille finishes the final pass, I think this is good to go. Great work, @lanesmith218.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
lanesmith218: Yeah, I agree. I was trying to find another way to get rid of the paired "LoadMW" labels, but this was the only way I could figure out. Since `groupby` orders the columns by default, we should be okay though.

lanesmith218:
> I've gone through the code. It is consistent with what @lanesmith218 and I have discussed before this PR is up and looks good to me. Once @rouille finishes the final pass, I think this is good to go. Great work, @lanesmith218.

Thank you!

prereise/gather/demanddata/nrel_efs/get_efs_data.py
danielolsen: Are the labels a tuple, or a MultiIndex? Either way, we should be able to set them based on what's actually there, not what we expect. If it's a tuple, we can do a list comprehension; if it's a MultiIndex, we can use the `get_level_values` method.

lanesmith218: They are a MultiIndex. I can look into the `get_level_values` method that you suggest

BainanXia: If after groupby, it gives something like`LoadMW_CA`, maybe doing following could be safer: `sect_flex[i].rename(lambda x:x[-2:], axis="columns", inplace=True)`

lanesmith218: After `groupby`, each column is essentially `('LoadMW', '[state]')`. It looks like `get_level_values` can be implemented easily by doing `sect_flex[i].columns.get_level_values(1)`

lanesmith218: Unless I'm missing something, it seems that I still need to use `set_axis` to keep this as a "one-line" piece of code: 

`sect_flex = {i: sect_flex[i].set_axis(sect_flex[i].columns.get_level_values("State"), axis="columns") for i in sect}`

However, this is safer than what is currently there because the use of `get_level_values` should generate column headers that are consistent with the order of the data.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/get_efs_data.py
lanesmith218: Done.

rouille (approved):
It looks great. Thanks.

lanesmith218:
> It looks great. Thanks.

Great, thanks @rouille!

PR #135, opened 2021-02-10
Fixes for gathering wind data

jon-hagg:
### Purpose
Update the subdomain where data is hosted, and actually use the fallback response when applicable.

### Testing
Ran the notebook (changes not checked in here yet) with additional logging and saw that fallback responses are utilized.

### Time estimate
5 min


rouille (approved):
Looks good

PR #136, opened 2021-02-10
Add retry to noaa download and update rap demo notebook

jon-hagg:
### Purpose
The api sometimes returns 500 responses that consistently succeed on the second attempt (this can be seen easily in the browser as well). Seemed like a good use case for retry logic, so I ran the notebook which was previously showing missing files, and it can now download 100% of the files for the given date. 

### What the code is doing
Retry the download when response code is 500. 

### Testing
Ran the notebook.

### Time estimate
10 min, mainly to check if the notebook still looks reasonable


rouille (approved):
Works perfectly. Thanks.

PR #138, opened 2021-02-11
feat: add code to generate HIFLD based topology

terrywqf:
### Purpose
Merging the python module of the parser to generate topology based on Homeland Infra Foundation-Level Data, covering transmission network and generators.

### What it does
Move the code from private GitHub proj to PreREISE. 
To give more context, this new HIFLD module has two python scripts to output the transmission topology and power generation data.

- data_trans.py is the one to parse the substation, transmission csv and json raw data downloaded from HIFLD. 
  - `DataTransform(E_csv, T_csv, Z_csv)` is the main entry point in this script to all the other methods, i.e., parse, calcualte and approximate the topology information.
  - The most important internal analysis functions 
    - `def Neighbors(sub_by_coord_dict, sub_name_dict, raw_lines)`, which figures out the topology with geo coordinate information for all the transmission lines
    - `calculate_reactance_and_rateA(bus_id_to_kv, lines, raw_lines)`, which is using pre-defined line parameters and rules
    - `GetMaxIsland(G)`, which detects the connectivity of the network topology; we leverage the networkx as the Graph computing library - some of the topology calculation is done easily.

- plant_agg.py is the one to parse the generation raw data from HIFLD and EIA
  - `Plant(E_csv, U_csv, G2019_csv)` is the main entry point in this script to all other methods, i.e., cleanup, aggregate the power plant generation information.
  - The most important internal analysis function is `Plant_agg(clean_data,ZipOfsub_dict,loc_of_plant,LocOfsub_dict,Pmin)`, which aggregate the power generation based on zip code grouping 

### Time to review
Most of codes have been reviewed before but mainly focusing on the core logic. For this one, we can focus on the review from migration perspective (e.g., code style, naming standard) - estimate is 1h.


prereise/gather/hiflddata/README.md
```diff
@@ -0,0 +1,13 @@
...
+- Transmission Line:
+https://hifld-geoplatform.opendata.arcgis.com/datasets/electric-power-transmission-lines
+- Substation:
+https://hifld-geoplatform.opendata.arcgis.com/datasets/electric-substations
+- PowerPlant:
+https://hifld-geoplatform.opendata.arcgis.com/datasets/power-plants
+
+
+### Sample Output Data
+- https://zenodo.org/record/3530899#.X3KI2C2z03E
```
danielolsen, line 13: This URL points to the first version of our dataset.

The URL that will always redirect to the latest version of our dataset is https://zenodo.org/record/3530898

danielolsen:
Functions should have Sphix-formatted docstrings, see https://breakthrough-energy.github.io/docs/dev/code_documentation.html for more information.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,433 @@
...
+
+        sub1 = min(candidate_coordinates, key=lambda p: computeGeoDist(p, start_coord))
+        sub2 = min(candidate_coordinates, key=lambda p: computeGeoDist(p, end_coord))
+
+        # dist = meter2Mile(raw_lines['SHAPE_Length'][line_id])
+        dist = computeGeoDist(start_coord, end_coord)
+        # vol = raw_lines['VOLTAGE'][line_id]
+        # reactance = computeReactance(line_type, dist, vol)
+        # rateA = computeRateA(line_type, vol)
+        lines.append([line_id, line_type, sub_by_coord_dict.get(sub1)[0], sub_by_coord_dict.get(sub1)[1], sub_by_coord_dict.get(sub2)[0], sub_by_coord_dict.get(sub2)[1], dist])
```
danielolsen, line 82: Code should be formatted using `black`, which does not allow lines >88 characters in length.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,433 @@
...
+        candidate_coordinates = sub_name_dict.get(line['properties']['SUB_1']) + sub_name_dict.get(line['properties']['SUB_2'])
+
+        sub1 = min(candidate_coordinates, key=lambda p: computeGeoDist(p, start_coord))
+        sub2 = min(candidate_coordinates, key=lambda p: computeGeoDist(p, end_coord))
+
+        # dist = meter2Mile(raw_lines['SHAPE_Length'][line_id])
+        dist = computeGeoDist(start_coord, end_coord)
+        # vol = raw_lines['VOLTAGE'][line_id]
+        # reactance = computeReactance(line_type, dist, vol)
+        # rateA = computeRateA(line_type, vol)
```
danielolsen, line 81: Are these comments that you have forgotten to remove?

danielolsen:
> ### What it does
> 
> Move the code from private GitHub proj to PreREISE.

Could you please provide some more context for those of us who are seeing these modules for the first time? What is the main goal of each module, what are the main functions that the user will be interacting with, what are the most important internal analysis functions?

prereise/gather/hiflddata/plant_agg.py
```diff
@@ -0,0 +1,153 @@
...
+def write_gen(plant_dict, type_dict):
+    gencost = open('output/gencost.csv','w',newline='')
+    csv_writer = csv.writer(gencost)
+    csv_writer.writerow(["plant_id","type","n","c2","c1","c0"])
+    for key in plant_dict:
+        csv_writer.writerow([key[0]+'-'+key[1],key[1],1,1,type_dict[key[1]],0])
+    gencost.close()
+
+def Plant(E_csv, U_csv, G2019_csv):
+    zone_dic = get_Zone("data/zone.csv")
```
danielolsen, line 134: In general, we prefer that constants like the path here are defined at the module top-level, and there is a way for the user to pass a different path to the function if desired.

prereise/gather/hiflddata/plant_agg.py
```diff
@@ -0,0 +1,153 @@
...
+    plant = open('output/plant.csv','w',newline='')
+    csv_writer = csv.writer(plant)
+    csv_writer.writerow(["plant_id","bus_id","Pg","status","Pmax","Pmin","ramp_30","type"])
+    for key in plant_dict:
+        list1 = plant_dict[key]
+        csv_writer.writerow([key[0]+'-'+key[1],list1[0], 1, "OP", min(list1[1],list1[2]),list1[3],list1[3],key[1]])
+    plant.close()
+
+def write_gen(plant_dict, type_dict):
+    gencost = open('output/gencost.csv','w',newline='')
```
danielolsen, line 126: We should use the `with open(...) as gencost:` syntax here to make sure that we always close the file, even if there is some bug before we get to the `gencost.close()` line.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,433 @@
...
+
+
+#Write bus2sub.csv
+def Write_bus2sub(clean_data):
+    bus2sub = open('output/bus2sub.csv','w',newline='')
+    csv_writer = csv.writer(bus2sub)
+    csv_writer.writerow(["Bus_id","sub_id"])
+    for index, row in clean_data.iterrows():
+        csv_writer.writerow([row['ID'],row['ID']])
+    bus2sub.close()
```
danielolsen, line 248: If `clean_data` is a pandas DataFrame, then why do we use the low-level `csv` module and `iterrows` rather than the `DataFrame.to_csv` method?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,433 @@
...
+    branch.close()
+
+# Reactance Calculate 1
+kv_xperunit_calculate_1 = (
+    (69, 0.2293),
+    (115, 0.1106),
+    (230, 0.0567),
+    (345, 0.0074),
+    (500, 0.0001),
+    (2147483647, 0.0001)
```
danielolsen, line 266: This is a very high voltage!

danielolsen:
We should add some tests, see https://breakthrough-energy.github.io/docs/dev/testing_guide.html

jon-hagg:
Just an FYI, I recommend rebasing on `develop` branch as there have been some changes that might be of interest. Namely, we have added additional linting to `tox` (which I recommend using, as it's the easiest way to run local validation and formatting), and it should also trigger the github CI checks.

prereise/gather/hiflddata/data_trans.py
terrywqf: Thanks. This was only to represent the case that VOLTAGE is >500 as the boundary condition. The highest VOLTAGE in HIFLD transmission line data is 1000. I am changing this to 1100 to avoid potential confusions.

prereise/gather/hiflddata/data_trans.py
terrywqf: yes. removed.

prereise/gather/hiflddata/README.md
terrywqf: Thanks. Fixed.

terrywqf:
All the functions now have docstrings added. 

> Functions should have Sphinx-format docstrings, see https://breakthrough-energy.github.io/docs/dev/code_documentation.html for more information.

terrywqf:
More introductions and contexts have been added to this merge request and the README in this module.

> Could you please provide some more context for those of us who are seeing these modules for the first time? What is the main goal of each module, what are the main functions that the user will be interacting with, what are the most important internal analysis functions?

prereise/gather/hiflddata/data_trans.py
terrywqf: We were not sure the mapping between bus and substation, i.e. whether we need additional mapping info between the bus and substation so it was doing the low-level loop. In the first version, bus-to-substaion is 1: 1 mapping so only the clean_data is passed in as param. Probably prefer to leave it for now.

prereise/gather/hiflddata/plant_agg.py
terrywqf: Thanks. All the file open blocks have been updated to use `with open`

terrywqf:
Thanks for the reminder. Obviously it breaks the GitHub CI Lint stuff after I rebase the latest develop branch from upstream. Will check how to fix it.

> Just an FYI, I recommend rebasing on `develop` branch as there have been some changes that might be of interest. Namely, we have added additional linting to `tox` (which I recommend using, as it's the easiest way to run local validation and formatting), and it should also trigger the github CI checks.

prereise/gather/hiflddata/data_trans.py
danielolsen: It would be nice if we can avoid mixing fake data in with real data, and algorithmically handle the case where a line voltage is higher than what we have in the data.

danielolsen:
> 
> 
> Thanks for the reminder. Obviously it breaks the GitHub CI Lint stuff after I rebase the latest develop branch from upstream. Will check how to fix it.
> 
> > Just an FYI, I recommend rebasing on `develop` branch as there have been some changes that might be of interest. Namely, we have added additional linting to `tox` (which I recommend using, as it's the easiest way to run local validation and formatting), and it should also trigger the github CI checks.

You can click the 'details' link next to each of the failed checks and it will show you where your code doesn't conform to the `black` and `flake8` formatting/linting. Installing the `black` python package and running `black .` at the command line from the package root should automatically reformat all of your code to the `black` spec, and get rid of most of those warnings.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+    :param float dist: distance between the two substations
+    :param int vol: transmission line voltage
+    :return: (*float*) --  rateA value.
+    """
+
+    if line_type.startswith('DC'):
+        # calculate 3 -> else
+        for kv, sil in kv_rate_A_calucate_3:
+            if vol <= kv:
+                return pow(sil * 43.261 * dist, -0.6678)
```
danielolsen, line 481: Where does this equation come from?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+    """
+
+    with open('output/branch.csv', 'w', newline='') as branch:
+        csv_writer = csv.writer(branch)
+        csv_writer.writerow(['branch_id', 'line_type', 'from_bus_id', 'from_bus_name', 'to_bus_id', 'to_bus_name', 'length_in_mile', 'reactance', 'rateA'])
+        csv_writer.writerows(lines)
+
+
+# Reactance Calculate 1
+kv_xperunit_calculate_1 = (
```
danielolsen, line 356: Where do all of these constant come from?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+kv_xperunit_calculate_2 = (
+    (69, 0.0096),
+    (100, 0.0063),
+    (115, 0.0039),
+    (138, 0.0026),
+    (161, 0.0021),
+    (230, 0.0011),
+    (345, 0.0005),
+    (500, 0.0002),
+    (765, 0.0001),
```
danielolsen, line 375: Can we increase the number of significant digits for some of these smaller numbers?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+def GetMaxIsland(G):
+    """Report the largest connected island to understand the topo nature of the graph
+
+    :param networkx.Graph G: Graph generated by :func:`GraphOfNet`
+    :return: (*list*) -- list of nodes inside the largest island.
+    """
+
+    Max_nodeSet = []
+    Max_size = 0
+    #list1 = []
```
danielolsen, line 176: Is this a comment that forgot to get removed?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+    num = 0
+    for c in nx.connected_components(G):
+        num = num +1
+        nodeSet = G.subgraph(c).nodes()
+        size = len(nodeSet)
+        #list1.append(size)
+        if(size > Max_size):
+            Max_nodeSet = nodeSet
+            Max_size = size
+    #list1.sort(reverse=True)
```
danielolsen, line 186: Is this a comment that forgot to get removed?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+
+    Max_nodeSet = []
+    Max_size = 0
+    #list1 = []
+    num = 0
+    for c in nx.connected_components(G):
+        num = num +1
+        nodeSet = G.subgraph(c).nodes()
+        size = len(nodeSet)
+        #list1.append(size)
```
danielolsen, line 182: Is this a comment that forgot to get removed?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+                continue
+        else:
+            if (Max_Vol < Max_Value and Max_Vol > Min_Value):
+                base_KV = (Max_Vol + Min_Vol)/2
+            else:
+                base_KV = Min_Vol
+        KV_dict[(row['LATITUDE'], row['LONGITUDE'])] = base_KV
+    return KV_dict, to_cal
+
+# Give a node, find its neighbors in 5 iteration
```
danielolsen, line 221: Is this comment out of date now that the function below has a docstring and seems to have `depth = 1` by default?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,569 @@
...
+
+def computeReactanceAndType(line_type, KV_from, KV_to, dist, vol):
+    """Compute the reactance based on the type of each transmission line
+
+    :param str line_type: type of the transmission line between two substations
+    :param int KV_from: KV of the substation1 at one end
+    :param int KV_to: KV of the substation2 at the other end
+    :param float dist: distance between the two substations
+    :param int vol: transmission line voltage
+    :return: (*float*) --  reactance value.
```
danielolsen, line 442: It seem that the function actually has two return values?

prereise/gather/hiflddata/README.md
```diff
@@ -0,0 +1,29 @@
...
+- PowerPlant:
+https://hifld-geoplatform.opendata.arcgis.com/datasets/power-plants
+
+
+### Sample Output Data
+- https://zenodo.org/record/3530898
+
+
+### What it does
+Move the code from private GitHub proj to PreREISE.
```
danielolsen, line 17: This line is unnecessary for this README.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+
```
danielolsen, line 4: I believe these lines are unnecessary. @rouille, what do you think?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    :param str Z_csv: path of the zone.csv file
+    :return: (*dict*) -- a dict mapping the STATE to its ID.
+    """
+
+    zone = pd.read_csv(Z_csv)
+
+    # Create dictionary to store the mapping of states and codes
+    zone_dic = {}
+    for i in range(len(zone)):
+        zone_dic[zone["STATE"][i]] = zone["ID"][i]
```
danielolsen, line 33: Is this equivalent to `zone_dic = dict(zip(zone.STATE, zone.ID))`?

prereise/gather/hiflddata/data_trans.py
rouille: I agree

rouille:
One general comment regarding function/class/method/variable/attribute name:
* we use snake case with lower case for functions, methods/attributes of a class and variabls, e.g.; `my_variable` 
* we use camel case with upper case for classes. e.g., `MyClass`

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    Num_sub = len(csv_data)
+    row_indexs = []
+    for i in range(Num_sub):
+        if (
+            (csv_data["STATE"][i] not in zone_dic)
+            or (csv_data["STATUS"][i] != "IN SERVICE")
+            or (csv_data["LINES"][i] == 0)
+        ):
+            row_indexs.append(i)
+    clean_data = csv_data.drop(labels=row_indexs)
```
danielolsen, line 55: Is this equivalent to:
```python
clean_data = csv_data.query(
    "STATE in @zone_dic and STATUS == 'IN SERVICE' and LINES != 0"
)
```
?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+        with zipfile.ZipFile(
+            "data/Electric_Power_Transmission_Lines.geojson.zip", "r"
+        ) as zip_ref:
+            zip_ref.extractall("data/unzip")
+    with open(
+        "data/unzip/Electric_Power_Transmission_Lines.geojson", "r", encoding="utf8"
+    ) as fp:
+        data = json.load(fp)
+    for i in range(len(data["features"])):
+        line = data["features"][i]
```
danielolsen, line 83: These are equivalent to `for i, line in enumerate(data["features"]):`

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+            # e.g., ID 201978 with SUB2 UNKNOWN202219, never shown in substation rawdata
+            print(
+                "INFO: sub1 or sub2 name not identified for transmission line: ", line
+            )
+            missinglines.append(line)
+            continue
+
+        start_coord = (
+            line["geometry"]["coordinates"][0][0][1],
+            line["geometry"]["coordinates"][0][0][0],
```
danielolsen, line 108: What do each of these axes represent?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+            row_indexs.append(i)
+    clean_data = csv_data.drop(labels=row_indexs)
+    return clean_data
+
+
+def Neighbors(sub_by_coord_dict, sub_name_dict, raw_lines):
+    """Create dict to store the neighbors of each substation based on transmission topology geojson from HIFLD
+
+    :param dict sub_by_coord_dict: dict mapping the (x, y) to a substation
+    :param dict sub_name_dict: dict mapping the substation name to a substation
```
danielolsen, line 63: Could you be a bit more specific here about the types of the keys and value for these dictionaries?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+def GetMaxIsland(G):
+    """Report the largest connected island to understand the topo nature of the graph
+
+    :param networkx.Graph G: Graph generated by :func:`GraphOfNet`
+    :return: (*list*) -- list of nodes inside the largest island.
+    """
+
+    Max_nodeSet = []
+    Max_size = 0
+    # list1 = []
```
danielolsen, line 212: Did this comment forget to get deleted?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    num = 0
+    for c in nx.connected_components(G):
+        num = num + 1
+        nodeSet = G.subgraph(c).nodes()
+        size = len(nodeSet)
+        # list1.append(size)
+        if size > Max_size:
+            Max_nodeSet = nodeSet
+            Max_size = size
+    # list1.sort(reverse=True)
```
danielolsen, line 222: Did this comment forget to get deleted?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+
+    Max_nodeSet = []
+    Max_size = 0
+    # list1 = []
+    num = 0
+    for c in nx.connected_components(G):
+        num = num + 1
+        nodeSet = G.subgraph(c).nodes()
+        size = len(nodeSet)
+        # list1.append(size)
```
danielolsen, line 218: Did this comment forget to get deleted?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+        G.add_node(node)
+
+    for key in N_dict:
+        for value in N_dict[key]:
+            G.add_edge(key, value)
+
+    return G
+
+
+def GetMaxIsland(G):
```
danielolsen, line 203: Can this function be replaced with `list(max(nx.connected_components(G), key=len))`?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+import os.path
+import zipfile
+from collections import defaultdict
+
+import networkx as nx
+import pandas as pd
+from haversine import Unit, haversine
+
+Max_Value = 3000
+Min_Value = 0
```
danielolsen, line 16: These variables are being reassigned from within `InitKV`, if they do not need to be global variables I suggest that they be removed.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+
+    KV_dict = {}
+    to_cal = []
+    Max_Value = 3000
+    Min_Value = 0
+    for index, row in clean_data.iterrows():
+        base_KV = default_base_kv
+        Min_Vol = row["MIN_VOLT"]
+        Max_Vol = row["MAX_VOLT"]
+        if Min_Vol >= Max_Value or Min_Vol <= Min_Value:
```
danielolsen, line 243: I think these comparisons can be chained, e.g. `Min_Value <= Min_Vol <= Max Value`

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    :return: (*dict*) -- KV_dict, a dict of substation's baseKV value.
+    :return: (*list*) -- to_cal, a list of substations to be calculated using neighbour search.
+    """
+
+    KV_dict = {}
+    to_cal = []
+    Max_Value = 3000
+    Min_Value = 0
+    for index, row in clean_data.iterrows():
+        base_KV = default_base_kv
```
danielolsen, line 240: It seems that `base_KV` is always assigned within the if/else blocks or the loop is `continue`-d before the loop completes. If that is the case, then, what's the point of the `default_base_kv`?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    Save the invalid substations as to be calculated later
+
+    :param pandas.DataFrame clean_data: substation dataframe as returned by :func:`Clean`
+    :return: (*dict*) -- KV_dict, a dict of substation's baseKV value.
+    :return: (*list*) -- to_cal, a list of substations to be calculated using neighbour search.
+    """
+
+    KV_dict = {}
+    to_cal = []
+    Max_Value = 3000
```
danielolsen, line 237: The data seems to only have substations with max voltages up to 1000 kV. Allowing anything up to 3000 kV seems to be overly lax.

It would be nice to have these as parameters with default values, rather than hardcoded.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+            [
+                line_id,
+                line_type,
+                sub_by_coord_dict.get(sub1)[0],
+                sub_by_coord_dict.get(sub1)[1],
+                sub_by_coord_dict.get(sub2)[0],
+                sub_by_coord_dict.get(sub2)[1],
+                dist,
+            ]
+        )
```
danielolsen, line 133: Returning values as an unlabeled list here makes the code a bit hard to understand, especially once we get to `calculate_reactance_and_rateA`. Could these be a dictionary or dataframe instead?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+# Give a node, find its neighbors in 5 iteration
+
+
+def get_neigbors(g, node, depth=1):
+    """Give a node, find its neighbors in given number depth
+
+    :param networkx.Graph g: Graph generated by :func:`GraphOfNet`
+    :param tuple node: substation coordinate tuple
+    :param int depth: number of hops that we are searching the neighbor substations
+    :return: (*list*) --  a list of closest substations.
```
danielolsen, line 267: Does this function really return a list? It looks like it returns a dictionary.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+        temp_KV = 0
+        num = 0
+        for i in range(1, 6):
+            for nei in neigh[i]:
+                if nei in KV_dict:
+                    temp_KV = temp_KV + KV_dict[nei]
+                    num = num + 1
+            if num > 0:
+                KV_dict[sub] = temp_KV / num
+                continue
```
danielolsen, line 303: Is this equivalent to:
```python
matching_neighbors = set(neigh[i]) & set(KV_dict.keys())
if len(matching_neighbors) > 0:
    KV_dict[sub] = sum(KV_dict[n] for n in matching_neighbors) / len(matching_neighbors)
    continue
```
?

danielolsen: Another question: why do we have a `continue` in this for loop, rather than a `break`? I would think once we've found neighbors with good voltages at one depth, we don't need to continue to find neighbors at further depths, right?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+
+def Cal_KV(N_dict, G, KV_dict, to_cal):
+    """Estimate missing substation's KV data using the neighbors average calculation
+
+    :param dict N_dict: dict of substation's neighbors as returned by :func:`Neighbors`
+    :param networkx.Graph g: Graph generated by :func:`GraphOfNet`
+    :param dict KV_dict: dict of existing baseKV for each substation
+    :param list to_cal: list of the substations to be calculate using average KV
+    :return: (*dict*) --  dict of all substation's baseKV being corrected and calculated.
+    """
```
danielolsen, line 288: It seems that this function modifies KV_dict inplace and also returns it? What is the point of doing both?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    nodes = [node]
+    for i in range(1, depth + 1):
+        output[i] = []
+        for x in nodes:
+            output[i].extend(layers.get(x, []))
+        nodes = output[i]
+    return output
+
+
+def Cal_KV(N_dict, G, KV_dict, to_cal):
```
danielolsen, line 280: This function needs tests.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+        for i in range(1, 6):
+            for nei in neigh[i]:
+                if nei in KV_dict:
+                    temp_KV = temp_KV + KV_dict[nei]
+                    num = num + 1
+            if num > 0:
+                KV_dict[sub] = temp_KV / num
+                continue
+            else:
+                KV_dict[sub] = -999999
```
danielolsen, line 305: If we assign a voltage of -1 GV to a substation within this loop, could that value get used in calculating voltages for this substations neighbors in future iterations of the loop?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+    """
+
+    sub_by_coord_dict = {}
+    sub_name_dict = {"sub": []}
+    raw_subs = pd.read_csv(E_csv)
+    for index, row in raw_subs.iterrows():
+        location = (row["LATITUDE"], row["LONGITUDE"])
+        if location in sub_by_coord_dict:
+            print("WARNING: substations coordinates conflict check:", location)
+        sub_by_coord_dict[location] = (row["ID"], row["NAME"])
```
danielolsen, line 324: It seems dangerous to simply overwrite the data with a printed warning to the user. I think it would be better to raise an Exception.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+            csv_writer.writerow(
+                [
+                    row["ID"],
+                    row["NAME"],
+                    row["LATITUDE"],
+                    row["LONGITUDE"],
+                    zone_dic[row["STATE"]],
+                    row["TYPE"],
+                ]
+            )
```
danielolsen, line 351: Is this equivalent to:
```python
output = pd.DataFrame()
output["sub_id"] = clean_data["ID"]
output["sub_name"] = clean_data["NAME"]
output["lat"] = clean_data["LATITUDE"]
output["lon"] = clean_data["LONGITUDE"]
output["zone_id"] = clean_data.apply(lambda x: zone_dic[x["STATE"]], axis=1)
output["type"] = clean_data["TYPE"]
output.to_csv("output/sub.csv", index=False)
```
?

prereise/gather/hiflddata/plant_agg.py
```diff
@@ -0,0 +1,211 @@
...
+            format(row["LONGITUDE"], coord_precision),
+        )
+        loc_of_plant[row["NAME"]] = loc
+    return loc_of_plant
+
+
+def Plant_agg(clean_data, ZipOfsub_dict, loc_of_plant, LocOfsub_dict, Pmin):
+    """Aggregate the plant by zip code and build the plant dict with geo-based aggregation
+
+    :return: (*dict*) -- a dict for power plant after aggregation
```
danielolsen, line 86: Can you be more specific with what exactly the keys and values of this dict represent? It seems that each value is an unlabeled list of floats, which makes it hard to understand what's going on in the `write_plant` function which is heavily dependent on the position of these values.

prereise/gather/hiflddata/plant_agg.py
```diff
@@ -0,0 +1,211 @@
+import csv
+
+import pandas as pd
+from geopy.distance import geodesic
+
+from prereise.gather.hiflddata.data_trans import Clean, get_Zone
+
+Max_Value = 3000
+Min_Value = 0
```
danielolsen, line 9: It seems that these values are not used in this module?

prereise/gather/hiflddata/plant_agg.py
```diff
@@ -0,0 +1,211 @@
...
+        loc = (
+            format(row["LATITUDE"], coord_precision),
+            format(row["LONGITUDE"], coord_precision),
+        )
+        loc_of_plant[row["NAME"]] = loc
+    return loc_of_plant
+
+
+def Plant_agg(clean_data, ZipOfsub_dict, loc_of_plant, LocOfsub_dict, Pmin):
+    """Aggregate the plant by zip code and build the plant dict with geo-based aggregation
```
danielolsen, line 84: This docstring needs information on the inputs.

danielolsen:
Do we use `geopy` for anything besides the distance between two pairs of coordinates? If not, can we substitute `haversine` instead?

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+from collections import defaultdict
+
+import networkx as nx
+import pandas as pd
+from haversine import Unit, haversine
+
+Max_Value = 3000
+Min_Value = 0
+coord_precision = ".9f"
+default_base_kv = -999999.0
```
danielolsen, line 18: We should strive to avoid the use of fake data which might accidentally get used in calculations.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,645 @@
...
+import zipfile
+from collections import defaultdict
+
+import networkx as nx
+import pandas as pd
+from haversine import Unit, haversine
+
+Max_Value = 3000
+Min_Value = 0
+coord_precision = ".9f"
```
danielolsen, line 17: It seems this variable is not being used in this module?

prereise/gather/hiflddata/tests/test_plant_agg.py
```diff
@@ -0,0 +1,135 @@
...
+    Loc_of_plant,
+    LocOfsub,
+    Plant_agg,
+)
+
+
+def test_clean_p():
+    csv_data_path = "General_Units.csv"
+    csv_data = {"STATE": ["AL", "AR"], "STATUS": ["OP", "Other"]}
+    pd.read_csv = Mock(return_value=pd.DataFrame(data=csv_data))
```
danielolsen, line 19: Over-writing the pandas `read_csv` function is what's causing other tests to break, this needs to be removed. You can refactor the cleaning functions to take a dataframe, rather than a path to a CSV file.

prereise/gather/hiflddata/data_trans.py
terrywqf: this formula coming from Yifan's doc

prereise/gather/hiflddata/data_trans.py
terrywqf: These number are also from Yifan's doc. I guess we can discuss with Yifan again on their precision and address them in future.

prereise/gather/hiflddata/data_trans.py
terrywqf: line["geometry"]["coordinates"][0][0] is the start point (substation) for this transmission line, with (line["geometry"]["coordinates"][0][0][0], line["geometry"]["coordinates"][0][0][1]) being the (latitude, longitude) for this substation.

prereise/gather/hiflddata/data_trans.py
terrywqf: it has that default value line because in early stage, we did not use the neighbor search avg method and mark it with `default_base_kv`. Mingze did not remove that line after we start using the new avg method. agree we can remove `default_base_kv`.

danielolsen:
The upstream `PreREISE/develop` now has automated checks of PEP8 function/variable naming, if you rebase to the latest the automated checks will tell you where there are issues to be addressed.

prereise/gather/hiflddata/data_trans.py
danielolsen: What do the first and second zeros in `[0][0][0]`/`[0][0][1]` represent? What are these layers of indexing/slicing? A comment in the source code would make this more clear.

prereise/gather/hiflddata/data_trans.py
terrywqf: agree - will use `InitKV(clean_data, Max_value=1100)` as parameter

prereise/gather/hiflddata/plant_agg.py
terrywqf: Changed zone csv path as a parameter like others.
```
def Plant(E_csv, U_csv, G2019_csv, Z_csv):
```

prereise/gather/hiflddata/data_trans.py
terrywqf: Good catch, this is way more concise! will modify. thanks.

prereise/gather/hiflddata/data_trans.py
terrywqf: Yes, will change.

prereise/gather/hiflddata/data_trans.py
terrywqf: will delete

prereise/gather/hiflddata/data_trans.py
terrywqf: will delete

prereise/gather/hiflddata/data_trans.py
terrywqf: will delete

prereise/gather/hiflddata/data_trans.py
terrywqf: Yes, I will remove them

prereise/gather/hiflddata/data_trans.py
terrywqf: will change.

prereise/gather/hiflddata/data_trans.py
terrywqf: you are right, should be `  :return: (*dict*) --  a dict mapping from depth to closest substations on that depth.
`

prereise/gather/hiflddata/data_trans.py
terrywqf: you are right, no need to return it. will remove the `return` in the next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: agree, will raise an exception in the next revision.

prereise/gather/hiflddata/plant_agg.py
terrywqf: will remove them in next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: This will change the float precision of the output file. Prefer to leave it for now.

prereise/gather/hiflddata/data_trans.py
terrywqf: will remove them in next revision.

prereise/gather/hiflddata/tests/test_plant_agg.py
terrywqf: I have applied `@patch("prereise.gather.hiflddata.plant_agg.pd.read_csv")` to the unit test. it will be only effective within the module.

prereise/gather/hiflddata/data_trans.py
terrywqf: will add test in next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: Awesome! this will get exactly the same result. will replace in next revision. thanks.

terrywqf:
not for now, but geopy is a more general lib for comprehensive geo computing purpose. will check with Mingze if he plan to use this for other data processing in plant generation profile.
> Do we use `geopy` for anything besides the distance between two pairs of coordinates? If not, can we substitute `haversine` instead?

prereise/gather/hiflddata/data_trans.py
terrywqf: We should use a `break` for this case. The algo is supposed to do a BFS to approximate the missing base_KV and break early. thanks for finding this.

terrywqf: > Is this equivalent to:
> 
> ```python
> matching_neighbors = set(neigh[i]) & set(KV_dict.keys())
> if len(matching_neighbors) > 0:
>     KV_dict[sub] = sum(KV_dict[n] for n in matching_neighbors) / len(matching_neighbors)
>     continue
> ```
> 
> ?

The code is more concise but not quite efficient. I have tried it and it would take long time to process the data. So I prefer to keep it as it is.

prereise/gather/hiflddata/plant_agg.py
terrywqf: will add input details in next revision.

prereise/gather/hiflddata/plant_agg.py
terrywqf: Will add details in next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: Too many places depend on the return value `lines`, prefer to leave it for now. but I will add more comments to describe each column.

prereise/gather/hiflddata/data_trans.py
terrywqf: will add more details in next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: Ok, I will add more detailed comments in next revision.

terrywqf:
> The upstream `PreREISE/develop` now has automated checks of PEP8 function/variable naming, if you rebase to the latest the automated checks will tell you where there are issues to be addressed.

I will refactor the function/variable naming in next revision.

danielolsen:
Looking at the correspondence between the currently-generated **branch.csv** and the currently-generated **bus.csv**, it seems there are 3,744 buses that are listed as either the `from_bus_id` or the `to_bus_id` for at least one branch, but not present in the bus table:
```python
>>> import pandas as pd
>>> branch = pd.read_csv("branch.csv", index_col=0)
>>> bus = pd.read_csv("bus.csv", index_col=0)
>>> len((set(branch.from_bus_id) - set(bus.index)) | (set(branch.to_bus_id) - set(bus.index)))
3744
```

prereise/gather/hiflddata/data_trans.py
danielolsen: What is the precision difference?

danielolsen:
There are 372 unique voltage levels in **bus.csv**, 50 of which are less than zero.
```python
>>> import pandas as pd
>>> bus = pd.read_csv("bus.csv", index_col=0)
>>> len(bus.base_KV.unique())
372
>>> len(bus.query("base_KV < 0").base_KV.unique())
50
```

prereise/gather/hiflddata/data_trans.py
danielolsen: I think the fact that many places rely on a particular ordering of unlabeled entries is all the more reason to change it. This seems like a perfect use case for a data frame instead.

prereise/gather/hiflddata/data_trans.py
danielolsen: What is the time difference, and where does it come from? The two seem to be doing the exact same calculations, with the exception of re-generating `set(KV_dict.keys())`, which could probably be moved outside of the loops. This would probably also address the use of voltages that are set in this loop in subsequent loop iterations, by freezing the set of 'real' data before we start filling in estimated data.

prereise/gather/hiflddata/data_trans.py
terrywqf: Ok, will change it to dataframe.

prereise/gather/hiflddata/data_trans.py
terrywqf: will use DataFrame.to_csv in next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: Please ignore my last comments. it does not lose precision. Excel displaying loss precision but the csv itself does not loss.  will change it to DataFrame in next revision.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,600 @@
...
+    (100, 30),
+    (115, 35),
+    (138, 50),
+    (161, 69),
+    (230, 140),
+    (345, 375),
+    (500, 1000),
+    (765, 2250),
+    (1100, 2250),
+)
```
danielolsen, line 419: This variable seems to never be used, but we use `for kv, sil in kv_rate_a_calucate_3` within `compute_rate_a`. Should we be using this variable instead of `kv_rate_a_calucate_3`?

terrywqf: That is a good catch! This should be applied when the line length>=50 miles. Will correct the copy-paste error.

prereise/gather/hiflddata/data_trans.py
terrywqf: The time difference is from `set(neigh[i]) & set(KV_dict.keys())`, your are right, no need to perform set operation for `KV_dict.keys()` in the for loop each time. 
Will replace by 
```python
            count = sum(1 for neighbor in neighbors[depth] if neighbor in kv_dict)
            if count > 0:
                sum_kv = sum(kv_dict[nei] for nei in neighbors[depth] if nei in kv_dict)
                kv_dict[sub] = sum_kv / count
                break
```

prereise/gather/hiflddata/data_trans.py
terrywqf: will change it in next revision.

prereise/gather/hiflddata/data_trans.py
terrywqf: will remove

prereise/gather/hiflddata/README.md
terrywqf: will remove

prereise/gather/hiflddata/data_trans.py
terrywqf: will add doc string for another return value in next revision.

terrywqf:
> Looking at the correspondence between the currently-generated **branch.csv** and the currently-generated **bus.csv**, it seems there are 3,744 buses that are listed as either the `from_bus_id` or the `to_bus_id` for at least one branch, but not present in the bus table:
> 
> ```python
> >>> import pandas as pd
> >>> branch = pd.read_csv("branch.csv", index_col=0)
> >>> bus = pd.read_csv("bus.csv", index_col=0)
> >>> len((set(branch.from_bus_id) - set(bus.index)) | (set(branch.to_bus_id) - set(bus.index)))
> 3744
> ```

This has been addressed in the latest commit - there were some `from_bus_id` or the `to_bus_id` not in the bus.csv because of some dirty data. Did some cleanup and this check should 0 now.

terrywqf:
> There are 372 unique voltage levels in **bus.csv**, 50 of which are less than zero.
> 
> ```python
> >>> import pandas as pd
> >>> bus = pd.read_csv("bus.csv", index_col=0)
> >>> len(bus.base_KV.unique())
> 372
> >>> len(bus.query("base_KV < 0").base_KV.unique())
> 50
> ```

The negative base_kv has been addressed. It is caused by the fake -999999.0 setup during the estimation. After the cleanup, we now have all positive base_kv in bus.csv.

For the 300+ diff base_kv level, it is because we applied the neighborhood average for those buses with missing voltage as bad data.

prereise/gather/hiflddata/data_trans.py
terrywqf: This has been addressed in the latest version. We no longer apply such fake setup with -999999.0

prereise/gather/hiflddata/data_trans.py
terrywqf: this does get used during some estimation in the future iteration and it caused some bus's base_kv being negative as the final results. I have removed this setup in the latest version and cleanup the data.

prereise/gather/hiflddata/data_trans.py
terrywqf: I have explained why it gets slow down and use an improved count for that purpose. Now the sum is applied to simplify these lines with the count. Please raise if any more concern. Resolving this now.

danielolsen:
> 
> 
> > Looking at the correspondence between the currently-generated **branch.csv** and the currently-generated **bus.csv**, it seems there are 3,744 buses that are listed as either the `from_bus_id` or the `to_bus_id` for at least one branch, but not present in the bus table:
> > ```python
> > >>> import pandas as pd
> > >>> branch = pd.read_csv("branch.csv", index_col=0)
> > >>> bus = pd.read_csv("bus.csv", index_col=0)
> > >>> len((set(branch.from_bus_id) - set(bus.index)) | (set(branch.to_bus_id) - set(bus.index)))
> > 3744
> > ```
> 
> This has been addressed in the latest commit - there were some `from_bus_id` or the `to_bus_id` not in the bus.csv because of some dirty data. Did some cleanup and run this check again - it returns 0 now.

How did you do the data cleanup? We want all the code that goes from the raw input files to our outputs to be in the codebase, so that we can repeat the whole process ourselves.

terrywqf:
> > > Looking at the correspondence between the currently-generated **branch.csv** and the currently-generated **bus.csv**, it seems there are 3,744 buses that are listed as either the `from_bus_id` or the `to_bus_id` for at least one branch, but not present in the bus table:
> > > ```python
> > > >>> import pandas as pd
> > > >>> branch = pd.read_csv("branch.csv", index_col=0)
> > > >>> bus = pd.read_csv("bus.csv", index_col=0)
> > > >>> len((set(branch.from_bus_id) - set(bus.index)) | (set(branch.to_bus_id) - set(bus.index)))
> > > 3744
> > > ```
> > 
> > 
> > This has been addressed in the latest commit - there were some `from_bus_id` or the `to_bus_id` not in the bus.csv because of some dirty data. Did some cleanup and run this check again - it returns 0 now.
> 
> How did you do the data cleanup? We want all the code that goes from the raw input files to our outputs to be in the codebase, so that we can repeat the whole process ourselves.

It is in the code. The input data are all as before (as you can see in the latest commit).
For this cleanup, I added below to filter out the transmission lines having `from_bus_id` or `to_bus_id` not in the bus table:

lines = lines[lines["from_bus_id"].isin(bus_id_to_kv.keys())]
lines = lines[lines["to_bus_id"].isin(bus_id_to_kv.keys())]

danielolsen:
I see now, it's added in `calculate_reactance_and_rate_a`. Thanks.

prereise/gather/hiflddata/data_trans.py
```diff
@@ -0,0 +1,583 @@
...
+    :return: (*networkx.Graph*) -- Graph generated by networkx.
+    """
+    nodes = list(n_dict.keys())
+    graph = nx.Graph()
+    any(graph.add_node(node) for node in nodes)
+    any(
+        graph.add_edge(node, neighbor)
+        for node, neighbors in n_dict.items()
+        for neighbor in neighbors
+    )
```
danielolsen, line 181: Why do we evaluate `any` on these repeated graph method calls?

danielolsen:
Evaluating the connectivity of the `branch` data frame, it seems that there is one very large connected component, and many smaller ones:
```python
>>> import networkx as nx
>>> import pandas as pd
>>> branch = pd.read_csv("branch.csv", index_col=0)
>>> g = nx.from_pandas_edgelist(branch, "from_bus_id", "to_bus_id")
>>> [len(c) for c in sorted(nx.connected_components(g), key=len)]
[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 7, 7, 7, 7, 7, 7, 7, 7, 8, 8, 8, 8, 9, 9, 10, 10, 11, 12, 13, 13, 13, 14, 24, 24, 37, 41, 54, 60219]
```
I would have expected to see three large ones (one for each interconnect). Maybe DC lines bridging the interconnects are still present in the branch CSV?

When we look at one of these connected components of length `1`, we see that it is because there is one branch that connects a bus to itself, and no other branches connected to that bus:
```python
>>> a = list(sorted(nx.connected_components(g), key=len))
>>> a[0]
{205988}
>>> branch.query("from_bus_id == 205988 or to_bus_id == 205988")
          line_type  from_bus_id  ... reactance  rateA
branch_id                         ...
205463         Line       205988  ...  0.000044  239.0

[1 rows x 8 columns]
>>> branch.loc[205463]
line_type              Line
from_bus_id          205988
from_bus_name     TAP205988
to_bus_id            205988
to_bus_name       TAP205988
length_in_mile     0.011229
reactance          0.000044
rateA                 239.0
Name: 205463, dtype: object
```

We also see that there are 608 buses that are not connected to any branches at all:
```python
>>> bus = pd.read_csv("bus.csv", index_col=0)
>>> bus.loc[~(bus.index.isin(branch.from_bus_id) | bus.index.isin(branch.to_bus_id))]
        pd  zone_id  base_kv
bus_id
108141   0        2    161.0
111141   0       25    230.0
115110   0       43    115.0
115925   0       16    138.0
116259   0       16     69.0
...     ..      ...      ...
310660   0       41     69.0
310684   0        4    115.0
310695   0        4     66.0
310814   0       41    138.0
310864   0        4     66.0

[608 rows x 3 columns]
```

prereise/gather/hiflddata/data_trans.py
terrywqf: Using `any` here is quite confusing and it was some initial attempt to refactor some lines into more Python fashion. I think it is better to revert this back using for loop. Will also switch to graph.add_nodes_from to bulk add all the nodes.

terrywqf:
> gth `1`, we see that it is because there is one branch that connects a bus to itself, and no other branches connected to that b

Thanks - these are good checks. Regarding the shape of the network (i.e. three interconnect), we can do more visualization or use some library to understand that. Need better knowledge about the Grid. Is there any tool in PowerSim for this purpose? 

For the observation of "608 buses that are not connected to any branches" - that is possible and we better remove these buses in our bus.csv. 
Checking some example in your report:
115110 is a bus_id with name being UNKNOWN115110. And it only appears in one line as below in HIFLD transmission line raw data:

73963 | 135870 | AC; OVERHEAD | NOT AVAILABLE | 221121 | ELECTRIC BULK POWER TRANSMISSION AND CONTROL | IMAGERY, OpenStreetMap, http://www.pjm.com/, EIA 861 | ########## | IMAGERY | ########## | VIRGINIA ELECTRIC & POWER CO | 115 | 100-161 | Y | **UNKNOWN115110 | NOT AVAILABLE** | 171.619686

Basically it is connecting to NOT AVAILABLE and we could not identify this line. 

310684 is another bus_id with name being SPRING GAP TAP and it does not show up in the HIFLD transmission line raw data.

I will add some logic to remove those buses (substations) with not connecting line.

danielolsen:
If you run the same checks on the `usa_tamu` branch.csv file, you can see that there are three interconnects with no branches that connect them:
```python
>>> import networkx as nx
>>> import pandas as pd
>>> branch = pd.read_csv("branch.csv", index_col=0)
>>> g = nx.from_pandas_edgelist(branch, "from_bus_id", "to_bus_id")
>>> [len(c) for c in sorted(nx.connected_components(g), key=len)]
[2000, 10024, 70047]
```
We do not have any specialized tools to help diagnose problems with the grid topology, PowerSimData assumes that the information going into it is internally consistent, i.e. every bus has at least one connected branch, and there are no branches whose endpoints are not present in the bus list, etc.:
```python
>>> bus = pd.read_csv("bus.csv", index_col=0)
>>> branch.loc[~(branch.from_bus_id.isin(bus.index) & branch.to_bus_id.isin(bus.index))]
Empty DataFrame
Columns: [from_bus_id, to_bus_id, r, x, b, rateA, rateB, rateC, ratio, angle, status, angmin, angmax, Pf, Qf, Pt, Qt, mu_Sf, mu_St, mu_angmin, mu_angmax, branch_device_type, interconnect]
Index: []
>>> bus.loc[~(bus.index.isin(branch.from_bus_id) | bus.index.isin(branch.to_bus_id))]
Empty DataFrame
Columns: [type, Pd, Qd, Gs, Bs, zone_id, Vm, Va, baseKV, loss_zone, Vmax, Vmin, lam_P, lam_Q, mu_Vmax, mu_Vmin, interconnect]
Index: []
```

danielolsen:
I was able to successfully instantiate a Grid object with this branch using the following steps:
1. Run **data_trans.py**
2. Run **plant_agg.py**
3. Copy the **zone.csv** file from the `data` folder to the `output` folder
4. Copy everything from the `output` folder to **powersimdata/network/hifld/data**
5. With the `daniel/read_hifld_grid` branch of PowerSimData checked out:
```python
from powersimdata import Grid
usa = Grid("USA", source="hifld")
```

Examining the **branch.csv** and **bus.csv** CSV files, we still see some strangeness with the grid topology--one big connected component and many smaller ones, and 296 buses that are not connected to any branches--but successfully instantiating a Grid object is a big milestone.

prereise/gather/hiflddata/plant_agg.py
```diff
@@ -0,0 +1,387 @@
...
+                cur = points[r]
+
+            else:
+                cur = 0
+            if r in region:
+                re = region[r]
+            else:
+                if row["STATE"] in West:
+                    re = "Western"
+                elif row["STATE"] is "TX":
```
danielolsen, line 200: When running this script, I get a warning at this line:
```python
plant_agg.py:200: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif row["STATE"] is "TX":
```
I believe `is` and `==` have different meaning here.

danielolsen:
With the latest **branch.csv** file, it seems there are 595 branches that connect from a bus labelled as being in one interconnection to a bus labelled as being in another:
```python
>>> import pandas as pd
>>> branch = pd.read_csv("branch.csv", index_col=0)
>>> bus = pd.read_csv("bus.csv", index_col=0)
>>> branch.loc[:, "from_interconnect"] = bus.loc[branch.from_bus_id, "interconnect"].tolist()
>>> branch.loc[:, "to_interconnect"] = bus.loc[branch.to_bus_id, "interconnect"].tolist()
>>> branch.query("from_interconnect != to_interconnect")
             line_type  from_bus_id  ... from_interconnect  to_interconnect
branch_id                            ...
212240     Transformer       205572  ...           Eastern          Western
212241     Transformer       209865  ...           Western          Eastern
200105     Transformer       201012  ...           Eastern          Western
200106            Line       200964  ...           Western          Eastern
200112            Line       201805  ...           Eastern          Western
...                ...          ...  ...               ...              ...
158580            Line       309150  ...             Texas          Eastern
103085            Line       108107  ...           Eastern            Texas
114127     Transformer       150150  ...           Eastern            Texas
162342     Transformer       149566  ...           Eastern            Texas
167421     Transformer       173167  ...           Eastern            Texas

[595 rows x 11 columns]
```

YifanLi86:
Some additional checks suggested (time permitting, not required to complete before project ends. @danielolsen)
1. Sanity check all R, X value to see if they are within normal range and from the document's value;
2. Sanity check all line length to see if they are within normal range for a respective voltage class (e.g.: 69kV line normally won't go beyond 50miles, most AC lines shall be less than 300 miles and rarely go beyond, etc.).
3. Sanity check all RateA for significant anomalies;
4. As DCOPF being feasible, compare all branch flows to RateA to identify major "overload";
5. For parallel branches (do we have any? if so,) compare the parameters between/among them to see if the numbers are significantly different;
6. (Do we have polynomial heat rate/gen cost curve? If so, check if the heat rate/gen cost for each capacity segment is ascending;
7. Make sure capacity segment itself is ascending.

danielolsen:
Several of the functions seem to use the **needs.csv** file. Where does this file come from? The same applies to the **curve.csv** file.

The data files that are from external sources and unchanged (besides file conversion, e.g. from xlsx to csv) should go through the Data Intake procedure: see https://breakthrough-energy.github.io/docs/dev/contribution_guide.html#data-intake-procedure. This should be fairly simple for the ones that come from the federal government, since works from the U.S. government are not covered by copyright.

The intermediate data files that we create from those external sources should have all of the code to re-generate them from source also in the codebase.

danielolsen:
> Some additional checks suggested (time permitting, not required to complete before project ends. @danielolsen)
> 
>     1. Sanity check all R, X value to see if they are within normal range and from the document's value;
> 
>     2. Sanity check all line length to see if they are within normal range for a respective voltage class (e.g.: 69kV line normally won't go beyond 50miles, most AC lines shall be less than 300 miles and rarely go beyond, etc.).
> 
>     3. Sanity check all RateA for significant anomalies;
> 
>     4. As DCOPF being feasible, compare all branch flows to RateA to identify major "overload";
> 
>     5. For parallel branches (do we have any? if so,) compare the parameters between/among them to see if the numbers are significantly different;
> 
>     6. (Do we have polynomial heat rate/gen cost curve? If so, check if the heat rate/gen cost for each capacity segment is ascending;
> 
>     7. Make sure capacity segment itself is ascending.

Thanks for the ideas @YifanLi86. A couple clarifying questions while these ideas are still fresh:
- For (1) & (2), what is 'normal' range?
- For (3), what is a 'significant' anomaly?
- For (4), the DCOPF will always respect branch flows. What would an 'overload' condition be?
- For (6), I think we will be getting linear or quadratic cost curves (@LFPower would know more). Do you mean that all constant, linear, and quadratic terms should be non-negative?
- For (7), I'm not sure I understand the meaning here. Could you elaborate?

LFPower:
> Several of the functions seem to use the **needs.csv** file. Where does this file come from? The same applies to the **curve.csv** file.
> 
> The data files that are from external sources and unchanged (besides file conversion, e.g. from xlsx to csv) should go through the Data Intake procedure: see https://breakthrough-energy.github.io/docs/dev/contribution_guide.html#data-intake-procedure. This should be fairly simple for the ones that come from the federal government, since works from the U.S. government are not covered by copyright.
> 
> The intermediate data files that we create from those external sources should have all of the code to re-generate them from source also in the codebase.

We will add the data source.

YifanLi86:
> > Some additional checks suggested (time permitting, not required to complete before project ends. @danielolsen)
> > ```
> > 1. Sanity check all R, X value to see if they are within normal range and from the document's value;
> > 
> > 2. Sanity check all line length to see if they are within normal range for a respective voltage class (e.g.: 69kV line normally won't go beyond 50miles, most AC lines shall be less than 300 miles and rarely go beyond, etc.).
> > 
> > 3. Sanity check all RateA for significant anomalies;
> > 
> > 4. As DCOPF being feasible, compare all branch flows to RateA to identify major "overload";
> > 
> > 5. For parallel branches (do we have any? if so,) compare the parameters between/among them to see if the numbers are significantly different;
> > 
> > 6. (Do we have polynomial heat rate/gen cost curve? If so, check if the heat rate/gen cost for each capacity segment is ascending;
> > 
> > 7. Make sure capacity segment itself is ascending.
> > ```
> 
> Thanks for the ideas @YifanLi86. A couple clarifying questions while these ideas are still fresh:
> 
> * For (1) & (2), what is 'normal' range?
> * For (3), what is a 'significant' anomaly?
> * For (4), the DCOPF will always respect branch flows. What would an 'overload' condition be?
> * For (6), I think we will be getting linear or quadratic cost curves (@LFPower would know more). Do you mean that all constant, linear, and quadratic terms should be non-negative?
> * For (7), I'm not sure I understand the meaning here. Could you elaborate?

Updated a little bit, please review and see if adjustment needed. Thanks.

PR #140, opened 2021-02-12
ci: remove obsolete token

jon-hagg:
### Purpose
Remove token used to clone previously private repository.

### Testing
The green check mark (pending, but assuming it passes)

### Time estimate
1 min


rouille (approved):
Cool

PR #141, opened 2021-02-12
ci: enable PR trigger

jon-hagg:
### Purpose
Add PR trigger to support forked repositories. Turns out this is not a security concern, as actions triggered by external PRs run in a limited security context. Some details [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request), and [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests).

### Testing
Will have to merge to actually test this.

### Time estimate
5 min

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,10 @@
 name: Lint
 
-on: push
+on: 
+  push:
+  pull_request:
+    branches:
+      - develop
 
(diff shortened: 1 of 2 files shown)


rouille (approved):
Thanks for sharing the links, it helps understanding what is going on

BainanXia (approved):
Thanks.

PR #143, opened 2021-02-13
ci: add flake8 and fix warnings

jon-hagg:
### Purpose
Add flake8 to linting workflows, as we have in other packages. Configuration in tox.ini is similar to elsewhere, but also excluding notebook files which exist locally, but outside of git. 

### What the code is doing
* Add flake8 option and configuration to tox.ini 
* Configured github action to run flake8, via tox
* Fixes to source files, mainly unused imports and variables

### Testing
tox

### Time estimate
10 min


rouille (approved):
Looks good

PR #147, opened 2021-02-17
refactor: upgrade pyproj

### LINKED ISSUE #144: Upgrade pyproj requirement from 1.9.6
pyproj v1.9.6 is only compatible with Python 3.7 and lower. (see https://pyproj4.github.io/pyproj/stable/history.html#id20)

We want to be able to run with Python 3.8 and above, so we need to upgrade to at least pyproj 2.x, if not the new 3.0.

Looking at wheat we did in PostREISE (https://github.com/Breakthrough-Energy/PostREISE/pull/154), there's not a perfect 1:1 translation to our usage here, since PostREISE had projections based on EPSG reference systems, while we've got a projection based on a PROJ string. It seems like using these strings, rather than an EPSG or similar, is gently advised-against (see https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems), so if it's easy to use an EPSG reference system then maybe we can kill two birds with one stone.

danielolsen:
### Purpose
Upgrade pyproj to v2.6.1 for compatibility with Python >3.7. Closes #144.

### What the code is doing
Functional changes are in **prereise/gather/solardata/ga_wind**:
- In **helpers.py**, we refactor the `ll2ij` function to take a pyproj Transformer object, for faster repeated transformations (see https://pyproj4.github.io/pyproj/v2.6.1rel/advanced_examples.html#advanced-examples for more details).
- In **ga_wind.py**, we instantiate a Transformer object using the projection string that had previously been used to create a `Proj` object in the old syntax, and then pass that to `ll2ij` along with the coordinates to transform. The documentation for this instantiation method can be found at https://pyproj4.github.io/pyproj/v2.6.1rel/api/transformer.html#pyproj.transformer.Transformer.from_pipeline.
- In **tests/test_solar_helpers.py**, we add a couple of test cases to ensure that we are getting the same outputs pre- & post-refactor.

The **Pipfile** and **Pipfile.lock** were regenerated using:
```bash
pipenv sync --python 3.7
pipenv uninstall pyproj --skip-lock
pipenv install pyproj==2.6.1.post1 --skip-lock
pipenv lock --pre
```

### Testing
Unit tests were added that match the old code, and they continue to work for the new code.

Previously, running `pipenv sync` failed on Windows using Python 3.8.3, but now works successfully (although there are a few warnings):
```bash
PS C:\Users\DanielOlsen\repos\bes\PreREISE> pipenv sync
Creating a virtualenv for this project…
Pipfile: C:\Users\DanielOlsen\repos\bes\PreREISE\Pipfile
Using C:/Python38/python.exe (3.8.3) to create virtualenv…
[ ===] Creating virtual environment...created virtual environment CPython3.8.3.final.0-64 in 3731ms
  creator CPython3Windows(dest=C:\Users\DanielOlsen\.virtualenvs\PreREISE-JbwmpG5i, clear=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=C:\Users\DanielOlsen\AppData\Local\pypa\virtualenv)
    added seed packages: pip==20.2.3, setuptools==51.3.3, wheel==0.36.2
  activators BashActivator,BatchActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator

Successfully created virtual environment!
Virtualenv location: C:\Users\DanielOlsen\.virtualenvs\PreREISE-JbwmpG5i
Installing dependencies from Pipfile.lock (e86f55)…
Ignoring importlib-metadata: markers 'python_version < "3.8"' don't match your environment
Ignoring typing-extensions: markers 'python_version < "3.8"' don't match your environment
  ================================ 49/49 - 00:01:56
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
All dependencies are now up-to-date!
```

### Time estimate
5-15 minutes, depending on how deeply you want to dig into the details of the `pyproj` internals.


jon-hagg (approved):

rouille (approved):
Looks good

Issue #145, opened 2021-02-16
Tests are very slow due to big file downloads


danielolsen:
On a pretty good home internet connection, with nothing else going on:
- **prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py::test_download_flexibility_data** takes >2 minutes
- **prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py::test_download_demand_data** takes >1 minute
- no other tests take more than 10 seconds

Can we cut this time down somehow, perhaps by testing the download functionality on smaller files?

lanesmith218:
Yeah, the files are pretty large so downloading and extracting takes a while. I could write tests for the private functions (e.g., `_download_data`, `_extract_data`) instead. This should allow us to test the functionality on smaller files.

PR #148, opened 2021-02-17
test: make NREL EFS download tests faster

lanesmith218:
### Purpose
The goal of this PR is to speed up the NREL EFS data download and extraction tests. Due to the size of the .zip files, the tests take minutes to run (as opposed to the other tests that take seconds). See #145 for reference.

### What the code is doing
`test_get_efs_data.py` was updated to not test `download_demand_data` and `download_flexibility_data` individually, but rather to test the individual private functions (e.g., `_download_data`, `_extract_data`) that comprise them. While testing the private functions create some more tests, they are much faster because `_download_data` can be tested with a much smaller file (the same can't be done with `download_demand_data` and `download_flexibility_data` since they are created to only download and extract the large NREL EFS data sets).

### Where to look
All of the changes are contained within `PreREISE/prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py`.

### Time estimate
This should only take a few minutes.

### DIFF
--- a/prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
+++ b/prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
@@ -1,163 +1,214 @@
 import os
+import zipfile
 
 import pandas as pd
 from pandas.testing import assert_frame_equal
 from powersimdata.network.usa_tamu.constants.zones import abv2state
 
 from prereise.gather.demanddata.nrel_efs.get_efs_data import (
+    _check_electrification_scenarios_for_download,
+    _check_path,
+    _check_technology_advancements_for_download,
+    _download_data,
+    _extract_data,
     account_for_leap_year,
-    download_demand_data,
-    download_flexibility_data,
     partition_demand_by_sector,
     partition_flexibility_by_sector,
 )
 
 
-def test_download_demand_data():
-    # Download one of the EFS demand data sets
-    download_demand_data(es={"Reference"}, ta={"Slow"})
+def test_check_electrification_scenarios_for_download():
+    # Run the check
+    test_es = _check_electrification_scenarios_for_download(es={"All"})
 
+    # Specify the expected values
+    exp_es = {"Reference", "Medium", "High"}
+
+    # Compare the two sets
+    assert test_es == exp_es
+
+
+def test_check_technology_advancements_for_download():
+    # Run the check
+    test_ta = _check_technology_advancements_for_download(ta={"All"})
+
+    # Specify the expected values
+    exp_ta = {"Slow", "Moderate", "Rapid"}
+
+    # Compare the two sets
+    assert test_ta == exp_ta
+
+
+def test_check_path():
+    # Run check
+    test_fpath = _check_path(fpath="")
+
+    # Specify the expected file path
+    exp_fpath = os.getcwd()
+
+    # Compare the two file paths
+    assert test_fpath == exp_fpath
+
+
+def test_download_data():
     try:
-        # Load the downloaded EFS demand data set
-        df = pd.read_csv("EFSLoadProfile_Reference_Slow.csv")
-
-        # Access the columns
-        test_cols = list(df.columns)
-        exp_cols = [
-            "Electrification",
-            "TechnologyAdvancement",
-            "Year",
-            "LocalHourID",
-            "State",
-            "Sector",
-            "Subsector",
-            "LoadMW",
-        ]
-
-        # Compare the two values
-        assert len(test_cols) == len(exp_cols)
-
-        # Remove the downloaded EFS data set
-        os.remove("EFSLoadProfile_Reference_Slow.csv")
-
-    except FileNotFoundError:
-        # If the automated extraction did not work, check that the .zip file was created
-        assert os.path.isfile("EFSLoadProfile_Reference_Slow.zip")
-
-        # Remove the downloaded .zip file
-        os.remove("EFSLoadProfile_Reference_Slow.zip")
-
-
-def test_download_flexibility_data():
-    # Download one of the EFS flexibility data sets
-    download_flexibility_data(es={"Reference"})
-
-    try:
-        # Load the downloaded EFS flexibility data set
-        df = pd.read_csv("EFSFlexLoadProfiles_Reference.csv")
-
-        # Access the columns
-        test_cols = list(df.columns)
-        exp_cols = [
-            "Electrification",
-            "TechnologyAdvancement",
-            "Flexibility",
-            "Year",
-            "LocalHourID",
-            "State",
-            "Sector",
-            "LoadMW",
-        ]
-
-        # Compare the two values
-        assert len(test_cols) == len(exp_cols)
-
-        # Remove the downloaded EFS data set
-        os.remove("EFSFlexLoadProfiles_Reference.csv")
-
-    except FileNotFoundError:
-        # If the automated extraction did not work, check that the .zip file was created
-        assert os.path.isfile(
-            "EFS Flexible Load Profiles - Reference Electrification.zip"
+        # Download a file using _download_data
+        _download_data(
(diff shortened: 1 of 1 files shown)


prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
```diff
@@ -1,107 +1,138 @@
...
     partition_flexibility_by_sector,
 )
 
+# Create a dummy demand data set
+cont_states = sorted(set(abv2state) - {"AK", "HI"})
+dummy_demand_data = {
+    "Electrification": ["High"] * 4 * 48 * 8760,
+    "TechnologyAdvancement": ["Rapid"] * 4 * 48 * 8760,
+    "Year": [2030] * 4 * 48 * 8760,
+    "LocalHourID": sorted([i for i in range(1, 8761)] * 4 * 48),
```
danielolsen, line 25: Can `[i for i in range(1, 8761)]` be replaced with `list(range(1, 8761))`?

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
```diff
@@ -1,107 +1,138 @@
...
+    "State": sorted([i for i in cont_states] * 4) * 8760,
```
danielolsen, line 26: Same as above: can `[i for i in cont_states]` be replaced with `list(cont_states)`?

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
```diff
@@ -1,107 +1,138 @@
...
-        os.remove("EFSLoadProfile_Reference_Slow.zip")
-
+    # Check that the expected .zip file was downloaded
+    assert os.path.isfile("project_resstock_efs_2013.zip")
 
-def test_download_flexibility_data():
-    # Download one of the EFS flexibility data sets
-    download_flexibility_data(es={"Reference"})
+    # Remove the downloaded .zip file
+    os.remove("project_resstock_efs_2013.zip")
```
danielolsen, line 99: This should be the `finally` block corresponding to a `try` block around the `_download_data` and `assert` calls, so that the cleanup happens no matter what.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
```diff
@@ -1,107 +1,138 @@
...
+    "TechnologyAdvancement": ["Rapid"] * 4 * 48 * 8760,
+    "Flexibility": ["Base"] * 4 * 48 * 8760,
+    "Year": [2030] * 4 * 48 * 8760,
+    "LocalHourID": sorted([i for i in range(1, 8761)] * 4 * 48),
+    "State": sorted([i for i in cont_states] * 4) * 8760,
+    "Sector": ["Commercial", "Industrial", "Residential", "Transportation"] * 48 * 8760,
+    "LoadMW": [1, 2, 3, 4] * 48 * 8760,
+}
+dummy_flex_df = pd.DataFrame(data=dummy_flex_data)
+dummy_flex_df.to_csv("EFSFlexLoadProfiles_High.csv", index=False)
```
danielolsen, line 51: These lines should be encapsulated in a function and executed at test time, not import time.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
```diff
@@ -1,107 +1,138 @@
...
-        assert os.path.isfile(
-            "EFS Flexible Load Profiles - Reference Electrification.zip"
-        )
+    # Compare the two values
+    assert len(test_cols) == len(exp_cols)
 
-        # Remove the downloaded .zip file
-        os.remove("EFS Flexible Load Profiles - Reference Electrification.zip")
+    # Remove the downloaded EFS data set
+    os.remove("test_demand.csv")
```
danielolsen, line 132: Same comment as above: these should be part of a `finally` block.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Yes it can! I'll fix this.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Ah, thanks for this tip! I'll make that change.

danielolsen:
Tests are indeed much faster, thanks for this quick PR. I have some comments on structure but the functional changes look good.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Okay, I'll make sure these are included in their respective functions.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
lanesmith218: Done.

prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
```diff
@@ -1,67 +1,113 @@
...
+        * 8760,
+        "LoadMW": [1, 2, 3, 4] * 48 * 8760,
+    }
+    dummy_demand_df = pd.DataFrame(data=dummy_demand_data)
+    dummy_demand_df.to_csv("test_demand.csv", index=False)
+
+    # Create a .zip file of the dummy demand data set
+    with zipfile.ZipFile("test.zip", "w") as z:
+        z.write("test_demand.csv")
+    os.remove("test_demand.csv")
```
danielolsen, line 90: Please also enclose these file writes within a `try`/`except` block as well.

danielolsen: I see that `"test_demand.csv"` will get cleaned up either way below, does `"test.zip"` get covered as well?

lanesmith218: Yeah, `test.zip` should be cleaned up within `_extract_data`. I can add an `except FileNotFoundError` in between the `try` and `finally` blocks of `test_extract_data` in case something goes wrong though.

Can you explain why the above code should be within a `try`/`except` block? I don't expect that code to raise an error.

danielolsen: Ah, I see about the `test.zip` deletion.

I'm thinking about what happens if e.g. `test.zip` fails to be written (e.g. in case another program already has it open). Then we get an exception, and `"test_demand.csv"` is not cleaned up. I guess it's not too big of a problem though, since whoever is running the tests will hopefully fix the problem and re-run and then the file should get cleaned up properly.

danielolsen (approved):
Tests are ~10x faster and seem to still test all the main functions, thanks!

PR #150, opened 2021-02-23
fix: set attribute on wrapped function so it actually works

jon-hagg:
### Purpose
We set an attribute on the function to retry which, initially for debugging, also enabled continuation in downloading wind data despite reaching the max attempts. However, it needs to be set on the wrapped function rather than the `func` object passed in, otherwise the decorated function won't be accessing the same value being incremented by the retry wrapper. 

### What the code is doing
Set the attribute on the right object, add a test to simulate the current usage.

### Testing
Added new unit test

### Time estimate
5 min

### DIFF
--- a/prereise/gather/request_util.py
+++ b/prereise/gather/request_util.py
@@ -69,16 +69,16 @@ def retry(
 
     def decorator(func):
         limiter = RateLimit(interval)
-        func.retry_count = 0
 
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
+            wrapper.retry_count = 0
             for i in range(max_attempts):
-                func.retry_count = i + 1
+                wrapper.retry_count = i + 1
                 try:
                     return limiter.invoke(lambda: func(*args, **kwargs))
                 except allowed_exceptions as e:
-                    if func.retry_count == max_attempts:
+                    if wrapper.retry_count == max_attempts:
                         print("Max retries reached!!")
                         if raises:
                             raise e
--- a/prereise/gather/tests/test_retry.py
+++ b/prereise/gather/tests/test_retry.py
@@ -50,3 +50,16 @@ def test_unhandled_exception():
 
     with pytest.raises(Exception):
         fail()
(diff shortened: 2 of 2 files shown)


danielolsen (approved):
It gets through the longer loops that had been giving me errors before, thanks.

prereise/gather/request_util.py
```diff
@@ -69,16 +69,16 @@ def retry(
 
     def decorator(func):
         limiter = RateLimit(interval)
-        func.retry_count = 0
 
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
+            wrapper.retry_count = 0
             for i in range(max_attempts):
```
BainanXia, line 76: Just curious about the logic of the for loop here, what is the idea behind it?

jon-hagg: Which part do you mean? The function attribute, overall retry logic, etc? Just want to make sure I answer the right question

BainanXia: Since we are not using the loop variable `i` here and it seems `wrapper.retry_count` also serves as a counter, I'm wondering why we need a `for` loop here rather than simply do `while wrapper.retry_count<=max_attempts`?

jon-hagg: Ah I see. The `while` loop would also work, but the distinction is the `wrapper.retry_count` variable is exposed to the decorated function (e.g. `download` in the `noaa_api` module) so while unusual, it could be modified which could break the loop. Compared to the `for` loop which is a bit more foolproof.

jon-hagg: The idea is that the user of the retry decorator shouldn't control the flow, that would defeat the point. The effect of exposing `wrapper.retry_count` does feel like leaking some details, but at least it's one-way (e.g. `download` doesn't modify the control flow defined by `retry`).

BainanXia: I see. Good to know. Thanks.

PR #152, opened 2021-02-26
ci: add packages workflow

jon-hagg:
### Purpose
Add workflow to regenerate Pipfile.lock, which can be triggered manually. 

### What the code is doing
Copy paste the workflow from PostREISE, and replaced the repo name.

### Testing
Tested in PostREISE

### Time estimate
3 min


rouille (approved):
Looks good

PR #157, opened 2021-03-25
ci: add pep8-naming to flake8 and fix warnings

jon-hagg:
### Purpose
Add pep8-naming check to flake8, so it is run by tox locally and in github. 
See https://github.com/Breakthrough-Energy/RenewableEnergyProject/issues/343 (not sure how else to link this)

### What the code is doing
In this case, we only had to ignore one warning, everything else is just switching to snake_case. 

### Testing
`tox -e flake8`

### Time estimate
5 min

### DIFF
--- a/prereise/gather/solardata/nsrdb/sam.py
+++ b/prereise/gather/solardata/nsrdb/sam.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pandas as pd
 import PySAM.Pvwattsv7 as PVWatts
-import PySAM.PySSC as pssc
+import PySAM.PySSC as pssc  # noqa: N813
 from powersimdata.network.usa_tamu.constants.zones import (
     abv2interconnect,
     id2abv,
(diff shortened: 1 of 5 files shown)


prereise/gather/solardata/nsrdb/sam.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import pandas as pd
 import PySAM.Pvwattsv7 as PVWatts
-import PySAM.PySSC as pssc
+import PySAM.PySSC as pssc  # noqa: N813
```
rouille, line 4: What is the issue here?

jon-hagg: We get the following warning:
`prereise/gather/solardata/nsrdb/sam.py:4:2: N813 camelcase 'PySAM.PySSC' imported as lowercase 'pssc'`

rouille (approved):
Looks good

danielolsen (approved):
This will be very helpful for our collaborators.

PR #159, opened 2021-03-31
chore: remove __all__ declarations

jon-hagg:
### Purpose
Remove unnecessary values from `__init__.py` files. This may also fix some warnings in the docs build.

### What the code is doing
Not setting the `__all__` variable.

### Testing
N/A 

### Time estimate
3 min


rouille (approved):
Thanks for cleaning up

PR #163, opened 2021-04-30
ci: add coverage report to pytest

jon-hagg:
### Purpose
Identical to Breakthrough-Energy/PowerSimData#471 except didn't add python 3.9 testing since pyproj is incompatible, so will address that separately.

One thing to note - I was having issues related to nrel-pysam when regenerating the lock file, so I upgraded to the latest, which is within the same major version. Haven't checked the impact of that yet.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,6 @@
+[run]
+branch = True
+omit =
+    **/test_*.py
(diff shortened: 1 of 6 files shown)


rouille (approved):
Thanks

PR #168, opened 2021-06-02
feat: add cli to download rap data

goccert25:

### Purpose
The larger goal of this change is to improve the UX of downloading data with PreREISE by creating a CLI. See [here](https://breakthrough-energy.github.io/docs/communication/project_ideas.html#project-1a-download-manager) for more details.

### What the code is doing
The code creates a CLI powered by argparse. Data on a per data source basis is consolidated in data_sources.py. This PR will only have usability for RAP as a MVP.

### Testing
Unit testing, also ran the CLI tool locally a couple of times

### Where to look
Everything is in a new /cli folder

### Usage Example/Visuals
`python -m prereise.cli.download.download_manager wind_data -r Texas -sd 2021-01-01 -ed 2021-01-02 -fp ./somedata` will download wind data from RAP for wind farms in Texas between the dates of 2021-01-01 to 2021-01-02 and pickle the data into the file `somedata`. 

### Time estimate
20 minutes

### DIFF
--- a/prereise/cli/data_sources.py
+++ b/prereise/cli/data_sources.py
@@ -1,0 +1,138 @@
+import logging
+from abc import ABCMeta, abstractmethod
+from datetime import datetime
+
+from powersimdata.input.grid import Grid
+
+from prereise.cli.constants import DATE_FMT
+from prereise.cli.helpers import validate_date, validate_file_path
+from prereise.gather.winddata.rap import rap
+
+
+class DataSource(metaclass=ABCMeta):
+    """Abstract class that defines a strict interface
+    that future data sources need to follow in order
+    to seamlessly integrate with the parser.
+
+    :param abc.ABCMeta metaclass: defaults to ABCMeta
+    """
+
+    @property
+    @abstractmethod
+    def command_name(self):
+        """String command name for argparse. See add_parser function
+        at `this link <https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers>`_.
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def command_help(self):
+        """String with a help description for the command. See add_parser function
+        at `this link <https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers>`_
+        for more details
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def extract_arguments(self):
+        """List of dictionaries defining the various flags and parameters
+        the user needs to pass into this command, which are
+        subsequently passed to the below extract function.
+        See `add_argument <https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument>`_
+        to understand what is acceptable/unacceptable in these
+        dictionaries.
+        """
+        pass
+
+    @abstractmethod
+    def extract(self):
+        """Function that takes in the arguments defined in
+        the above property extract_arguments and downloads
+        the appropriate data and saves it somewhere.
+        """
+        pass
+
+
+class WindData(DataSource):
+    @property
+    def command_name(self):
+        """See :py:func:`prereise.cli.data_sources.DataSource.command_name`
+
+        :return: (*str*)
+        """
+        return "wind_data"
+
+    @property
+    def command_help(self):
+        """See :py:func:`prereise.cli.data_sources.DataSource.command_help`
+
+        :return: (*str*)
+        """
+        return "Download wind data"
+
+    @property
+    def extract_arguments(self):
+        """See :py:func:`prereise.cli.data_sources.DataSource.extract_arguments`
+
+        :return: (*str*)
+        """
+        return [
+            {
+                "command_flags": ["--region", "-r"],
+                "required": True,
+                "choices": ["Texas", "Eastern", "Western"],
+                "type": str,
+                "action": "append",
+            },
+            {
+                "command_flags": ["--start_date", "-sd"],
+                "required": True,
+                "type": validate_date,
+                "help": "The start date for the data download in format 'YYYY-MM-DD'",
+            },
+            {
+                "command_flags": ["--end_date", "-ed"],
+                "required": True,
+                "type": validate_date,
+                "help": "The end date for the data download in format 'YYYY-MM-DD'",
+            },
+            {
+                "command_flags": ["--file_path", "-fp"],
+                "required": True,
+                "type": validate_file_path,
+                "help": "The file path to store the downloaded data. Must include filename "
+                "and .pkl file extension.",
+            },
+        ]
+
+    def extract(self, region, start_date, end_date, file_path, **kwargs):
+        """See :py:func:`prereise.cli.data_sources.DataSource.extract`
+
+        :param list region: list of regions to download wind farm data for
+        :param str start_date: date designating when to start the data download
+        :param str end_date: date designating when to end the data download
+        :param str file_path: file location on local filesystem on where to store the data
+        """
+        regions = list(set(region))
+        assert datetime.strptime(start_date, DATE_FMT) <= datetime.strptime(
+            end_date, DATE_FMT
+        )
+
+        grid = Grid(regions)
+        wind_farms = grid.plant.groupby("type").get_group("wind")
+        data, missing = rap.retrieve_data(
+            wind_farms, start_date=start_date, end_date=end_date
+        )
+        if len(missing) > 0:
+            logging.warning(f"There are {len(missing)} files missing")
+        data.to_pickle(file_path)
+
+
+def get_data_sources_list():
+    """Provides list of all data sources
+
+    :return: (*list*)
+    """
+    return [WindData()]
--- a/prereise/cli/download/tests/test_download_manager.py
+++ b/prereise/cli/download/tests/test_download_manager.py
@@ -1,0 +1,48 @@
+import pytest
+
+from prereise.cli.download.download_manager import generate_parser
+
+START_DATE = "2020-01-01"
+END_DATE = "2021-01-01"
+FILE_PATH = "./data.pkl"
+REGION = "Texas"
+
+WIND_DATA_ARG_LIST_FULL_FLAGS = [
+    "wind_data",
+    f"--start_date={START_DATE}",
+    f"--end_date={END_DATE}",
+    f"--file_path={FILE_PATH}",
+    f"--region={REGION}",
+]
+
+WIND_DATA_ARG_LIST_SHORT_FLAGS = [
+    "wind_data",
+    f"-sd={START_DATE}",
+    f"-ed={END_DATE}",
+    f"-fp={FILE_PATH}",
+    f"-r={REGION}",
+]
+
+
+@pytest.mark.parametrize(
+    "args", [WIND_DATA_ARG_LIST_FULL_FLAGS, WIND_DATA_ARG_LIST_SHORT_FLAGS]
+)
+def test_parser_wind_data(args):
+    parser = generate_parser()
+    result = parser.parse_args(args)
+    assert START_DATE == result.start_date
+    assert END_DATE == result.end_date
+    assert REGION in result.region
+    assert FILE_PATH == result.file_path
+
+
+@pytest.mark.parametrize(
+    "args", [WIND_DATA_ARG_LIST_FULL_FLAGS, WIND_DATA_ARG_LIST_SHORT_FLAGS]
+)
+def test_parser_wind_data_missing_flags(args):
+    parser = generate_parser()
+    for arg in args:
+        bad_arg_list = args.copy()
+        bad_arg_list.remove(arg)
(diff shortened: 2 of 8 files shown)


prereise/cli/data_sources.py
```diff
@@ -0,0 +1,67 @@
...
+            }
+        ]
+
+    def extract(self, region, start_date, end_date, file_path, **kwargs):
+        regions = list(set(region))
+        assert start_date <= end_date
+
+        grid = Grid(regions)
+        wind_farms = grid.plant.groupby("type").get_group("wind")
+        data, missing = rap.retrieve_data(
```
goccert25, line 52: Any advice on how `missing` should be handled?

goccert25:
Hey all, my code doesn't have comments/tests yet, but I wanted to open up the PR early to talk over the code and make sure it's headed in the right direction/avoid pitfalls early

prereise/cli/data_sources.py
jon-hagg: I don't have much to add to that summary, but thinking the simplest way to handle it is just output the number of missing files, or the percentage, and let the user decide afterward whether the data available is sufficient. In the [rap_demo](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/gather/winddata/rap/demo/rap_demo.ipynb) for example, that's all we really do, other than plotting the results which shows where the gaps are. 

Actually now that I think about it, I'd expect if the missing data is all sequential (so there is a large gap), it would be harder to impute the data. But there isn't an obvious threshold, so still think handling it programmatically might not be worth it. In this case though, maybe printing a full list of which days are missing in addition to the percentage might give the user more context. Wouldn't get more complex than this since most people would probably just rerun it until it works.

prereise/cli/helpers.py
```diff
@@ -0,0 +1,46 @@
...
+    """Helper function to validate file paths
+
+    :param str file_path: path include filename for where
+        to save the file
+    :raises Exception: exception telling user to pick a
+        valid file path
+    :return: (*str*) validated file path
+    """
+    if not os.access(os.path.dirname(file_path), os.W_OK):
+        raise Exception("Please choose a valid file path")
```
jon-hagg, line 25: I would use `ValueError` here, and for the docstring suggest something like `:raises ValueError: if the provided file path is not accessible`

requirements.txt
```diff
@@ -9,6 +9,7 @@ pyproj==2.6.1.post1
 python-dateutil==2.7.5
 requests~=2.24.0
 scipy==1.4.1
+tox==3.23.1
```
jon-hagg, line 12: I would leave this out since tox is usually part of the global environment - this avoids some redundant installation and enables running `tox` whether or not a virtualenv is activated.

prereise/cli/tests/test_helpers.py
```diff
@@ -0,0 +1,56 @@
...
+    data_source_mock.command_help = "command_help"
+    data_source_mock.extract_arguments = [
+        {"command_flags": ["--command"], "keyarg": "value"}
+    ]
+    data_source_mock.extract = MagicMock()
+    return data_source_mock
+
+
+def test_validate_date():
+    validate_date("2020-01-03")
```
jon-hagg, line 25: For completeness, we could also assert that it returns the value passed in. Similarly for `test_validate_file_path`

goccert25: Thanks! Fixed

prereise/cli/helpers.py
goccert25: Thanks! Fixed

requirements.txt
goccert25: Thanks! Fixed

prereise/cli/data_sources.py
```diff
@@ -0,0 +1,130 @@
...
+from powersimdata.input.grid import Grid
+
+from prereise.cli.helpers import validate_date, validate_file_path
+from prereise.gather.winddata.rap import rap
+
+
+class DataSource(metaclass=ABCMeta):
+    """Abstract class that defines a strict interface
+        that future data sources need to follow in order
+        to seamlessly integrate with the parser.
```
rouille, line 13: The two lines above should not be indented

goccert25: updated! thanks

prereise/cli/data_sources.py
```diff
@@ -0,0 +1,130 @@
...
+    def extract(self, region, start_date, end_date, file_path, **kwargs):
+        """See :py:func:`prereise.cli.data_sources.DataSource.extract`
+
+        :param list region: list of regions to download wind farm data for
+        :param str start_date: date designating when to start the data download
+        :param str end_date: date designating when to end the data download
+        :param str file_path: file location on local filesystem on where to store the data
+        """
+        regions = list(set(region))
+        assert start_date <= end_date
```
ahurli, line 113: Since the `strptime` format check doesn't require a leading zero, this has the potential to produce some weird errors where `-sd 2021-5-1 -ed 2021-12-1` fails this assertion even though they are both valid dates and the start date is before the end date.

prereise/cli/data_sources.py
```diff
@@ -0,0 +1,130 @@
...
+        return [
+            {
+                "command_flags": ["--region", "-r"],
+                "required": True,
+                "choices": ["Texas", "Eastern", "Western"],
+                "type": str,
+                "action": "append",
+            },
+            {
+                "command_flags": ["--start_date", "-sd"],
```
ahurli, line 88: It might be helpful to include help comments about the format here (i.e. `YYYY-MM-DD`). You can reference the [parser](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/pyreisejl/utility/parser.py) in `REISE.jl` for an example.

prereise/cli/data_sources.py
```diff
@@ -0,0 +1,130 @@
...
+                "required": True,
+                "type": validate_date,
+            },
+            {
+                "command_flags": ["--end_date", "-ed"],
+                "required": True,
+                "type": validate_date,
+            },
+            {
+                "command_flags": ["--file_path", "-fp"],
```
ahurli, line 98: Similar to the above comment, it would be nice to specify that this filepath needs to include the filename and that the resulting file will be a pickle file so the user can use the appropriate extension (to keep consistent with the rest of the codebase, we use `.pkl` as the pickle file extension).

prereise/cli/helpers.py
```diff
@@ -0,0 +1,45 @@
...
+def validate_file_path(file_path):
+    """Helper function to validate file paths
+
+    :param str file_path: path include filename for where
+        to save the file
+    :raises ValueError: if the provided file path is not accessible
+    :return: (*str*) validated file path
+    """
+    if not os.access(os.path.dirname(file_path), os.W_OK):
+        raise ValueError("Please choose a valid file path")
```
ahurli, line 24: We can also make sure that `file_path` is not a directory (fails with an `IsADirectoryError` after all the data has been downloaded). It might also be helpful to have a warning or confirmation prompt if the file already exists and will be overwritten.

ahurli (commented):
Thanks for contributing!

prereise/cli/data_sources.py
goccert25: great catch, updated to use strptime at this assertion point as well

prereise/cli/data_sources.py
goccert25: Thanks for pointing me to that! didn't know it existed

prereise/cli/data_sources.py
goccert25: Done

prereise/cli/helpers.py
goccert25: will do, added additional tests as well

prereise/cli/tests/test_data_sources.py
```diff
@@ -0,0 +1,89 @@
...
+        [TEXAS_REGION_LIST[0], TEXAS_REGION_LIST[0]],
+        STRING_DATE_2021_5_1,
+        STRING_DATE_2021_12_1,
+        CURRENT_DIRECTORY_FILEPATH,
+    )
+    grid.assert_called_with(TEXAS_REGION_LIST)
+
+
+@patch("prereise.cli.data_sources.WindData")
+def test_get_data_sources_list(wind_data_mock):
```
jon-hagg, line 86: Any reason not to test the call directly here? E.g. `assert get_data_sources_list() == [WindData()]`. I tend to limit mocking to cases where there is an external dependency (or any nondeterministic behavior).

goccert25: Good point. I think when I initially wrote up the test I wanted to assert equality of two lists, and it was easier to do that with a mock, but just checking that the list value is an instance of WindData is probably better

goccert25: updated!

prereise/cli/download/tests/test_download_manager.py
```diff
@@ -0,0 +1,12 @@
+from unittest.mock import MagicMock, patch
+
+from prereise.cli.download.download_manager import generate_parser
+
+
+@patch("prereise.cli.download.download_manager.get_data_sources_list")
+@patch("prereise.cli.download.download_manager.add_data_source_to_download_parser")
+def test_generate_parser(add_data_source_to_download_parser, get_data_sources_list):
```
jon-hagg, line 8: I would consider leaving this one out - whether or not something is called should ideally be an implementation detail that a test shouldn't know about. Note, I'd consider this distinct from the examples in `test_data_sources.py` since those involve testing the effects of the function being called.

One thing that might be nice is to test the parser itself. Here is an example I put together, feel free to add it or make changes if you like.
```
def test_parser_wind_data():
    parser = generate_parser()
    start_date = "2020-01-01"
    end_date = "2021-01-01"

    arglist = [
        "wind_data",
        f"--start_date={start_date}",
        f"--end_date={end_date}",
        "--file_path=./data",
        "-r=Texas",
    ]
    result = parser.parse_args(arglist)
    assert start_date == result.start_date
    assert end_date == result.end_date

```

jon-hagg:
Ok I think I've made all the comments I plan on. Thanks for the great work and responsiveness!

prereise/cli/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+from datetime import datetime
+
+from prereise.cli.constants import DATE_FMT
+
+
+def validate_date(date_string):
+    """Helper function to validate date strings
+
+    :param str date_string: date in string format to be validated
+    :return: (*str*) validated date string
```
rouille, line 12: We use the following convention for the return parameter: `:return: (*type*) -- description`
So here it would be:  `:return: (*str*) -- validated date string`

prereise/cli/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+    return date_string
+
+
+def validate_file_path(file_path):
+    """Helper function to validate file paths
+
+    :param str file_path: path include filename for where
+        to save the file
+    :raises ValueError: if the provided file path is not accessible
+    :return: (*str*) validated file path
```
rouille, line 24: Same comment here: `:return: (*str*) -- validated file path`

prereise/cli/helpers.py
```diff
@@ -0,0 +1,52 @@
...
+    return file_path
+
+
+def add_data_source_to_download_parser(data_source, parser):
+    """Helper function that adds an additional source of data
+    that the user can interact with and download via the
+    command line tool
+
+    :param class:`prereise.cli.data_sources.DataSource` data_source: an
+        instance of a class that implements the abstract class DataSource
```
rouille, line 41: Sphinx is ale to parse `:param prereise.cli.data_sources.DataSource data_source: description` and this is what we use throughout the code base

goccert25: Great. Thanks for the catch. I wonder if there's a linter that could be added that checks/lints these docstrings, I feel like that would save some work hunting malformed docstrings 😅

prereise/cli/download/tests/test_download_manager.py
goccert25: nice suggestion, i just realized i don't really have many tests validating the actual parsing - was testing that mostly manually locally via terminal. Added tests for those and some 👍

rouille:
My last comment would be to write a code snippet in the README. I would say under section 2.A.

prereise/cli/download/tests/test_download_manager.py
jon-hagg: Nice, looks good!

prereise/cli/helpers.py
rouille: That would be great. I have seen some linter for the Google and Numpy style. For some rason we decided to use the more condensed Sphinx style.

jon-hagg (approved):

rouille (approved):
Works like a charm. Thanks.

PR #169, opened 2021-06-02
Use integration marker for test fetching data

rouille:

### Purpose
Use `pytest` marker to deselect test fetching third party data

### What the code is doing
There is no code

### Testing
Test is deselected when `tox pytest-local` (equivalent to `pytest -m 'not integration'`) is invoked

### Where to look
I imported `pytest` and mark `test_download_data` as being an integration test

### Usage Example/Visuals
N/A

### Time estimate
2min

### DIFF
--- a/prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
+++ b/prereise/gather/demanddata/nrel_efs/tests/test_get_efs_data.py
@@ -2,6 +2,7 @@ import os
 import zipfile
 
 import pandas as pd
+import pytest
 from pandas.testing import assert_frame_equal
 from powersimdata.network.usa_tamu.constants.zones import abv2state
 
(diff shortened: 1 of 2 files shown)


jon-hagg (approved):
Thanks

lanesmith:
Looks good to me @rouille, thanks for doing this. Is this a procedure that we should follow moving forward (i.e., include a similar marker for tests that access external data)?

rouille:
> Looks good to me @rouille, thanks for doing this. Is this a procedure that we should follow moving forward (i.e., include a similar marker for tests that access external data)?

I don't think we have a clear guideline but tests requiring access to the server have the **ssh** marker and those that use external dependencies have the **integration** marker.

lanesmith:
> I don't think we have a clear guideline but tests requiring access to the server have the **ssh** marker and those that use external dependencies have the **integration** marker.

Great, thank you for clarifying!

lanesmith (approved):

rouille:
I have updated the description of the marker to match what we have in **PowerSimData**

PR #177, opened 2021-06-25
chore: update packages manually

jon-hagg:
### Purpose
Manually update packages to add networkx. The workflow failed since the current scipy version is incompatible with constraints in networkx, so I just bumped that and ran `pipenv lock --pre`. See the [logs](https://github.com/Breakthrough-Energy/PreREISE/runs/2916529401?check_suite_focus=true#step:5:50) for the error. 

### What the code is doing
Updated Pipfile and requirements.txt. Regnerate Pipfile.lock.

### Testing
CI

### Time estimate
3 min


rouille (approved):
Thanks

PR #179, opened 2021-06-28
Report code coverage

jon-hagg:
### Purpose
Plumbing for reporting code coverage to codecov.io.

### What the code is doing
* Add separate github workflow, to run on merges to develop
* Parameterize `pytest-local` tox env so we can specify the xml format required by codecov
* Add dependencies label to packages.yml (unrelated, but trivial)
* Include tests in the coverage measurement
* add some missing badges

### Testing
Ran the workflow to initialize the repository on codecov and generate the badge.

### Time estimate
5 min

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,2 @@
 [run]
 branch = True
-omit =
-    **/test_*.py
-    **/__init__.py
-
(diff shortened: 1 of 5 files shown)


rouille (approved):
Thanks

PR #180, opened 2021-06-29
feat: map bus to balancing authority

ywang303:
### Purpose
Map each bus in an area to a balancing authority

### What the code is doing
map a list of bus gis (ie. list of latitude& longitude associated with buses) to counties, and then map those counties to balancing authorities. The returned bus correspondence will be in the same length of as the of passed bus gis. The returned BA value is NaN for bus_id that fails to find a correspondence from map_buses_to_county().

### Where to look
The method is implemented under prereise/gather/demanddata/eia/map_ba.py 

This method reads in BA_County_map.json file for BA_County_Map information. And therefore, should have BA_County_map.json under directory prereise/gather/demanddata/eia/

### Testing
Unit test and integration test

### Usage Example/Visuals
from powersimdata.scenario.scenario import Scenario
from prereise.gather.demanddata.eia.map_ba import map_buses_to_ba
ref_scenario = Scenario(603) 
west = ref_scenario.state.get_grid()
bus_gis=USA.bus[['lat','lon']].iloc[0:2,] # two gis from Maine state
map_buses_to_ba(bus_gis)
Screen Shot 2021-07-01 at 5 41 54 PM 
the matched BA 'ISNE' is correct for the two Maine gis.

# pass a fake gis of Shanghai
data=[[31.2304,121.4737]] 
fake_gis=pd.DataFrame(data, columns = ['lat', 'lon'])
map_buses_to_ba(fake_gis)
Screen Shot 2021-07-01 at 6 17 12 PM
the method returns NaN as a BA value.

### Notification
The time estimated to finish mapping the entire USA grid is longer than 2 hours, please remember to cache the result into a local csv file the time you call the method.

### Time estimate for review
<30 min


BainanXia:
- It seems there is something wrong with the PR description, which needs a revision. One could click "Preview" tag to check what it looks like before submit and it is always editable.
- We usually put data files in a data folder, which we already have one in the root folder of this repo.

ywang303:
put json file under data folder now. And revise the description format.

BainanXia:
> put json file under data folder now. And revise the description format.

What I could see is 
- two same files in different directories are committed to this PR.
- all checks have failed.

ywang303:
looks good now.

prereise/gather/demanddata/eia/map_ba.py
```diff
@@ -95,3 +95,32 @@ def map_buses_to_county(bus_county_map):
...
+
+    ba2county = pd.DataFrame(
+        [[k, v] for k, a in ba_county_map.items() for v in a], columns=["BA", "County"]
+    )
+
+    return (
+        b2c_key.reset_index()
+        .merge(ba2county, how="left", on=["County"], suffixes=("", ""))
+        .set_index("bus_id")
+    )
```
BainanXia, line 126: Instead of transforming dictionary into a data frame then trying to merge two data frames based on "County" column, we could simply reverse the dictionary (keys=County, values=BA) then `apply/map` the dictionary based on the "County" column like following:
```python
county_ba_map = {
    county: ba
    for ba, county_list in ba_county_map.items()
    for county in county_list
}
b2c_key["BA"] = b2c_key["County"].map(county_ba_map)
```
In this case, we don't need to drop column "BA" in l.110

BainanXia: Again, county name mismatches will kick in at this step if there is any, i.e. the county names returned by `map_buses_to_county` may be in a different format as we have in `county_ba_map` (this is something we need to check, I'm not sure. At least, the county names are from two different sources, i.e. census API vs BA's internal documentations).

BainanXia (commented):
In general, we need both unit tests and integration tests for such function
- unit tests: we may add a simple test in `prereise/gather/demanddata/eia/tests/test_map_ba.py` for this function by feeding it designed inputs, such as a fake bus with lat, lon in New York and check if it returns the expected BA, i.e. NYISO.
- integration test: this doesn't need to go into the code base, but I think we would like to confirm all buses in our USA grid is mapped to a BA successfully. We can post the test results in the `Testing` section of the PR description.

Also, it is worth mentioning what is the computational performance of the procedure. If it takes quite some time to generate the mapping for all buses to BA, should we cache the resultant data frame in a csv, either in codebase or blob storage?

prereise/gather/demanddata/eia/map_ba.py
```diff
@@ -95,3 +95,32 @@ def map_buses_to_county(bus_county_map):
...
+    """Find the Balancing Authority in the U.S. territory that each bus in the query grid
+    belongs to.
+
+    :param (*pandas.DataFrame*) bus_gis: data frame contains a list of
+        entries with lat and long of buses.
+    :return: (*pandas.DataFrame*) data frame of BA that buses
+        locate
+    """
+
+    b2c_key = map_buses_to_county(bus_gis)[0].iloc[:, :-1].dropna()
```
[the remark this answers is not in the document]
ywang303: These are ignored because we want the calculated metrics to be valid (not NaN values) so that it can be compared to the same bus to BA metrics from Plexos system. I will add this explanation on documentation

prereise/gather/demanddata/eia/map_ba.py
```diff
@@ -95,3 +95,32 @@ def map_buses_to_county(bus_county_map):
...
+
+    # read json file for BA_County Map
+    import json
```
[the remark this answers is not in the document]
ywang303: Will revise this.

prereise/gather/demanddata/eia/map_ba.py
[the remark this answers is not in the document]
ywang303: Do you mean write to a csv file as a return value? and put it under data folder?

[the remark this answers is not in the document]
ywang303: resolved by changes the matching algorithm as suggested and add explanation for possible NaN values in the function return.

prereise/gather/demanddata/eia/map_ba.py
```diff
@@ -95,3 +97,30 @@ def map_buses_to_county(bus_county_map):
...
+
+    # read json file for BA_County Map
+    with open("../../data/BA_County_map.json") as f:
+        ba_county_map = json.load(f)
+
+    ba2county = pd.DataFrame(
+        [[k, v] for k, a in ba_county_map.items() for v in a], columns=["BA", "County"]
+    )
+
+    county_ba_map = ba2county.set_index("County").to_dict("dict")["BA"]
```
BainanXia, line 122: I think in this way, we don't need `ba2county` here, right? We can simply reorganize the dict `ba_county_map` to come up with the `county_ba_map`. Then we don't need to drop last "BA" column in b2c_key data frame, we can simply do `b2c_key["BA"] = b2c_key["County"].map(county_ba_map)` or can `return b2c_key["County"].map(county_ba_map)` directly if we don't care about the rest info in the data frame, or can `return b2c_key` since we modify everything inplace.

BainanXia:
Thanks for the notification added into the description. It seems it takes hours to generate the mapping for all buses in the US system. Thus it is worth caching the result csv in the data folder. Also, a sanity check on the resultant csv will be helpful, i.e. whether the number of buses we feed in the function equals to the number of buses in the output? Are there any NaN values in the "BA" column? The sanity check or integration test doesn't need to be in the code, but it is good to post it in the description as well.

prereise/gather/data/BA_County_map.json
```diff
@@ -0,0 +1 @@
... Oak__TX", "Llano__TX", "Loving__TX", "McCulloch__TX", "McLennan__TX", "McMullen__TX", "Madison__TX", "Martin__TX", "Mason__TX", "Matagorda__TX", "Maverick__TX", "Medina__TX", "Menard__TX", "Midland__TX", "Milam__TX", "Mills__TX", "Mitchell__TX", "Motley__TX", "Nacogdoches__TX", "Navarro__TX", "Nolan__TX", "Nueces__TX", "Ochiltree__TX", "Oldham__TX", "Palo Pinto__TX", "Parker__TX", "Pecos__TX", "Presidio__TX", "Rains__TX", "Reagan__TX", "Real__TX", "Reeves__TX", "Refugio__TX", "Roberts__TX", "Robertson__TX", "Rockwall__TX", "Runnels__TX", "Rusk__TX", "San Patricio__TX", "San Saba__TX", "Schleicher__TX", "Scurry__TX", "Shackelford__TX", "Somervell__TX", "Starr__TX", "Stephens__TX", "Sterling__TX", "Stonewall__TX", "Sutton__TX", "Swisher__TX", "Tarrant__TX", "Taylor__TX", "Terrell__TX", "Throckmorton__TX", "Titus__TX", "Tom Green__TX", "Travis__TX", "Upton__TX", "Uvalde__TX", "Val Verde__TX", "Van Zandt__TX", "Victoria__TX", "Waller__TX", "Ward__TX", "Washington__TX", "Webb__TX", "Wharton__TX", "Wheeler__TX", "Wichita__TX", "Wilbarger__TX", "Willacy__TX", "Williamson__TX", "Wilson__TX", "Winkler__TX", "Wise__TX", "Young__TX", "Zapata__TX", "Zavala__TX"], "SCL_TPWR": ["Thurston__WA"]}
```
danielolsen, line 1: Do we want to format this as if it were a python dict?

prereise/gather/demanddata/eia/map_ba.py
ywang303: modified it.

prereise/gather/data/BA_County_map.json
BainanXia: It is a python dict if I remember correctly. You are suggesting we put it in a `const.py` as a dict explicitly?

danielolsen: `BA_County_map.json`. That's why the linter does not complain.

BainanXia: Yes. The linter won't go through a `json` file.

danielolsen: It could be in a `const.py`, or a `ba_county_map.py`, or in `BA_County_map.json` but formatted better.

Unrelated, but what is the source of this data? Do we need to go through the data intake procedure?

BainanXia: I generated this `json` file simply by dumping a python `dict` to it. Details see `Dropbox (Gates Ventures)/BreakthroughEnergySciences/Explorations/BainanX/external connection requests/County to BA map`
If we go with `json`, how should we reformat it.

This is actually the output of my script + manual process when addressing an external request last year, again details could be found in the file directory above. Short answer, the raw data can't be redistributed and we use the data we generated here, thus no data intake need to be conducted

danielolsen: `black` can be used to manually reformat this file as if it were python, and it should still be valid JSON I think.

BainanXia: That's good to know. Will see how it works.

BainanXia: There is an issue for using black to reformat json file: 
image

BainanXia: Update: it works after I manually fix the issue above.

BainanXia:
Close this with #181

PR #184, opened 2021-07-08
feat: add ability to select grid model to cli

goccert25:

### Purpose
Add ability to select grid_models to cli in anticipation of additional grid_models being added. This is a larger effort around the download manager project see here https://breakthrough-energy.github.io/docs/communication/project_ideas.html#project-1a-download-manager

### What the code is doing
How is the purpose executed?
Adds additional flags that allows user to specify grid model when calling the cli tool

### Testing
Existing and additional unit tests 

### Where to look
Main new code added are flags in solar_data.py and wind_data.py. Everything else supports those new additions

### Usage Example/Visuals
N/A

### Time estimate
around 15 minutes


prereise/cli/constants.py
```diff
@@ -3,4 +3,6 @@
 START_DATE_HELP_STRING = "The start date for the data download in format 'YYYY-MM-DD'"
 END_DATE_HELP_STRING = "The end date for the data download in format 'YYYY-MM-DD'"
 FILE_PATH_HELP_STRING = "The file path to store the downloaded data. Must include filename and .pkl file extension."
+GRID_MODEL_HELP_STRING = "A .mat file path for a grid model or a string supported by powersimdata.input.grid.Grid.SUPPORTED_MODELS"
```
rouille, line 6: While it is true that the `source` parameter of the `Grid` class can possibly take a .mat file path I don't believe we will ever generate a profile for a scenario grid enclosed in one of this file. We save the grid of a scenario in a mat file at the end of the run and the this transformed grid is recovered in the analysis phase by instantiating the `Grid` class with the `source` parameter set to the path of the .mat file. The transformed profiles are obtained from the original  profiles using the transformed grid and change table of the scenario.

Anyway, I believe the message could be simplified to `A string supported by powersimdata.input.grid.Grid.SUPPORTED_MODELS`

What do you think @danielolsen and @BainanXia

BainanXia: Agree. We never generate profiles based on a modified grid directly, instead, the modified profiles are generated by base profiles + change table. In the future, if we add the capability of generating a change table from a modified grid comparing with base grid or generating profiles based on a modified grid, we might allow specifying a `.mat` file path to do so.

goccert25: Great, thanks @BainanXia and @rouille for the context. Should I only update the help string? Or should I remove the ability of the user to specify a .mat file all together? I'd personally prefer the latter as it makes the code much simpler and makes any upcoming refactor of the `Grid` class easier, but open to both

rouille: How would you implement it? We still need to be able to create a `Grid` object from a .mat file for other purposes.

jon-hagg: I assume we are only talking about limiting the values for this cli to named grid models. In any case, couple ideas:
1) modify the help string so that it pulls the values from `Grid.SUPPORTED_MODELS`
2) depending on preferences, we could remove the `.mat` check from validation here, so the cli is more strict than the grid constructor itself, or remove the validation altogether, allowing the grid class to handle it. The latter suggests to a user that only `SUPPORTED_MODELS` should be used, but if a `.mat` file path is provided, it would still work.

goccert25: Yeah @jon-hagg was right, when I said "should I remove the ability of the user to specify a .mat file all together" I was talking about removing the ability of the CLI to accept a .mat file, not the ability of the `Grid` class to accept a .mat file. From what this discussion sounds like, seems as though not allowing .mat files to be passed into the CLI is perfectly fine, so I'll update the code to only check for the model being in `SUPPORTED_MODELS`

goccert25:
updated, code becomes much simpler! :)

jon-hagg (approved):

PR #188, opened 2021-07-09
chore: update packages with dependency on urllib3

jon-hagg:
### Purpose
Fix the dependabot warning for urllib3. To do that we need to update top level packages `h5pyd` and `requests` which both currently depend on it, and have a maximum version constraint. Note: this is why the github action has no effect on urllib3 version.

### What the code is doing
Update the packages mentioned above, as well as bump numpy to be consistent with powersimdata. Unfortunately, h5pyd now requires a bunch of google auth packages that we don't need, but not much we can do about that. Also, updating pandas breaks a couple of tests, so I will defer that for another PR. 

### Testing
tox

### Where to look
Note that the urllib3 in Pipfile.lock is now an acceptable version.

### Time estimate
5 min


rouille (approved):
Thanks

jon-hagg:
Forgot to update the requirements.txt

PR #189, opened 2021-07-11
fix: add imputation of missing wind data for cli

goccert25:

### Purpose
Make cli download of wind data on par with example notebook. This is a part of the download manager project https://breakthrough-energy.github.io/docs/communication/project_ideas.html#project-1a-download-manager 

### What the code is doing
How is the purpose executed?
Adding a function call to `impute.gaussian`

### Testing
unit testing

### Time estimate
1 minute

### DIFF
--- a/prereise/cli/data_sources/wind_data.py
+++ b/prereise/cli/data_sources/wind_data.py
@@ -14,7 +14,7 @@ from prereise.cli.constants import (
 )
 from prereise.cli.data_sources.data_source import DataSource
 from prereise.cli.helpers import validate_date, validate_file_path
-from prereise.gather.winddata.rap import rap
+from prereise.gather.winddata.rap import impute, rap
 
 
 class WindDataRapidRefresh(DataSource):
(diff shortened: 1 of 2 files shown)


prereise/cli/data_sources/wind_data.py
```diff
@@ -85,4 +85,7 @@ def extract(self, region, start_date, end_date, file_path, **kwargs):
         )
         if len(missing) > 0:
             logging.warning(f"There are {len(missing)} files missing")
+
+        # Imputing any missing data in place
+        impute.gaussian(data, wind_farms, inplace=True)
```
goccert25, line 90: @rouille is this what you had in mind (for completing the wind data download from RAP)?

rouille: Yes. I would say it can be moved within the `if` statement with a message stating that imputation of missing data will be performed

rouille: Perhaps, we can have in the command line a boolean `impute` argument that will trigger the imputation of missing data (True, default) or not (False) just in case the user has a homemade algorithm he/she/they would rather apply

goccert25: good point, done.

rouille (approved):
LGTM

BainanXia (approved):
Thanks!

PR #192, opened 2021-07-13
Load zone file

rouille:

### Purpose
Add file enclosing information on load zone in a newly created data folder

### What the code is doing
N/A

### Testing
N/A

### Where to look
The file is located in **prereise/gather/griddata/hifld/data/**

### Usage Example/Visuals
N/A

### Time estimate
2min

### DIFF
--- a/prereise/gather/griddata/hifld/data/zone.csv
+++ b/prereise/gather/griddata/hifld/data/zone.csv
@@ -1,0 +1,55 @@
+zone_id,zone_name,state,interconnect,time_zone
+1,Alabama,Alabama,Eastern,ETC/GMT+6
+2,Arkansas,Arkansas,Eastern,ETC/GMT+6
+3,Arizona,Arizona,Western,ETC/GMT+7
+4,California,California,Western,ETC/GMT+8
+5,Colorado,Colorado,Western,ETC/GMT+7
+6,Connecticut,Connecticut,Eastern,ETC/GMT+5
+7,Delaware,Delaware,Eastern,ETC/GMT+5
+8,Florida,Florida,Eastern,ETC/GMT+6
+9,Georgia,Georgia,Eastern,ETC/GMT+5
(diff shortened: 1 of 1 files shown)


rouille:
Note that New Mexico *fully* belongs to the Western Interconnection.

danielolsen:
> Note that New Mexico _fully_ belongs to the Western Interconnection.

This is not consistent with reality.

BainanXia:
> Note that New Mexico _fully_ belongs to the Western Interconnection.

Based on the discussion [here](https://besciencescommunity.slack.com/archives/G01CC0B6139/p1626203626037900), a temporal solution for this is 
- We include another Eastern NM entry
- We add a clean up step for the output of this file excluding it to make it consistent with the current behavior.

BainanXia:
I have another concern: we have same zone_names with different zone_ids in this mapping, however, in our current framework, they are used interchangeable. This might put us in trouble when this model is interacting with the rest of the modules, especially for the functions in PostREISE.

danielolsen:
> I have another concern: we have same zone_names with different zone_ids in this mapping, however, in our current framework, they are used interchangeable. This might put us in trouble when this model is interacting with the rest of the modules, especially for the functions in PostREISE.

I think you are right, there are probably several places in the code that assume a 1-to-1 mapping of zone_id to zone_name and vice versa. Luckily this should be easy enough to fix, following the model of what's in PowerSimData now for the TAMU network: https://github.com/Breakthrough-Energy/PowerSimData/blob/021cb90279f5f57170d830673814a043b54e956f/powersimdata/network/usa_tamu/constants/zones.py

rouille:
The new version of the file has 2 Montana, 2 New Mexico, 2 South Dakota and 4 Texas. The `zone_name` column is unique now.

danielolsen (approved):
Thanks

PR #193, opened 2021-07-13
Feat: Load and clean/remap initial data

ahurli:

### Purpose
Load in initial csv's, then cleanup and remap some of the data for use in future calculations.

### What the code is doing
The entry point is in `prereise.gather.griddata.hifld.model.py`, specifically the `get_grid_data` function. This is roughly analogous to the `data_transform` function called in the `main` of `data_trans.py` from the original `HIFLD` PR.

### Testing
Compared the results from these functions to analogous functions in the original HIFLD PR.

### Where to look
In the `prereise/gather/griddata/hifld` folder, there is a new `model.py` to load in the grid data.
In the `data_access` subfolder, there is a new `load.py` to load in the csv's.
In the `calculate` subfolder, there is both `remap.py` to reformat data into another usable form (i.e. converting dataframes to dictionaries) and `clean.py` to filter and clean data.

### Usage Example/Visuals
N/A

### Time estimate
10 min

### DIFF
--- a/prereise/gather/griddata/hifld/calculate/remap.py
+++ b/prereise/gather/griddata/hifld/calculate/remap.py
@@ -6,8 +6,9 @@ def get_zone_mapping(zone):
 
     :param pandas.DataFrame zone: the zone.csv data
     :return: (*tuple*) -- a tuple of two dictionaries, where the first
-    dictionary is a dict mapping the STATE to its ID and the second is a dict
-    mapping a tuple of state and interconnect to its ID.
+        dictionary is a dict mapping the zone_name to its zone_id and the
+        second is a dict mapping a tuple of state and interconnect to its
+        zone_name.
     """
 
     zone_dic = {}
@@ -22,7 +23,7 @@ def get_zone_mapping(zone):
 def get_line_mapping(lines):
     """Create dict to store all the raw transmission line csv data
 
-    :param str t_csv: path of the HIFLD transmission csv file
+    :param str lines: path of the HIFLD transmission csv file
     :return: (*dict*) -- a dict mapping the transmission ID to its raw parameters.
     """
 
@@ -35,9 +36,11 @@ def get_line_mapping(lines):
 def get_sub_mapping(clean_data):
     """Generate the subs
 
-    :param str E_csv: path of the HIFLD substation csv file
-    :return: (*dict*) --  sub_by_coord_dict, a dict mapping from (x, y) to substation detail.
-    :return: (*dict*) --  sub_name_dict, a dict mapping from substation name to its coordinate (x, y).
+    :param str clean_data: cleaned HIFLD substation csv data
+    :return: (*tuple*) --  a tuple of two dictionaries, where the first
+        dictionary is sub_by_coord_dict, a dict mapping from (x, y) to
+        substation detail, and the second is sub_name_dict, a dict mapping from
+        substation name to its coordinate (x, y).
     """
 
     sub_by_coord_dict = {}
--- a/prereise/gather/griddata/hifld/calculate/remap.py
+++ b/prereise/gather/griddata/hifld/calculate/remap.py
@@ -13,12 +13,12 @@ def get_zone_mapping(zone):
         zone_name.
     """
 
-    zone_dic = pd.Series(zone.zone_id, index=zone.zone_name).to_dict()
-    zone_interconnect_dic = pd.Series(
+    zone = pd.Series(zone.zone_id, index=zone.zone_name).to_dict()
+    zone_interconnect = pd.Series(
         zone.zone_id, index=zip(zone.zone_name, zone.interconnect)
     ).to_dict()
 
-    return zone_dic, zone_interconnect_dic
+    return zone, zone_interconnect
 
 
 def get_line_mapping(lines):
@@ -39,26 +39,26 @@ def get_sub_mapping(clean_data):
 
     :param str clean_data: cleaned HIFLD substation csv data
     :return: (*tuple*) --  a tuple of two dictionaries, where the first
-        dictionary is sub_by_coord_dict, a dict mapping from (x, y) to
-        substation detail, and the second is sub_name_dict, a dict mapping from
-        substation name to its coordinate (x, y).
+        dictionary is sub_by_coord, a dict mapping from (x, y) to substation
+        detail, and the second is sub_name, a dict mapping from substation name
+        to its coordinate (x, y).
     """
 
-    sub_by_coord_dict = {}
-    sub_name_dict = {"sub": []}
+    sub_by_coord = {}
+    sub_name = {"sub": []}
     for index, row in clean_data.iterrows():
         location = (row["LATITUDE"], row["LONGITUDE"])
-        if location in sub_by_coord_dict:
+        if location in sub_by_coord:
             raise Exception(
                 f"WARNING: substations coordinates conflict check: {location}"
             )
-        sub_by_coord_dict[location] = (
+        sub_by_coord[location] = (
             row["ID"],
             row["NAME"],
             row["STATE"],
             row["COUNTY"],
         )
-        if row["NAME"] not in sub_name_dict:
-            sub_name_dict[row["NAME"]] = []
-        sub_name_dict[row["NAME"]].append((row["LATITUDE"], row["LONGITUDE"]))
-    return sub_by_coord_dict, sub_name_dict
+        if row["NAME"] not in sub_name:
+            sub_name[row["NAME"]] = []
+        sub_name[row["NAME"]].append((row["LATITUDE"], row["LONGITUDE"]))
+    return sub_by_coord, sub_name
--- a/prereise/gather/griddata/hifld/calculate/remap.py
+++ b/prereise/gather/griddata/hifld/calculate/remap.py
@@ -1,0 +1,16 @@
+def get_zone_mapping(zone):
+    """Generate dictionaries of zone using the zone.csv
+
+    :param pandas.DataFrame zone: the zone.csv data
+    :return: (*tuple*) -- a tuple of two dictionaries, where the first
+    dictionary is a dict mapping the STATE to its ID and the second is a dict
+    mapping a tuple of state and interconnect to its ID.
+    """
+
+    zone_dic = {}
+    zone_interconnect_dic = {}
+    for i in range(len(zone)):
+        tu = (zone["zone_name"][i], zone["interconnect"][i])
+        zone_dic[zone["zone_name"][i]] = zone["zone_id"][i]
+        zone_interconnect_dic[tu] = zone["zone_id"][i]
+    return zone_dic, zone_interconnect_dic
--- a/prereise/gather/griddata/hifld/calculate/remap.py
+++ b/prereise/gather/griddata/hifld/calculate/remap.py
@@ -1,5 +1,7 @@
 import copy
 
+import pandas as pd
+
 
 def get_zone_mapping(zone):
     """Generate dictionaries of zone using the zone.csv
@@ -11,12 +13,11 @@ def get_zone_mapping(zone):
         zone_name.
     """
 
-    zone_dic = {}
-    zone_interconnect_dic = {}
-    for i in range(len(zone)):
-        tu = (zone["zone_name"][i], zone["interconnect"][i])
-        zone_dic[zone["zone_name"][i]] = zone["zone_id"][i]
-        zone_interconnect_dic[tu] = zone["zone_id"][i]
+    zone_dic = pd.Series(zone.zone_id, index=zone.zone_name).to_dict()
+    zone_interconnect_dic = pd.Series(
+        zone.zone_id, index=zip(zone.zone_name, zone.interconnect)
+    ).to_dict()
+
     return zone_dic, zone_interconnect_dic
 
 
--- a/prereise/gather/griddata/hifld/calculate/remap.py
+++ b/prereise/gather/griddata/hifld/calculate/remap.py
@@ -30,3 +30,31 @@ def get_line_mapping(lines):
     raw_data["ID"] = raw_data["ID"].astype("str")
     raw_lines = raw_data.set_index("ID").to_dict()
     return raw_lines
+
+
+def get_sub_mapping(clean_data):
+    """Generate the subs
+
+    :param str E_csv: path of the HIFLD substation csv file
+    :return: (*dict*) --  sub_by_coord_dict, a dict mapping from (x, y) to substation detail.
+    :return: (*dict*) --  sub_name_dict, a dict mapping from substation name to its coordinate (x, y).
+    """
+
+    sub_by_coord_dict = {}
+    sub_name_dict = {"sub": []}
+    for index, row in clean_data.iterrows():
+        location = (row["LATITUDE"], row["LONGITUDE"])
+        if location in sub_by_coord_dict:
+            raise Exception(
+                f"WARNING: substations coordinates conflict check: {location}"
+            )
+        sub_by_coord_dict[location] = (
+            row["ID"],
+            row["NAME"],
+            row["STATE"],
+            row["COUNTY"],
+        )
+        if row["NAME"] not in sub_name_dict:
+            sub_name_dict[row["NAME"]] = []
+        sub_name_dict[row["NAME"]].append((row["LATITUDE"], row["LONGITUDE"]))
+    return sub_by_coord_dict, sub_name_dict
--- a/prereise/gather/griddata/hifld/calculate/remap.py
+++ b/prereise/gather/griddata/hifld/calculate/remap.py
@@ -1,3 +1,6 @@
+import copy
+
+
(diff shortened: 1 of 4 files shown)


prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -0,0 +1,10 @@
+import pandas as pd
+
+
+def load_csv(csv_filename, dtypes=None):
+    if dtypes:
+        data = pd.read_csv(csv_filename, dtype=dtypes)
+    else:
+        data = pd.read_csv(csv_filename)
```
danielolsen, line 8: How is this behavior any different than the existing `pandas.read_csv`?

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+    dictionary is a dict mapping the STATE to its ID and the second is a dict
+    mapping a tuple of state and interconnect to its ID.
+    """
+
+    zone_dic = {}
+    zone_interconnect_dic = {}
+    for i in range(len(zone)):
+        tu = (zone["zone_name"][i], zone["interconnect"][i])
+        zone_dic[zone["zone_name"][i]] = zone["zone_id"][i]
+        zone_interconnect_dic[tu] = zone["zone_id"][i]
```
danielolsen, line 18: Is this equivalent to:
```python
zone_dic = pd.Series(zone.zone_id, index=zone.zone_name"].to_dict()
zone_interconnect_dic = pd.Series(zone.zone_id, index=zip(zone.zone_name, zone.interconnect)).to_dict()
```
?
We could also skip the `to_dict` call and just return Series objects.

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+    return zone_dic, zone_interconnect_dic
```
rouille, line 19: I would do:
```
if zone.index.name != "zone_id":
    zone.set_index("zone_id", inplace=True)
return {v: k for k, v in toto["zone_name"].to_dict().items()}, zone.groupby(["zone_name", "interconnect"]).groups
```

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
+import copy
+
+
+def get_zone_mapping(zone):
+    """Generate dictionaries of zone using the zone.csv
+
+    :param pandas.DataFrame zone: the zone.csv data
+    :return: (*tuple*) -- a tuple of two dictionaries, where the first
+    dictionary is a dict mapping the STATE to its ID and the second is a dict
+    mapping a tuple of state and interconnect to its ID.
```
rouille, line 10: the above 2 lines should be indented

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+        tu = (zone["zone_name"][i], zone["interconnect"][i])
+        zone_dic[zone["zone_name"][i]] = zone["zone_id"][i]
+        zone_interconnect_dic[tu] = zone["zone_id"][i]
+    return zone_dic, zone_interconnect_dic
+
+
+def get_line_mapping(lines):
+    """Create dict to store all the raw transmission line csv data
+
+    :param str t_csv: path of the HIFLD transmission csv file
```
rouille, line 25: `t_csv` --> `lines`

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+    return raw_lines
+
+
+def get_sub_mapping(clean_data):
+    """Generate the subs
+
+    :param str E_csv: path of the HIFLD substation csv file
+    :return: (*dict*) --  sub_by_coord_dict, a dict mapping from (x, y) to substation detail.
+    :return: (*dict*) --  sub_name_dict, a dict mapping from substation name to its coordinate (x, y).
+    """
```
danielolsen, line 41: Docstring does not match inputs.

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+    raw_data = copy.deepcopy(lines)
+    raw_data["ID"] = raw_data["ID"].astype("str")
+    raw_lines = raw_data.set_index("ID").to_dict()
+    return raw_lines
+
+
+def get_sub_mapping(clean_data):
+    """Generate the subs
+
+    :param str E_csv: path of the HIFLD substation csv file
```
rouille, line 38: `E_csv` --> `clean_data`

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+    :return: (*dict*) --  sub_by_coord_dict, a dict mapping from (x, y) to substation detail.
+    :return: (*dict*) --  sub_name_dict, a dict mapping from substation name to its coordinate (x, y).
+    """
+
+    sub_by_coord_dict = {}
+    sub_name_dict = {"sub": []}
```
danielolsen, line 44: What is the purpose of this entry?

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+    raw_lines = raw_data.set_index("ID").to_dict()
+    return raw_lines
+
+
+def get_sub_mapping(clean_data):
+    """Generate the subs
+
+    :param str E_csv: path of the HIFLD substation csv file
+    :return: (*dict*) --  sub_by_coord_dict, a dict mapping from (x, y) to substation detail.
+    :return: (*dict*) --  sub_name_dict, a dict mapping from substation name to its coordinate (x, y).
```
rouille, line 40: There should be only one return parameter. It is a tuple and each element is a dictionary

prereise/gather/griddata/hifld/calculate/remap.py
rouille, line 44: Can we remove `dict` from the variable name? I am not sure what it brings here except that the name is longer

prereise/gather/griddata/hifld/calculate/remap.py
```diff
@@ -0,0 +1,60 @@
...
+            )
+        sub_by_coord_dict[location] = (
+            row["ID"],
+            row["NAME"],
+            row["STATE"],
+            row["COUNTY"],
+        )
+        if row["NAME"] not in sub_name_dict:
+            sub_name_dict[row["NAME"]] = []
+        sub_name_dict[row["NAME"]].append((row["LATITUDE"], row["LONGITUDE"]))
```
danielolsen, line 59: Is this equivalent to:
```python
if len(clean_data.groupby(["LATITUDE", "LONGITUDE"])) == len(clean_data):
    raise Exception(...)
sub_by_coord_dict = clean_data.set_index(["LATITUDE", "LONGITUDE"])["ID", "NAME", "STATE", "COUNTY"]
sub_name_dict = clean_data.set_index("NAME")["LATITUDE", "LONGITUDE"]
```
?

prereise/gather/griddata/hifld/model.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def get_grid_data(e_csv, t_csv, z_csv):
+    """Load HIFLD data and create dataframes corresponding to the input csv's
+    for the BES grid.
+
+    :param str e_csv: path of the HIFLD substation csv file
+    :param str t_csv: path of the HIFLD transmission csv file
+    :param str z_csv: path of the zone csv file
+    """
```
danielolsen, line 17: Missing returns.

prereise/gather/griddata/hifld/model.py
```diff
@@ -0,0 +1,29 @@
...
+
+    zone_data = load_csv(z_csv)
+    zone_dic, zone_interconnect_dic = get_zone_mapping(zone_data)
+
+    sub_data = load_csv(e_csv, dtypes={"COUNTYFIPS": str})
+    clean_data = clean_substations(sub_data, zone_dic)
+    sub_by_coord_dict, sub_name_dict = get_sub_mapping(clean_data)
+    
+    line_data = load_csv(t_csv)
+    raw_lines = get_line_mapping(line_data)
```
danielolsen, line 27: `raw_lines` is defined but never used?

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -0,0 +1,10 @@
...
+
+    return data
```
rouille, line 10: Is the `dtypes` used somewhere? Just asking because it is not used in the `get_grid_data` function of the `model` module

rouille: Oops, it is. Sorry about that.

prereise/gather/griddata/hifld/model.py
```diff
@@ -0,0 +1,29 @@
...
+    :param str t_csv: path of the HIFLD transmission csv file
+    :param str z_csv: path of the zone csv file
+    """
+
+    zone_data = load_csv(z_csv)
+    zone_dic, zone_interconnect_dic = get_zone_mapping(zone_data)
+
+    sub_data = load_csv(e_csv, dtypes={"COUNTYFIPS": str})
+    clean_data = clean_substations(sub_data, zone_dic)
+    sub_by_coord_dict, sub_name_dict = get_sub_mapping(clean_data)
```
rouille, line 24: same thing. Can we remove the `dict` from the name?

prereise/gather/griddata/hifld/model.py
```diff
@@ -0,0 +1,29 @@
...
+from prereise.gather.griddata.hifld.calculate.clean import clean_substations
+
+
+def get_grid_data(e_csv, t_csv, z_csv):
+    """Load HIFLD data and create dataframes corresponding to the input csv's
+    for the BES grid.
+
+    :param str e_csv: path of the HIFLD substation csv file
+    :param str t_csv: path of the HIFLD transmission csv file
+    :param str z_csv: path of the zone csv file
```
rouille, line 16: the `:return:` is missing here

prereise/gather/griddata/hifld/calculate/remap.py
rouille: @BainanXia and @danielolsen, do we need to return the second element now that the `zone_name` are unique?

prereise/gather/griddata/hifld/model.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    zone_data = load_csv(z_csv)
+    zone_dic, zone_interconnect_dic = get_zone_mapping(zone_data)
```
rouille, line 20: can we use `zone2id` as a variable name here? This is what we use in **PowerSimData**

danielolsen, line 20: `zone_interconnect_dic` is defined but never used?

prereise/gather/griddata/hifld/calculate/remap.py
danielolsen: We don't seem to use the second element at all, at least so far. These functions are generating lots of mappings, but it's unclear what the mappings will be used for. Once we refactor to use built-in pandas functions, if these functions are only used in one place, I'm not sure it's even worth defining them rather than just including the logic in the higher-level function.

prereise/gather/griddata/hifld/model.py
rouille: `zone_interconnect_dic` is not used I believe so we do `zone_id, _ = get_zone_mapping(zone_data)` or just `zone_id = get_zone_mapping(zone_data)` if we don't return the second element (see comment under `get_zone_mapping` function)

danielolsen: Or just `zone_id = pd.Series(zone_data.zone_id, index=zone_data.zone_name").to_dict()`, and avoid defining the function at all.

danielolsen: Or really, since the current `zone_dic` is only used to get an interable for use within `clean_substations`, we just need `zone_data.zone_id`.

prereise/gather/griddata/hifld/model.py
danielolsen: We don't seem to even use these dictionaries.

prereise/gather/griddata/hifld/calculate/remap.py
BainanXia: This function is getting zone mapping, but according to the doc strings here, none of the returns is zone related, which is ambiguous. I presume ID should be zone_id, STATE should be zone_name?

prereise/gather/griddata/hifld/calculate/remap.py
BainanXia: @rouille I agree with @danielolsen, these returns contain duplicated information given we have one-to-one mapping between `zone_id` and `zone_name`. Also, we should be clear what mapping is useful during grid model creation to avoid duplicate code with `powersimdata.network.hifld.constants.zones.py` in the future.

BainanXia:
Reading through the PR and I've got some higher level comments:
- Not sure what is the purpose of the functions in `remap` module, which essentially turn data frames into dictionaries and the rest of the code doesn't show a reason of doing so.
- The name of `calculate` folder is vague since anything can be calculations. We should rename it into something more specific, however, I don't see the core functionalities of this module yet. The data cleaning module can be moved to `data_access` folder, given the main responsibility of `data_access` in this module is just read csvs, which is different from what we have in `powersimdata`.
- Given `model.py` is used to assemble the info from raw csvs into a raw grid model, most of the existing simple operations in `calculate` can be moved to `get_grid_data.py` directly.

prereise/gather/griddata/hifld/calculate/remap.py
ahurli: Done. Updated the docstrings to reflect the headers found in the input csv.

prereise/gather/griddata/hifld/calculate/remap.py
ahurli: Thanks for the pandas recommendation!

As far as the `to_dict` called, this data gets accessed in dictionary form later on in the (`data_trans.py`) code. The conversion can either happen here or wherever it gets accessed, but the `zone_dic` dictionary is already used in the pandas query in `clean.py` and both will be used to write the data to the csv's expected by our existing `Grid` object. I would say it's easier to refactor everything after the fact to just Series objects than do that now.

rouille: Why do we need this function at all?

rouille: By the way the `zone_dic` won't work to query the substations in your `clean` module since `STATE` in the substation file are state abbreviation and the key of `zone_dic` are load zone name not abbreviated.

ahurli:
So regarding the mapping functions, the idea was to prepare the data for future functions that expect input in this format. It's a little difficult to decide where to stop this PR as ultimately, I think `get_grid_data` should return the csv's that are expected by our `Grid` object, analogous to the existing `data_trans.py` file in the original HIFLD code. This is difficult because the existing function is a main function that, as far as I can tell, covers 3 distinct parts: 1) load and cleaning/reformatting data into the expected formats 2) filtering out islands from the initial data set 3) writing the results from the loading and filtering into the various csv's (i.e. bus.csv, branch.csv, sub.csv, etc.). This PR covers that first part, so judgements about the outputs are a little difficult without taking the second and third parts into account. Including all of these in one PR, however, would be too large.

The content of `get_grid_data` is confusing right now as it really more of an example giving context of how to use the functions in the subfolders, but needs more work to complete. Without it, though, the individual functions make even less sense.

I'll be working through the other comments by EOD.

prereise/gather/griddata/hifld/calculate/remap.py
ahurli: Done. I did the same for the zone mappings as well to be consistent.

prereise/gather/griddata/hifld/data_access/load.py
ahurli: It's not necessary, but it's a placeholder for the larger section of code as we move the data to other places. I can include the code now to read from blob storage, but I thought that was perhaps outside the scope of this PR.

danielolsen:
I think we need to get consensus about what the scope of this PR is. My initial impression was that it was about functions for loading/cleaning data. If instead the scope is designing a user-facing `get_grid_data` function that will end up writing CSVs compatible with PowerSimData, then I think what we should be discussing is very different than the code that's in this PR.

prereise/gather/griddata/hifld/data_access/load.py
rouille: Reading from blob storage is the same, e.g,
`pd.read_csv("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Substations_Jul2020.csv")`. Just the docstring would need to be modified to indicate that it can be a url instead of a file path

PR #197, opened 2021-07-23
feat: process plant data to create data frame

danielolsen:

### Purpose
Parse HIFLD data into the beginning of a `plant` data frame, for use with a Grid object.

### What the code is doing
`build_plant` is the entry-point function, which will eventually return a data frame with all information, and all lower-level functions are accessed from this function. `build_plant` effectively replaces the old `plant_agg` function, with different structure. The lower level functions are:
- `map_generator_to_sub_by_location`: look up substations within that interconnect and ZIP code, or if there are none, look within that interconnect and 'nearby' ZIP codes (+- 100). The logic of 'nearby' could probably be improved, but this is what was implemented in the collaborators' code.
- `map_generator_to_bus_by_sub`: pick the lowest voltage bus within a specified substation.
- `add_locations_to_generators`: use the locations from **Power_Plants.csv**, map them to the generating units as best as possible.
- `map_generators_to_interconnects`: look up the county for each substation using the Power Plant mapping, then determine the interconnect.
- `add_pmin_to_generators`: look up Pmin values from EIA Form 860.
- `helpers.map_state_and_county_to_interconnect`: look up information from **const.py** to map a state/county to an interconnection. We also call this from `data_process.transmission` so that we can access this information for substations from within `map_generator_to_sub_by_location`.

There are also placeholders for:
- `estimate_heat_rate_from_ampd`
- `add_heat_rate_to_generators`

These will need further research into the best way to implement, but they are independent and so can wait for a follow-up PR.

### Usage Example
```python
...     build_transmission,
...     create_buses,
... )
>>> from prereise.gather.griddata.hifld.data_process.generators import build_plant
>>> bus = create_buses(lines)
>>> plant = build_plant(bus, substations)
>>> plant.isna().sum()
ID                0
NAME              0
CITY              0
STATE             0
ZIP               0
TYPE              0
STATUS            0
COUNTRY           0
NAICS_CODE        0
NAICS_DESC        0
SOURCE            0
SOURCEDATE        0
PLANT             0
PLANT_CODE       86
OPERATOR          0
OPERAT_ID         0
NAMPLT_CAP        0
SUMMER_CAP        0
WINTER_CAP        0
PRIM_FUEL         0
FUEL1_TRAN        0
SEC_FUEL          0
FUEL2_TRAN        0
PRIM_MVR          0
RUN_DATE          0
RET_DATE          0
lat             402
lon             402
interconnect    402
sub_id          687
bus_id          687
Pmin              0
dtype: int64
```

Not all plant codes can be mapped successfully, and same with lats/lons, and therefore sub_id & bus_id, but we're only missing about 3%. Suggestions for improvement welcome.

### Time to review
1 hour.


prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,71 @@
+import pandas as pd
+
+from prereise.gather.griddata.hifld.data_access import load
+
+
+def map_generator_to_bus_by_location(generator, substation_groupby):
```
rouille, line 6: Should it be: `map_generator_to_sub_by_location`?

danielolsen: Eventually we need to have the generators mapped to bus IDs. This function will probably need another input (e.g. a bus table) to do that properly.

danielolsen: Fulfilling #205 should give us bus IDs, and then we probably want to connect each generator to the lowest-voltage bus within each substation.

danielolsen:
This branch is now ready for a full review. The original post has been updated with more details.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def map_generator_to_sub_by_location(generator, substation_groupby):
+    """Determine a likely substation for a generator to be connected to. Priority order
+    of mapping is: 1) if location is available and one or more substations exist in that
+    ZIP code, map by location to closest substation within that ZIP code, 2) if location
+    is available but no substations exist in that ZIP code, map to the closest
+    substation within neighboring ZIP codes, 3) if only ZIP code is available
+    (no location), and one or more substations exist, map to an arbitrarily chosen
+    substation within that ZIP code, 4) if if only ZIP code is available (no location)
```
rouille, line 18: `if if` --> `if`

prereise/gather/griddata/hifld/data_process/helpers.py
```diff
@@ -0,0 +1,23 @@
+from prereise.gather.griddata.hifld import const
+
+
+def map_state_and_county_to_interconnect(state_abv, county):
+    """Map a state and a county to an assumed interconnection.
+
+    :param str state_raw: two-letter state abbreviation.
```
rouille, line 7: `state_raw` --> `state_abv`

prereise/gather/griddata/hifld/const.py
```diff
@@ -118,3 +120,129 @@
...
+        },
+        "west": {"EL PASO", "HUDSPETH"},
+    },
+}
+
+hifld_storage_types = {
+    "BATTERIES",
+    "NATURAL GAS WITH COMPRESSED AIR STORAGE",
+    "FLYWHEELS",
+}
```
rouille, line 248: I am wondering if at some point we don't want to move these constants in:
* `powersimdata/network/hifld/constants/plants.py`
* powersimdata/network/hifld/constants/zones.py
once they exist

prereise/gather/griddata/hifld/data_process/helpers.py
```diff
@@ -0,0 +1,23 @@
...
+    state_upper = state_abv.upper()
+    for region in ("east", "west"):
+        if state_upper in const.interconnect2state[region]:
+            return region
+    if state_upper in const.interconnect2state["split"]:
+        for region in set(const.state_county_splits[state_upper].keys()) - {"default"}:
+            if county.upper() in const.state_county_splits[state_upper][region]:
+                return region
+        return const.state_county_splits[state_upper]["default"]
+    else:
```
jon-hagg, line 22: I think we could drop the `else`

prereise/gather/griddata/hifld/data_process/helpers.py
```diff
@@ -0,0 +1,23 @@
...
+
+def map_state_and_county_to_interconnect(state_abv, county):
+    """Map a state and a county to an assumed interconnection.
+
+    :param str state_raw: two-letter state abbreviation.
+    :param str county: county name.
+    :raises ValueError: if the provided state abbreviation isn't present in mappings.
+    :return: (*str*) -- interconnection name.
+    """
+    # Coerce state abbreviation to upper-case
```
jon-hagg, line 12: This line seems self explanatory

prereise/gather/griddata/hifld/const.py
danielolsen: It could make sense, `const.py` is getting a bit long.

prereise/gather/griddata/hifld/data_process/helpers.py
danielolsen: Fair enough, this was more of a debug/development `else` so that I didn't get mysterious NoneType returns downstream. Do you think it's worth refactoring as an input check, or ditch completely?

jon-hagg: I was thinking just keep the error at the end, but remove the `else`, since it's implied. Makes sense to raise instead of letting it return `None`.

danielolsen: Aha

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,197 @@
...
+    )
+    return distance_to_subs.idxmin()
+
+
+def map_generator_to_bus_by_sub(generator, bus_groupby):
+    """Determine a likely bus for a generator to be connected to, based on the bus with
+    the lowest voltage within the generator's specified substation.
+
+    :param pandas.Series generator: one generating unit from data frame.
+    :param pandas.GroupBy bus_groupby: data frame of buses, grouped by substation ID.
```
jon-hagg, line 64: Needs `:return:`

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,197 @@
...
+    ZIP code, map by location to closest substation within that ZIP code, 2) if location
+    is available but no substations exist in that ZIP code, map to the closest
+    substation within neighboring ZIP codes, 3) if only ZIP code is available
+    (no location), and one or more substations exist, map to an arbitrarily chosen
+    substation within that ZIP code, 4) if if only ZIP code is available (no location)
+    but no substations exist in that ZIP code, return NA.
+
+    :param pandas.Series generator: one generating unit from data frame.
+    :param pandas.GroupBy substation_groupby: data frame of substations, grouped by
+        (interconnect, ZIP).
```
jon-hagg, line 23: Needs `:return:`

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def map_generators_to_interconnects(generators, power_plants):
+    """Add 'interconnect' column to ``generators``, based on matching plant codes in
+    ``power_plants``. Non-matched generating units will get NaN values.
+
+    :param pandas.DataFrame generators: data frame of generating units
+        (will be modified with new columns).
+    :param pandas.DataFrame power_plants: data frame of power plants
+        (locations come from here).
```
jon-hagg, line 98: Needs `:return:`

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,197 @@
...
+        power_plants[["LATITUDE", "LONGITUDE"]],
+        index=power_plants.index.astype("str"),
+    )
+    filtered_generators = generators.query("not PLANT_CODE.isnull()")
+    parsed_plant_codes = filtered_generators.PLANT_CODE.astype("int").astype("str")
+    generators["lat"] = parsed_plant_codes.map(plant_locations["LATITUDE"])
+    generators["lon"] = parsed_plant_codes.map(plant_locations["LONGITUDE"])
+
+
+def map_generators_to_interconnects(generators, power_plants):
```
jon-hagg, line 91: I think the description for this could be slightly modified - it sounds like the intent is to modify `generators` but from the way it's used, it is creating the interconnect column which is assigned within `build_plant`.

danielolsen: You're right, that was how the function was originally written before I refactored it. Will fix.

danielolsen:
Following some discussion in #222, I'm going to refactor this to use all EIA source data for generators rather than the current HIFLD/EIA mix.

danielolsen:
The `build_plant` function has been rewritten from scratch starting with EIA plant data rather than HIFLD plant data (the old version can be found at `daniel/hifld_plant_hifld_base` for now):
```python
>>> from prereise.gather.griddata.hifld.data_process.generators import build_plant
...     build_transmission,
...     create_buses,
... )
>>>
2119 line voltages can't be found via neighbor consensus
310 line voltages can't be found via neighbor minimum
>>> bus = create_buses(lines)
>>> generators = build_plant(bus, substations)
sys:1: DtypeWarning: Columns (15,37,38,42) have mixed types.Specify dtype option on import or set low_memory=False.
>>> columns_of_interest = [
...     "Plant Code",
...     "Generator ID",
...     "Technology",
...     "lat",
...     "lon",
...     "interconnect",
...     "sub_id",
...     "bus_id",
...     "Pmax",
...     "Pmin",
... ]
>>> generators[columns_of_interest].isna().sum()
Plant Code         0
Generator ID       0
Technology         0
lat                0
lon                0
interconnect       0
sub_id           383
bus_id           383
Pmax              72
Pmin            3975
dtype: int64
```
Fewer generators can't be mapped to buses as compared to before, and the missing Pmin and Pmax value reflect missing entries in the original source data. Un-mapped buses represent only 1.8% of the generator count, and only 1.0% of the generator capacity:
```python
>>> generators.loc[generators["bus_id"].isna(), "Pmax"].sum() / generators["Pmax"].sum()
0.009851291657831352
>>> generators["bus_id"].isna().sum() / len(generators)
0.01755672702269081
```

I think this is ready for review again.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+from powersimdata.utility.distance import haversine
+
+from prereise.gather.griddata.hifld import const
+from prereise.gather.griddata.hifld.data_access import load
+
+
+def floatify(value, default=pd.NA):
+    """Return a float if possible, otherwise return a default value.
+
+    :param object value: value to be coerced to float, if possible.
```
rouille, line 11: object --> str (from what I can see in the `isinstance`)

danielolsen: It could be any object type, even a float already. We just have extra logic to assist with strings.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+    :param object default: value to be returned if float conversion isn't possible.
```
rouille, line 12: Should we use `np.NaN` instead. Its type is a `float` while `pd.NA` is not:
```
>>> type(pd.NA)
<class 'pandas._libs.missing.NAType'>
>>> type(np.NaN)
<class 'float'>
```

danielolsen: Good idea. We could go even simpler with just `float("nan")`:
```python
>>> type(float("nan"))
<class 'float'>
```

danielolsen: Done.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+from prereise.gather.griddata.hifld import const
+from prereise.gather.griddata.hifld.data_access import load
+
+
+def floatify(value, default=float("nan")):
+    """Return a float if possible, otherwise return a default value.
+
+    :param object value: value to be coerced to float, if possible.
+    :param object default: value to be returned if float conversion isn't possible.
+    :return: (*float/object*) -- float or default value as applicable.
```
rouille, line 13: I think the returned type is `float` now, no?

danielolsen: Not if the user supplies an alternate default value that's not a float.

rouille: Yeah, indeed

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+    is available but no substations exist in that ZIP code, map to the closest
+    substation within neighboring ZIP codes, 3) if only ZIP code is available
+    (no location), and one or more substations exist, map to an arbitrarily chosen
+    substation within that ZIP code, 4) if only ZIP code is available (no location)
+    but no substations exist in that ZIP code, return NA.
+
+    :param pandas.Series generator: one generating unit from data frame.
+    :param pandas.GroupBy substation_groupby: data frame of substations, grouped by
+        (interconnect, ZIP).
+    :return: (*int/pd.NA*) -- substation ID if the generator can be mapped successfully
```
rouille, line 36: same comment regarding NaN.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+        to a substation, else pd.NA.
+    """
+    lookup_params = tuple(generator.loc[["interconnect", "ZIP"]])
+    if pd.isna(generator["lat"]) or pd.isna(generator["lon"]):
+        # No location available
+        try:
+            matching_subs = substation_groupby.get_group(lookup_params)
+            return matching_subs.index[0]
+        except KeyError:
+            return pd.NA
```
rouille, line 46: `return float("nan")`

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+            matching_subs = pd.concat(
+                [
+                    substation_groupby.get_group((generator.loc["interconnect"], z))
+                    for z in zip_range_strings
+                    if (generator.loc["interconnect"], z) in substation_groupby.groups
+                ]
+            )
+        except ValueError:
+            # If no matching subs within the given interconnection and ZIPs, give up
+            return pd.NA
```
rouille, line 65: `return float("nan")`

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+def map_generator_to_bus_by_sub(generator, bus_groupby):
+    """Determine a likely bus for a generator to be connected to, based on the bus with
+    the lowest voltage within the generator's specified substation.
+
+    :param pandas.Series generator: one generating unit from data frame.
+    :param pandas.GroupBy bus_groupby: data frame of buses, grouped by substation ID.
+    :return: (*int/pd.NA*) -- bus ID if the generator has a substation ID, else pd.NA.
+    """
+    if pd.isna(generator.sub_id):
+        return pd.NA
```
rouille, line 82: `return float("nan")`

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Not sure what the best approach is when we would return an `int`. If we return `pd.NA`, we get a Series of dtype `object`, but the int values stay int (which is important when we want to use them to index into something else). If we return `float("nan")`, the Series gets converted to all floats, including the ones that were originally int:
```python
>>> import pandas as pd
>>> pd.Series([2, float("nan")])
0    2.0
1    NaN
dtype: float64
>>> pd.Series([2, pd.NA])
0       2
1    <NA>
dtype: object
```

rouille: You are thinking in the user facing function when we map through a `lambda` function?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+                ]
+            )
+        except ValueError:
+            # If no matching subs within the given interconnection and ZIPs, give up
+            return pd.NA
+    distance_to_subs = matching_subs.apply(
+        lambda x: haversine((x.LATITUDE, x.LONGITUDE), (generator.lat, generator.lon)),
+        axis=1,
+    )
+    return distance_to_subs.idxmin()
```
rouille, line 70: Do we use the interconnect of the substation at all in the above function? It seems that we could only use the ZIP of the substations

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Exactly:
```python
>>> import pandas as pd
>>> a = pd.Series([1,2,3])
>>> a.map(lambda x: x if x % 2 == 1 else pd.NA)
0       1
1    <NA>
2       3
dtype: object
>>> a.map(lambda x: x if x % 2 == 1 else float("nan"))
0    1.0
1    NaN
2    3.0
dtype: float64
```

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: I'm following the original design from the collaborators, and one of the considerations may have been to avoid mapping to a 'neighboring' ZIP code in a different interconnect (see the `zip_range` section).

rouille: I see, it can happen that when we augment the ZIP coverage, some could be in a different interconnect. Are we sure that each generator has an interconnect assigned before grouping?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -0,0 +1,136 @@
...
+    generators["ZIP"] = generators["Plant Code"].map(plants["Zip"])
+    generators["sub_id"] = generators.apply(
+        lambda x: map_generator_to_sub_by_location(x, substation_groupby), axis=1
+    )
+    generators["bus_id"] = generators.apply(
+        lambda x: map_generator_to_bus_by_sub(x, bus_groupby), axis=1
+    )
+    generators["Pmax"] = generators[
+        ["Summer Capacity (MW)", "Winter Capacity (MW)"]
+    ].max(axis=1)
```
rouille, line 133: Is that preferable than to take the average?

danielolsen: Good question. I erred on the side of creating a more-feasible system. Ideally we would have the ability to have seasonal capacities within PowerSimData/REISE.jl.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Using the `[ZIP-100, ZIP+100]` as the set of neighboring ZIPs to look for substations in could probably be improved as well, since the numerical distance between two ZIPs is only a rough proxy for the geographical distance between two ZIPs. E.g. looking at [this map](https://www.mapshop.com/united-states-3-digit-zip-code-wall-map/) we see that ZIP codes 822xx in Wyoming are bordered by ZIP codes 820xx, 826xx, and 827xx, not 821xx or 823xx.

rouille: http://www.geonames.org/export/web-services.html when we need to improve the method

prereise/gather/griddata/hifld/data_process/generators.py
rouille: Sounds good. I guess there is no seasonal difference for nuclear, natural gas and coal but could be significant for hydro, solar and wind

danielolsen: Hydro, solar, and wind will be controlled by the profiles, but thermal generators can have also different net capacities in summer and winter because of ambient temperature differences, cooling water temperatures, etc.

rouille: Ah. Thanks for the explanation

rouille (approved):
Nice

prereise/gather/griddata/hifld/data_process/helpers.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/helpers.py
danielolsen: Removed.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/generators.py

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: This function was made obsolete by starting from EIA data, which have NERC regions, which map cleanly to interconnects.

PR #198, opened 2021-08-02
data: manually fix BA county map to remove duplicate assignments

BainanXia:

### Purpose
I noticed I've got different bus BA assignments when reviewing @ywang303 's [PR](https://github.com/Breakthrough-Energy/PowerSystemComparison/pull/2). To track down the issue, I went through the notebook that is used to generate BA_County_map.json and found we didn't enforce the BA county mapping to be 1-to-1, i.e. one county could be mapped to multiple BAs due to the overlap service areas (however, this is rare, only 10 counties have such issue). This will cause a problem when we generate bus BA mapping via function [map_buses_to_county](https://github.com/Breakthrough-Energy/PreREISE/blob/f3777f61379e850585030634afce60f97df2fbae/prereise/gather/demanddata/eia/map_ba.py#L73). Buses belong those 10 counties will be assigned to different BAs depending on the order we parse the dictionary when getting county BA mapping, however, the key ordering within a dictionary is unstable. 

### What the code is doing
Manually assign those 10 counties to one BA. Update `BA_County_map.json` and `all_bus_ba_map.csv` correspondingly.

### Testing
N/A

### Where to look
```
prereise/gather/data/BA_County_map.json
prereise/gather/data/all_bus_ba_map.csv
```

### Usage Example/Visuals
N/A

### Time estimate
5 min longer if dive into the nuances.


danielolsen (approved):
The logic makes sense.

PR #196, opened 2021-07-15
Load HIFLD data

rouille:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Load HIFLD data. The `ben/read` branch has been branched off `ben/const` that is is under review (#194)

### What the code is doing
For each of the 5 HIFLD, the file is loaded and filtered. The user defined table enclosing the information on the load zone is also loaded.

### Testing
Manual testing, e.g.:
```
>>> from prereise.gather.griddata.hifld.data_access.load import get_hifld_electric_power_transmission_lines_topology
>>> topology = get_hifld_electric_power_transmission_lines_topology("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Power_Transmission_Lines_Jul2020.geojson.zip")
>>> type(topology)
<class 'list'>
>>> len(topology)
71554
>>> from prereise.gather.griddata.hifld.data_access.load import get_hifld_electric_power_transmission_lines
>>> lines = get_hifld_electric_power_transmission_lines("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Power_Transmission_Lines_Jul2020.csv")
>>> len(lines)
71554
```

### Where to look
Everything is in the ` prereise.gather.griddata.hifld.data_access.load` module

### Usage Example/Visuals
N/A

### Time estimate
20min

PR #199, opened 2021-08-12
feat: add functions to perform transmission line filtering

danielolsen:

### Purpose
Implement initial filtering of transmission lines and substations, following the lead from @LFPower, @terrywqf, and @Mingze-Li128.

### What the code is doing
- `build_transmission` is the main user-facing function, which kicks off several lower-level functions in turn. This will eventually produce the final transmission network topology (substations, buses, branches), although this PR does not go that far.
- `get_substations_with_nonzero_lines` filters based on the `LINES` column of the substation table (as done in the `clean` function of @Mingze-Li128's branch).
- `check_for_location_conflicts` checks for any substations with duplicate lat/lon pairs.
- `filter_lines_with_unavailable_substations` looks for lines for which one or more substation is listed as `"NOT AVAILABLE"` in the line table (as done in the `neighbors` function of @Mingze-Li128's branch).
- `filter_lines_with_no_matching_substations` looks for lines for which there is no matching substation in the substation table (as done in the `neighbors` function of @Mingze-Li128's branch).
- `filter_lines_with_nonmatching_substation_coords` checks that the line start coordinates are within 100 miles of the coordinates of the substation listed in the `SUB_1` column, and that the line end coordinates are within 100 miles of the coordinates of the substations listed in the `SUB_2` column (as done in the `neighbors` function of @Mingze-Li128's branch).
- `filter_lines_with_identical_substation_names` does what the name suggests (as done in the `neighbors` function of @Mingze-Li128's branch).

### Testing
Manually tested, by replacing `BLOB_PATH` and `REPO_PATH` placeholders with locally-appropriate locations.

### Usage Example/Visuals
```python
>>> build_transmission()

[...big ugly dataframe output truncated manually by Daniel]

[67389 rows x 16 columns]
```

### Time estimate
30 minutes.

### DIFF
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -1,0 +1,41 @@
+import os
+
+from powersimdata.utility.distance import haversine
+
+from prereise.gather.griddata.hifld import const
+from prereise.gather.griddata.hifld.data_access.load import (
+    get_hifld_electric_power_transmission_lines,
+    get_hifld_electric_substations,
+    get_zone,
+)
+
+
+def filter_substations_with_zero_lines(substations):
+    """Filter substations with LINES attribute equal to zero, and report the number
+    dropped.
+
+    :param pandas.DataFrame substations: data frame of all substations.
+    :return: (*pandas.DataFrame*) -- substations with non-zero values for LINES.
+    """
+    num_substations = len(substations)
+    num_substations_without_lines = len(substations.query("LINES == 0"))
+    print(
+        f"dropping {num_substations_without_lines} substations "
+        f"of {num_substations} total due to LINES parameter equal to 0"
+    )
+
+    return substations.query("LINES != 0").copy()
+
+
+def build_transmission():
+    """Main user-facing entry point."""
+    # Load input data
+    hifld_substations = get_hifld_electric_substations(const.blob_paths["substations"])
+    hifld_lines = get_hifld_electric_power_transmission_lines(
+        const.blob_paths["transmission_lines"]
+    )
+    hifld_data_dir = os.path.join(os.path.dirname(__file__), "..", "data")
+    hifld_zones = get_zone(os.path.join(hifld_data_dir, "zone.csv"))  # noqa: F841
+
+    # Filter substations
+    substations_with_lines = filter_substations_with_zero_lines(hifld_substations)
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -58,6 +58,27 @@ def filter_lines_with_unavailable_substations(lines):
     return lines.query("SUB_1 != 'NOT AVAILABLE' and SUB_2 != 'NOT AVAILABLE'").copy()
 
 
+def filter_lines_with_no_matching_substations(lines, substations):
+    """Filter lines with one or more substation name not present in the ``substations``
+    data frame, and report the number dropped.
+
+    :param pandas.DataFrame lines: data frame of lines.
+    :param pandas.DataFrame substations: data frame of substations.
+    :return: (*pandas.DataFrame*) -- lines with matching substations.
+    """
+    num_lines = len(lines)
+    matching_names = substations["NAME"]  # noqa: F841
+    filtered = lines.query(
+        "SUB_1 not in @matching_names or SUB_2 not in @matching_names"
+    )
+    num_filtered = len(filtered)
+    print(
+        f"dropping {num_filtered} lines with one or more substations not found in "
+        f"substations table out of a starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 in @matching_names and SUB_2 in @matching_names").copy()
+
+
 def build_transmission():
     """Main user-facing entry point."""
     # Load input data
@@ -74,3 +95,6 @@ def build_transmission():
 
     # Filter lines
     lines_with_substations = filter_lines_with_unavailable_substations(hifld_lines)
+    lines_with_matching_substations = filter_lines_with_no_matching_substations(
+        lines_with_substations, substations_with_lines
+    )
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -41,6 +41,23 @@ def filter_substations_with_zero_lines(substations):
     return substations.query("LINES != 0").copy()
 
 
+def filter_lines_with_unavailable_substations(lines):
+    """Filter lines with SUB_1 or SUB_2 attribute equal to 'NOT AVAILABLE', and report
+    the number dropped.
+
+    :param pandas.DataFrame lines: data frame of all lines.
+    :return: (*pandas.DataFrame*) -- lines with available substations.
+    """
+    num_lines = len(lines)
+    filtered = lines.query("SUB_1 == 'NOT AVAILABLE' or SUB_2 == 'NOT AVAILABLE'")
+    num_filtered = len(filtered)
+    print(
+        f"dropping {num_filtered} lines with one or more substations listed as "
+        f"'NOT AVAILABLE' out of a starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 != 'NOT AVAILABLE' and SUB_2 != 'NOT AVAILABLE'").copy()
+
+
 def build_transmission():
     """Main user-facing entry point."""
     # Load input data
@@ -54,3 +71,6 @@ def build_transmission():
     # Filter substations
     substations_with_lines = filter_substations_with_zero_lines(hifld_substations)
     check_for_location_conflicts(substations_with_lines)
+
+    # Filter lines
+    lines_with_substations = filter_lines_with_unavailable_substations(hifld_lines)
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -10,6 +10,20 @@ from prereise.gather.griddata.hifld.data_access.load import (
 )
 
 
+def check_for_location_conflicts(substations):
+    """Check for multiple substations with identical lat/lon.
+
+    :param pandas.DataFrame substations: data frame of substations.
+    :raises ValueError: if multiple substations with identical lat/lon.
+    """
+    num_substations = len(substations)
+    num_lat_lon_groups = len(substations.groupby(["LATITUDE", "LONGITUDE"]))
+    if num_lat_lon_groups != num_substations:
+        num_collisions = num_substations - num_lat_lon_groups
+        raise ValueError(
+            f"There are {num_collisions} substations with duplicate lat/lon values"
+        )
+
 def filter_substations_with_zero_lines(substations):
     """Filter substations with LINES attribute equal to zero, and report the number
     dropped.
@@ -39,3 +53,4 @@ def build_transmission():
 
     # Filter substations
     substations_with_lines = filter_substations_with_zero_lines(hifld_substations)
+    check_for_location_conflicts(substations_with_lines)
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -128,6 +128,23 @@ def filter_lines_with_nonmatching_substation_coords(lines, substations, threshol
     return remaining.copy()
 
 
+def filter_lines_with_identical_substation_names(lines):
+    """Filter lines with SUB_1 or SUB_2 attributes equal to each other, and report the
+    number dropped.
+
+    :param pandas.DataFrame lines: data frame of lines.
+    :return: (*pandas.DataFrame*) -- lines with distinct substations.
+    """
+    num_lines = len(lines)
+    filtered = lines.query("SUB_1 == SUB_2")
+    num_filtered = len(filtered)
+    print(
+        f"dropping {num_filtered} lines with matching SUB_1 and SUB_2 out of a "
+        f"starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 != SUB_2").copy()
+
+
 def build_transmission():
     """Main user-facing entry point."""
     # Load input data
@@ -150,3 +167,7 @@ def build_transmission():
     lines_with_matching_substations = filter_lines_with_nonmatching_substation_coords(
         lines_with_matching_substations, substations_with_lines
     )
+    lines_with_matching_substations = filter_lines_with_identical_substation_names(
+        lines_with_matching_substations
+    )
+    return lines_with_matching_substations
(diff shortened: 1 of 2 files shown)


prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+        f"dropping {num_filtered} lines with matching SUB_1 and SUB_2 out of a "
+        f"starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 != SUB_2").copy()
+
+
+def build_transmission():
+    """Main user-facing entry point."""
+    # Load input data
+    hifld_substations = load.get_hifld_electric_substations(BLOB_PATH)  # noqa: F821
```
jon-hagg, line 150: Where is `BLOB_PATH` defined?

danielolsen: It's not, and I didn't have the blob storage URL handy. I guess this is something that could be added to **const.py**.

jon-hagg: Yeah either that or just setting a placeholder value would be nice, so it's clear what the source is and how to override it. Whatever is easier.

danielolsen: It will definitely be fixed before merging this branch. My goal was to get eyes on the rest of the code in the meantime.

kasparm: @danielolsen the location for the files are already in the annotation file:
https://github.com/Breakthrough-Energy/PreREISE/blob/hifld/ATTRIBUTION.md#grid

jon-hagg (approved):

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+    :return: (*pandas.DataFrame*) -- substations with non-zero values for LINES.
+    """
+    num_substations = len(substations)
+    num_substations_without_lines = len(substations.query("LINES == 0"))
+    print(
+        f"dropping {num_substations_without_lines} substations "
+        f"of {num_substations} total due to LINES parameter equal to 0"
+    )
+
+    return substations.query("LINES != 0").copy()
```
BainanXia, line 35: According to the [documentation](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html), I think `query` will return a copy of filtered data frame, `copy` is redundant here.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+    num_substations = len(substations)
+    num_lat_lon_groups = len(substations.groupby(["LATITUDE", "LONGITUDE"]))
+    if num_lat_lon_groups != num_substations:
+        num_collisions = num_substations - num_lat_lon_groups
+        raise ValueError(
+            f"There are {num_collisions} substations with duplicate lat/lon values"
+        )
+
+
+def get_substations_with_nonzero_lines(substations):
```
BainanXia, line 21: I was comparing the [clean](https://github.com/Mingze-Li128/PreREISE/blob/31693d4121a7fa4e856a3567124e8f34b0caab8c/prereise/gather/hiflddata/data_trans.py#L99) function with this and found they didn't filer out not "IN SERVICE" substations. I remember we discussed this before, just wanna get an impression how many substations we filtered out?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
+from powersimdata.utility.distance import haversine
+
+from prereise.gather.griddata.hifld.data_access import load
+
+
+def check_for_location_conflicts(substations):
```
BainanXia, line 6: Does it mean in the original data, the substation list we get after the call of `get_hifld_electric_substations`, has substations with same lat/lon but different names/ids? How did people deal with it originally?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
+from powersimdata.utility.distance import haversine
+
+from prereise.gather.griddata.hifld.data_access import load
```
BainanXia, line 3: Why not load functions explicitly as we usually do?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+    :return: (*pandas.DataFrame*) -- lines with available substations.
+    """
+    num_lines = len(lines)
+    filtered = lines.query("SUB_1 == 'NOT AVAILABLE' or SUB_2 == 'NOT AVAILABLE'")
+    num_filtered = len(filtered)
+    print(
+        f"dropping {num_filtered} lines with one or more substations listed as "
+        f"'NOT AVAILABLE' out of a starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 != 'NOT AVAILABLE' and SUB_2 != 'NOT AVAILABLE'").copy()
```
BainanXia, line 52: Same comment as above, `copy` is redundant here. Also, we might wanna keep the same expression of filtering logic to make it straightforward to read, i.e. either `A or B` or `not A and not B`.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+    matching_names = substations["NAME"]  # noqa: F841
+    filtered = lines.query(
+        "SUB_1 not in @matching_names or SUB_2 not in @matching_names"
+    )
+    num_filtered = len(filtered)
+    print(
+        f"dropping {num_filtered} lines with one or more substations not found in "
+        f"substations table out of a starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 in @matching_names and SUB_2 in @matching_names").copy()
```
BainanXia, line 73: Same comment as above.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def find_distance_to_closest_end_substation(line, substations):
+        # Coordinates are initially (lon, lat); we reverse to (lat, lon) for haversine
+        end_coordinates = line.loc["COORDINATES"][-1][::-1]
+        end_name = line.loc["SUB_2"]
+        matching_end_substations = substations.loc[substations.NAME == end_name]
+        minimum_end_distance = matching_end_substations.apply(
+            lambda x: haversine(end_coordinates, (x.LATITUDE, x.LONGITUDE)), axis=1
+        ).min()
+        return minimum_end_distance
```
BainanXia, line 105: Maybe we can combine these two similar functions into one. Two options I could think of:
1. Add another optional parameter to specify `start` or `end`
2. Simply don't bother, compare both `start` and `end`, then return the maximum of the two. In this way, we can simplify the filtering logic below as well.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+    ]
+    num_filtered = len(filtered)
+    num_lines = len(lines)
+    print(
+        f"dropping {num_filtered} lines with one or more substations with non-matching "
+        f"coordinates out of a starting total of {num_lines}"
+    )
+
+    remaining = lines.loc[
+        (start_distance_mismatch <= threshold) & (end_distance_mismatch <= threshold)
```
BainanXia, line 125: Same comment as above.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -0,0 +1,171 @@
...
+    :return: (*pandas.DataFrame*) -- lines with distinct substations.
+    """
+    num_lines = len(lines)
+    filtered = lines.query("SUB_1 == SUB_2")
+    num_filtered = len(filtered)
+    print(
+        f"dropping {num_filtered} lines with matching SUB_1 and SUB_2 out of a "
+        f"starting total of {num_lines}"
+    )
+    return lines.query("SUB_1 != SUB_2").copy()
```
BainanXia, line 144: Same comment as above.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: No reason. Do you think an explicit import would be clearer here?

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: The check comes out clean for this iteration of the HIFLD input data. I'm following the logic in the collaborators' branch, which includes such an input validation check.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Filtering to only `IN SERVICE` yields 72,091 substations, out of a starting total of 75,328.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I think we do need this, if only to prevent unexpected warning messages. If we try to set the value on a data frame created via `query`, we get a `SettingWithCopyWarning`.
```python
>>> import pandas as pd
>>> df = pd.DataFrame({"foo": ["a", "b"], "bar": [1, 2]})
>>> df
  foo  bar
0   a    1
1   b    2
>>> q = df.query("foo == 'a'")
>>> q
  foo  bar
0   a    1
>>> q["bar"] = 3
<stdin>:1: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
>>> q
  foo  bar
0   a    3
>>> df
  foo  bar
0   a    1
1   b    2
>>> q.loc[0, "bar"] = 4
C:\Python39\lib\site-packages\pandas\core\indexing.py:1817: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  self._setitem_single_column(loc, value, pi)
>>> q
  foo  bar
0   a    4
>>> df
  foo  bar
0   a    1
1   b    2
>>> q2 = df.query("foo == 'a'").copy()
>>> q2["bar"] = 5
>>> q2.loc[0, "bar"] = 6
>>>
```

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I like this idea, I'll implement it.

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: That's more than 3k substations. In this way, I presume our output csvs will be much different from theirs.

danielolsen: This filtering originally went in via @rouille's #196. I agree that we will end up with a different set of substations. However, based on previous conversations, there are also other changes we anticipate making, so I expect that our code will produce different output CSVs for the bus, branch, and plant tables as well. I guess there are two questions:
1. Do we want to be including substations with statuses like `INACTIVE` (14), `NOT AVAILABLE` (3178), `PROPOSED` (5), and `UNDER CONST` (40)?
2. Should our PRs into the `hifld` branch be based on what we currently think are the most reasonable assumptions, or based on a faithful reproduction of the old branch (with potential improvements after all existing functionality is integrated)?

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: Fair enough. Although I don't quite understand why it throws this warning even we set values on a copy of the get return, which is guaranteed by `query` unless `inplace` is set.

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: 1. No
2. Yes, it should definitely based on what we currently think are the most reasonable assumptions.
I put the note down here just in case if we would like to compare our csvs with theirs at some point in the future, here is one position where we start deviating from their track..

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I think if `inplace=True`, then the return is `None` and the original data frame is modified; otherwise, we _may_ get a 'view' of the DataFrame, rather than a copy. I guess either one will be a DataFrame object, but there are differences in some of the underlying attributes?
https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy

BainanXia: My references are here:
1. pandas documentation, where it says explicitly "a copy" under the inplace parameter doc strings. 
image
2. The accepted answer of this post: https://stackoverflow.com/questions/23296282/what-rules-does-pandas-use-to-generate-a-view-vs-a-copy

danielolsen: I'll admit, I don't fully understand the internal pandas logic, and I'm poking at a semi-opaque box via toy examples. Maybe there's a difference between a copy of a slice of a dataframe vs. a copy of a dataframe?

BainanXia: I don't know either ;) always learning along the way. I'm good with taking an explicit copy regardlessly.

prereise/gather/griddata/hifld/data_process/transmission.py
kasparm: What is the meaning of `NOT AVAILABLE`? Would it be useful to toggle some filters so we can test some combinations?

danielolsen: I'm not sure the meaning of the `NOT AVAILABLE` substations. Since this filtering comes from the `load` module, I think modifying it is out of scope of this PR.

kasparm: Sure, I was just wondering if that would be of use. But `NOT AVAILABLE` is a small percentage so maybe that is not that useful then.
Regarding your two questions I agree with @BainanXia

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I implemented something similar: one function, different parameters.

prereise/gather/griddata/hifld/data_process/transmission.py

prereise/gather/griddata/hifld/data_process/transmission.py

BainanXia (approved):
Thanks!

PR #202, opened 2021-08-17
feat: identify closest substations by ID, not just name

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Currently, for each line, we look up the closest endpoint substations based on matches of the `NAME` attribute to validate that there's at least one matching name that's within a reasonable distance, otherwise this line is discarded. This modification leverages this step to at the same time also grab the unique ID of the closest matching substation, which will be useful in further processing, since the `NAME` is not a unique identifier.

Later, we will want to build up a graph of the network using the substation IDs as node labels, rather than the substation names. This is analogous to what the collaborators did, except that they used lat/lon pairs instead of IDs (both are unique after `LINES > 0` filtering).

### What the code is doing
The `find_distance_to_closest_substation` function is renamed `find_closest_substation_and_distance` and now returns a pandas Series instead of a float, so that the dataframe `apply` calls on this function return a pandas DataFrame rather than a Series. We use the distance column as before, and we also add the substation ID columns to the returned data frame.

### Testing
Tested manually.

### Usage Example/Visuals
```python
>>> from prereise.gather.griddata.hifld.data_process.transmission import build_transmission
>>> lines = build_transmission()
dropping 6892 substations of 70857 total due to LINES parameter equal to 0
dropping 759 lines with one or more substations listed as 'NOT AVAILABLE' out of a starting total of 71554
dropping 3143 lines with one or more substations not found in substations table out of a starting total of 70795
Evaluating endpoint location mismatches... (this may take several minutes)
dropping 120 lines with one or more substations with non-matching coordinates out of a starting total of 67652
dropping 143 lines with matching SUB_1 and SUB_2 out of a starting total of 67532
>>> lines.columns
Index(['ID', 'TYPE', 'STATUS', 'NAICS_CODE', 'NAICS_DESC', 'SOURCE',
       'SOURCEDATE', 'VAL_METHOD', 'VAL_DATE', 'OWNER', 'VOLTAGE',
       'VOLT_CLASS', 'INFERRED', 'SUB_1', 'SUB_2', 'COORDINATES', 'SUB_1_ID',
       'SUB_2_ID'],
      dtype='object')
```

Validating that the ID is a better identifier of substations:
```python
>>> from prereise.gather.griddata.hifld import const
>>> from prereise.gather.griddata.hifld.data_access.load import get_hifld_electric_substations
[...]

PR #204, opened 2021-08-18
refactor: use groupby to speed up distance calculations

danielolsen:

### Purpose
Speed up the substation matching/distance function introduced in #199 and modified in #202. No change in logic, just in implementation.

### What the code is doing
Within `filter_lines_with_nonmatching_substation_coords` we call `find_closest_substation_and_distance` N times, where `N` is the number of lines.

- Before: In each call to `find_closest_substation_and_distance` we filter the entire substations dataframe to find substations with a given name.
- After: within `filter_lines_with_nonmatching_substation_coords` we group the substations by name once, and then within `find_closest_substation_and_distance` we just call `get_group` on the already-grouped-by-name substations `GroupBy` object.

I would expect that filtering a dataframe is O(n) (`n` is the number of rows) but that getting a group is probably O(1) (if it's dict-like), which explains the dramatic speed-up.

### Testing
Tested manually. Before, `filter_lines_with_nonmatching_substation_coords` took approximately 10 minutes, now it takes approximately 2 minutes. The inefficient filtering was giving us >=400% overhead on the distance calculations themselves.

### Usage Example/Visuals
```python
lines = build_transmission()
```

### Time estimate
5-10 minutes.

### DIFF
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -93,24 +93,25 @@ def filter_lines_with_nonmatching_substation_coords(lines, substations, threshol
     :return: (*pandas.DataFrame*) -- lines with matching substations.
     """
 
-    def find_closest_substation_and_distance(coordinates, name, substations):
-        matching_substations = substations.loc[substations.NAME == name]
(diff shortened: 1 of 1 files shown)


BainanXia (approved):
LGTM

PR #211, opened 2021-09-01
ci: replace docs trigger with generic one

jon-hagg:
### Purpose
Switch to the generic `workflow-trigger` action to be consistent with other repos, and will allow deprecating the `publish-docs` action. 

### What the code is doing
Copy paste the action from PowerSimData and removed the docker build which doesn't apply here.

### Testing
n/a

### Time estimate
2 min


rouille (approved):
LGTM

BainanXia (approved):
Thanks

PR #216, opened 2021-09-03
feat: add functions to connect islands with minimum cost in topology module

BainanXia:

### Purpose
Implement functions to connect islands, i.e. connected components, together into a single connected component with minimum cost.

### What the code is doing
In `const.py`, a new dict `abv_state_neighbor` defines neighbor states of each state by abbreviation is added. Note that this dict is created based on ground truth.

A new module under `topology.py`  under `hifld/data_process` is created, which consists of three functions
- `min_dist_of_2_conn_comp`: it calculates the minimum distance of two connected components by exhaustive search over every pair of nodes within the two islands defined by the two node lists.
- `find_adj_state_of_2_conn_comp`: it finds the adjacency states of two connected components against each other based on user specified adjacency relationship.
- `connect_islands_with_minimum_cost`: This is the main user facing function. It connects a group of connected components defined by line and substation data frames into one single island with the help of the two functions above. The algorithm is as following: 
    - Sort all connected components by size in ascending order
    - Loop through all 2 combinations of connected components
        - Given two islands, identify adjacency states using `find_adj_state_of_2_conn_comp`
        - Find minimum distance between two islands based on nodes within adjacency states
        - Add the edge with minimum distance into a potential edge list, mark the start and end points as corresponding connected component index.
     - Build a Minimum Spanning Tree based on the edge list
     - Return the full edge list and the edges within the MST.

### Testing
Unit tests for each function within the `topology` module are added in `test_topology.py`

Also tested manually using the current returns of `build_transmission` function, which returns 798 edges as expected (799 islands in total, indexes are ranked based on island size in ascending order)

Note that it takes around 2.5 hours to get the results. Problem size N = 799. Time complexity O(N**2*V**2)+O(Elog(V)). 
After pruning, total number of potential edges reduced from 318801 (799*798/2) down to 60019, then 798 out of 60019 edges are chosen in the final MST which gives minimum km transmission lines added in order to connect all 798 islands together.

### Where to look
```
prereise/gather/griddata/hifld/const.py
prereise/gather/griddata/hifld/data_process/topology.py
prereise/gather/griddata/hifld/data_process/tests/test_topology.py
```

### Usage Example/Visuals
```python
lines, subs = build_transmission()
lines.set_index("ID")
subs.set_index("ID")
all_edges, mst_edges = connect_islands_with_minimum_cost(lines, subs)
```
`all_edges`: 
[(0, 1, {'weight': 214.95101429233497, 'start': 209798.0, 'end': 209809.0}),
 (0, 2, {'weight': 38.54542121816697, 'start': 200352.0, 'end': 209799.0}),
 (0, 3, {'weight': 131.76661335343925, 'start': 209798.0, 'end': 200726.0}),
 (0, 4, {'weight': 36.18031310851549, 'start': 200352.0, 'end': 200790.0}),
 (0, 5, {'weight': 287.3617444712553, 'start': 200352.0, 'end': 202058.0}),
 (0, 6, {'weight': 180.5714132569954, 'start': 200352.0, 'end': 206944.0}),
 (0, 7, {'weight': 189.0843708698131, 'start': 200352.0, 'end': 205792.0}),
 (0, 8, {'weight': 270.81396338422655, 'start': 200352.0, 'end': 202144.0}),
 (0, 9, {'weight': 293.7457905627537, 'start': 209798.0, 'end': 202147.0}),
...
`mst_edges`:
[(0, 427, {'weight': 15.455266544025594, 'start': 200352.0, 'end': 200683.0}),
 (1, 798, {'weight': 6.208971094154102, 'start': 200594.0, 'end': 200602.0}),
 (2, 768, {'weight': 5.5379898350458605, 'start': 209799.0, 'end': 201219.0}),
 (3, 798, {'weight': 24.520859113861935, 'start': 200726.0, 'end': 200469.0}),
 (4, 26, {'weight': 5.850891966321624, 'start': 209806.0, 'end': 200718.0}),
 (4, 428, {'weight': 1.8072005167497585, 'start': 200790.0, 'end': 209805.0}),
 (5, 798, {'weight': 5.500516607957389, 'start': 202059.0, 'end': 200944.0}),
 (6, 589, {'weight': 6.691353162232742, 'start': 202096.0, 'end': 202236.0}),
 (7, 798, {'weight': 8.344506805655724, 'start': 205792.0, 'end': 201973.0}),
 (8, 798, {'weight': 24.362086818563245, 'start': 202144.0, 'end': 201383.0}),
...
Full list of these two variables are cached in `Breakthrough Energy Dropbox/BreakthroughEnergySciences/Explorations/BainanX/connect_islands_edges_cache`
### Time estimate
30-60 mins, 15 mins to understand the code, 30 mins to read the logic, forever trying to improve computational performance using any greedy heuristic.


prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -0,0 +1,119 @@
...
+        cc1_adj, cc2_adj = find_adj_state_of_2_conn_comp(cc1, cc2, state_neighbor, subs)
+        if len(cc1_adj) > 0:
+            nodes1 = subs.loc[cc1].query("STATE in @cc1_adj")[["LATITUDE", "LONGITUDE"]]
+            nodes2 = subs.loc[cc2].query("STATE in @cc2_adj")[["LATITUDE", "LONGITUDE"]]
+            # Run an exhaustive search on the filtered substations of both connected
+            # components to find a line with minimum cost the connects the two islands
+            min_dist, sub1, sub2 = min_dist_of_2_conn_comp(nodes1, nodes2, cost_metric)
+            edge_list.append(
+                (ind1, ind2, {"weight": min_dist, "start": sub1, "end": sub2})
+            )
```
danielolsen, line 114: This `append` inside a loop seems potentially inefficient. Could we potentially speed this up with an `apply` or something similar which would allocate memory once, instead of n^2 times?

BainanXia: This is minor. The true complexity here is n^2*v^2. We only do append operations in number of potential edges times, i.e. 60019. I tested locally by replacing `edge_list` by a hash map, and didn't observe noticeable difference, so I kept a cleaner implementation here.

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -0,0 +1,119 @@
...
+    :param func dist_metric: a function defines the distance metric.
+    :return: (*tuple*) -- a tuple contains three elements, the minimum distance
+        follows by a pair of nodes from the two connected components respectively
+        in order
+    """
+    min_dist = float("inf")
+    for v1, v2 in product(nodes1.index, nodes2.index):
+        tmp_dist = dist_metric(nodes1.loc[v1], nodes2.loc[v2])
+        if tmp_dist < min_dist:
+            min_dist, res_v1, res_v2 = tmp_dist, v1, v2
```
danielolsen, line 29: Can we potentially speed this up with chained `apply` calls to fill a dataframe and then the use of `min` & `idxmin` to select the winners?

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -0,0 +1,119 @@
...
+    return min_dist, res_v1, res_v2
+
+
+def find_adj_state_of_2_conn_comp(cc1, cc2, state_adj, subs):
+    """Find the adjacency states between two given connected components.
```
jon-hagg, line 34: adjacency -> adjacent?

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -0,0 +1,119 @@
...
+        component that are adjacent to the other one.
+    """
+    cc1_all_state = set(subs.loc[cc1].STATE.unique())
+    cc2_all_state = set(subs.loc[cc2].STATE.unique())
+    cc1_neighbors = cc1_all_state.copy()
+    cc2_neighbors = cc2_all_state.copy()
+    for s in cc1_all_state:
+        cc1_neighbors |= set(state_adj[s])
+    for s in cc2_all_state:
+        cc2_neighbors |= set(state_adj[s])
```
danielolsen, line 52: Is this equivalent to:
```python
cc1_neighbors = cc1_all_state.union(*[set(state_adj[s]) for s in cc1_all_state])
cc2_neighbors = cc2_all_state.union(*[set(state_adj[s]) for s in cc2_all_state])
```
?

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: I did following test and they looks similar:

image

Potential reasons of `apply` being slower is calculating `min` and `idxmin` (not included in the test above yet) needs another two sweep of O(N).

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -0,0 +1,119 @@
...
+        states, defaults to None and the constant dictionary ``abv_state_neighbor``
+        defined in the ``const`` module is used.
+    :return: (*tuple*) -- a pair of lists, the first one is a list of all potential
+        line candidates among given connected components; the second one is a list
+        of subsequence of the first entry, representing the chosen lines to form a
+        minimum spanning tree. Each entry of both lists is a 3-tuple, index of the
+        first connected component, index of the second connected component,
+        a dictionary with keys containing ``start``, ``end`` and ``weight``, defines
+        the ``from substation ID``, ``to substation ID`` and the weight of the line
+        calculated by the ``cost_metric``.
```
jon-hagg, line 82: Curious what the use case is for returning the full edge list, I would expect the main goal is to get the MST.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: I did following test and they looks similar:
image
Potential reasons of apply being slower is calculating ``min`` and
idxmin
(not included in the test above yet) needs another two sweep of O(N).

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: We had an offline discussion via Slack regarding caching the full "matrix" for potential subsequent usage. I don't know how we gonna use it at the moment though.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Nice and done.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Done.

prereise/gather/griddata/hifld/data_process/topology.py
danielolsen: I think we only want to call `min` and `idxmin` on a full dataframe that we construct with the chained applies, not within one of the applies to end up with a Series. We may still not achieve any/much speedup though.

danielolsen: Running the following code to simulate the heaviest computation combination, I see a ~6x speedup with an `apply` version:
```
cc1 = all_cc[-1]
cc2 = all_cc[-2]
nodes1 = subs.loc[subs.STATE.isin({"ME", "NH"}) & subs.index.isin(cc1), ["LATITUDE", "LONGITUDE"]]
nodes2 = subs.loc[cc2, ["LATITUDE", "LONGITUDE"]]
start_time = timeit.default_timer()
min_dist, sub1, sub2 = min_dist_of_2_conn_comp(nodes1, nodes2, haversine)
print(timeit.default_timer() - start_time)
start_time = timeit.default_timer()
distances = nodes1.apply(lambda x: nodes2.apply(lambda y: haversine(x, y), axis=1), axis=1)
nodes1_min = distances.min(axis=1).idxmin()
nodes2_min = distances.loc[nodes1_min].idxmin()
distance_min = distances.loc[nodes1_min, nodes2_min]
print(timeit.default_timer() - start_time)
```
The first one executes in 1.2-1.3 seconds on my machine, compared to 0.2 seconds for the second one.

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -0,0 +1,115 @@
...
+    edge_list = []
+    for ind1, ind2 in tqdm(combinations(range(len(all_cc)), r=2)):
+        cc1 = all_cc[ind1]
+        cc2 = all_cc[ind2]
+        # Find the adjacency states within the two connected components against each
+        # other
+        cc1_adj, cc2_adj = find_adj_state_of_2_conn_comp(cc1, cc2, state_neighbor, subs)
+        if len(cc1_adj) > 0:
+            nodes1 = subs.loc[cc1].query("STATE in @cc1_adj")[["LATITUDE", "LONGITUDE"]]
+            nodes2 = subs.loc[cc2].query("STATE in @cc2_adj")[["LATITUDE", "LONGITUDE"]]
```
danielolsen, line 104: `query` inside a big loop always makes me nervous. If we outside the loop call:
```python
state_subs = pd.Series(subs.groupby("STATE").groups).map(set)
```
and in the loop call:
```python
nodes1 = subs.loc[
    subs.index.isin(cc1 & set().union(*[state_subs[s] for s in cc1_adj]),
    ["LATITUDE", "LONGITUDE"]
]
```
does that improve performance at all?

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: If we don't calculate minimum on the fly, that's indeed faster. This is actually my previous implementation. I wasn't able to test the computational performance of it due to running out of memory (with other stuffs running in parallel). I think here is a clear trade-off between time complexity and memory complexity. I'm fine with either one. What do you think @rouille @jon-hagg ?

danielolsen: If we don't filter the big CC to just the appropriate states, we see a similar level of speedup:
```python
>>> cc1 = all_cc[-1]
>>> cc2 = all_cc[-2]
>>> nodes1 = subs.loc[cc1, ["LATITUDE", "LONGITUDE"]]
>>> nodes2 = subs.loc[cc2, ["LATITUDE", "LONGITUDE"]]
>>> start_time = timeit.default_timer()
>>> min_dist, sub1, sub2 = min_dist_of_2_conn_comp(nodes1, nodes2, haversine)
>>> print(timeit.default_timer() - start_time)
185.3390242999999
>>> start_time = timeit.default_timer()
>>> distances = nodes1.apply(lambda x: nodes2.apply(lambda y: haversine(x, y), axis=1), axis=1)
>>> nodes1_min = distances.min(axis=1).idxmin()
>>> nodes2_min = distances.loc[nodes1_min].idxmin()
>>> distance_min = distances.loc[nodes1_min, nodes2_min]
>>> print(timeit.default_timer() - start_time)
27.511743399999887
```

Even in this intentionally inefficient case, the `distances` data frame is dimension (50607, 41), which for a float64 dtype seems to be roughly 17 MB (per some back-of-the-envelope calculations).

BainanXia: We have going to have hundreds of such data frame though unless we clean up memory on the fly.

danielolsen: I'm not an expert in memory management, but won't the old data frame be automatically discarded (and its memory freed) after the function which creates it returns, or the next time that we re-define a new object into that variable name? I think python's garbage collector should handle that automatically.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Yeah, I used chain indexing instead of query, which is kind of equivalent here and again no noticeable performance improvement. `query` is supposed to be slower due to string evaluation, but not that bad, reference post [here](https://stackoverflow.com/questions/49936557/pandas-dataframe-loc-vs-query-performance/49937318).

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: I'm not an expert either! Python indeed automatically release memory after function call but not all of them, even in the case an object loses references completely. The logic behind that could be complex, my quick googling finds following [post](https://stackoverflow.com/questions/15455048/releasing-memory-in-python). 

The story is I ran out of memory in a resource limited situation during implementation, thus, I switched to another one. I have no strong preferences in such trade-offs.

danielolsen (approved):
The code to generate the MST looks good.

For the lines we've found to add, we have substation ID endpoints. How do you envision this function fitting in with the rest of the `build_transmission` chain? Will we run this function, give the new lines `{"VOLTAGE": pd.NA, "VOLT_CLASS": "NOT AVAILABLE"}`, and then run `augment_line_voltages`?

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: I've added an optional parameter `memory_efficient` for function `min_dist_of_2_conn_comp` and the user facing function `connect_islands_with_minimum_cost` which passes the parameter through. Once the parameter is set, the current implementation is triggered. Otherwise, the default behavior is using `apply`.

BainanXia:
> The code to generate the MST looks good.
> 
> For the lines we've found to add, we have substation ID endpoints. How do you envision this function fitting in with the rest of the `build_transmission` chain? Will we run this function, give the new lines `{"VOLTAGE": pd.NA, "VOLT_CLASS": "NOT AVAILABLE"}`, and then run `augment_line_voltages`?

That aligns with my intention. 
- Start with initial filtering of raw data with either approach, the current one or the one @rouille is implementing.
- Connect all islands together to form a big connected component.
- Assign voltages to lines with missing info.
- Build bus/bus2sub table according to line table.

PR #210, opened 2021-08-27
feat: add function to create buses from substations and lines

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Address #205 by starting the bus data frame. ~This PR depends on #208, which is why it is currently listed as the PR target (so that only the changes since then are shown), but this PR will eventually target the `hifld` branch.~ EDIT: #208 has been merged, so this targets `hifld` now.

### What the code is doing
- The `lines` and `substations` data frames are used to identify the set of voltages attaches to each substation.
- This set is exploded to yield a Series entry for (substation, voltage) combination.
- The index is set to be a column, then this column is popped out to the data frame to become the `bus2sub` mapping.

### Testing
A unit test has been added to demonstrate/verify the functionality on a simple test case.

### Usage Example
We can see that `bus` and `bus2sub` are created as intended, and that there are more unique buses than unique substations (i.e. multiple buses are created within substations)
```python
>>> from prereise.gather.griddata.hifld.data_process.transmission import build_transmission
>>> from prereise.gather.griddata.hifld.data_process.transmission import create_buses
>>> lines, substations = build_transmission()
dropping 6892 substations of 70857 total due to LINES parameter equal to 0
dropping 759 lines with one or more substations listed as 'NOT AVAILABLE' out of a starting total of 71554
dropping 3143 lines with one or more substations not found in substations table out of a starting total of 70795
Evaluating endpoint location mismatches... (this may take several minutes)
dropping 120 lines with one or more substations with non-matching coordinates out of a starting total of 67652
dropping 143 lines with matching SUB_1 and SUB_2 out of a starting total of 67532
2119 line voltages can't be found via neighbor consensus
310 line voltages can't be found via neighbor minimum
>>> bus, bus2sub = create_buses(lines)
>>> bus
0         69.0
1        161.0
2         69.0
3        115.0
4        115.0
         ...
58279    138.0
58280    138.0
58281    345.0
58282    138.0
58283    345.0
Name: baseKV, Length: 58284, dtype: float64
>>> bus2sub
0        107655.0
1        107655.0
2        107656.0
3        107656.0
4        107657.0
           ...
58279    310905.0
58280    310906.0
58281    310907.0
58282    310908.0
58283    310924.0
Name: sub_id, Length: 58284, dtype: float64
>>> len(substations)
63965
>>> len(bus2sub.unique())
52885
>>> bus.value_counts(dropna=False)
69.0     19852
[...]

PR #217, opened 2021-09-07
feat: add transmission to substation track selection

rouille:

### Purpose
Select HIFLD substations based on HIFLD lines

### What the code is doing
* `map_lines_to_substations_using_coords`: create a data frame mapping lines to substations base on coordinates
* `assign_substations_to_lines`: update `lines` and `substations` data frames. In particular: 
   * update *SUB_1* and *SUB_2* columns in `lines` data frame (name of selected `substations` are used) 
   * new *SUB_1_ID*, *SUB_2_ID* columns giving the selected substation ID
   * new *OTHER_SUB_1_ID* and *OTHER_SUB_2_ID* indicating other candidates at same endpoint, either a substation, tap, riser or dead end
 * `build_transmission`: the function has been updated to use the above method to get the lines/substations tuple

### Testing

### Where to look
All the code is in the `transmission` module 

### Usage Example/Visuals
```
>>> from prereise.gather.griddata.hifld.data_process.transmission import map_lines_to_substations_using_coords, build_transmission
>>> lines, substations  = build_transmission(method="line2sub", kwargs={"rounding": 2})
filter substations based on lines
initial number of substations: 63965
substations with same location after rounding: 13769

initial number of lines: 71554
zero distance lines after rounding: 11589

there are 1117 lines with two substations missing
there are 1848 lines with one substation missing
there are 2042 unique orphan endpoint
there are 7622 substations unconnected

finding closest neighbor to unconnected lines
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2965/2965 [02:50<00:00, 17.43it/s]
2661 line voltages can't be found via neighbor consensus
238 line voltages can't be found via neighbor minimum
>>> lines
                TYPE      STATUS  NAICS_CODE                                    NAICS_DESC  ... SUB_1_ID OTHER_SUB_1_ID SUB_2_ID    OTHER_SUB_2_ID
ID                                                                                          ...                                                   
212144  AC; OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   200352           None   209798              None
212145  AC; OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   200356       [201930]   200356          [201930]
212146  AC; OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   200594           None   200593  [209809, 203487]
212147  AC; OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   200642       [209799]   200642          [209799]
212148  AC; OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   200656           None   209800              None
...              ...         ...         ...                                           ...  ...      ...            ...      ...               ...
175393      OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   173638           None   132385              None
175394      OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   141104           None   173640              None
175395      OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   172406           None   173641              None
175396      OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   173641           None   173640              None
175397      OVERHEAD  IN SERVICE      221121  ELECTRIC BULK POWER TRANSMISSION AND CONTROL  ...   173642           None   141147              None

[71554 rows x 19 columns]
>>> substations
                   X             Y            NAME            CITY STATE    ZIP        TYPE  ... VAL_METHOD             VAL_DATE  LINES MAX_VOLT  MIN_VOLT  MAX_INFER  MIN_INFER
ID                                                                                           ...                                                                                
131072 -1.069084e+07  5.898522e+06   UNKNOWN131072         AUDUBON    MN  56511  SUBSTATION  ...    IMAGERY  2015/08/28 00:00:00      1    115.0     115.0          N          N
131089 -9.311883e+06  5.323699e+06   UNKNOWN131089     GENESEE TWP    MI  48506  SUBSTATION  ...    IMAGERY  2015/08/17 00:00:00      1    138.0     138.0          Y          Y
131090 -1.049062e+07  5.162095e+06  GRAND JUNCTION  GRAND JUNCTION    IA  50107  SUBSTATION  ...    IMAGERY  2018/11/28 00:00:00      4    161.0     161.0          Y          Y
131091 -1.049057e+07  5.161922e+06   UNKNOWN131091  GRAND JUNCTION    IA  50107  SUBSTATION  ...    IMAGERY  2015/08/18 00:00:00      3    161.0 -999999.0          Y          N
131092 -1.047791e+07  5.136316e+06   UNKNOWN131092           PERRY    IA  50220  SUBSTATION  ...    IMAGERY  2015/09/11 00:00:00      1    161.0     161.0          Y          Y
...              ...           ...             ...             ...   ...    ...         ...  ...        ...                  ...    ...      ...       ...        ...        ...
131067 -1.034558e+07  5.161820e+06   UNKNOWN131067    MARSHALLTOWN    IA  50158  SUBSTATION  ...    IMAGERY  2015/09/10 00:00:00      2    161.0     161.0          Y          Y
131068 -1.034633e+07  5.144523e+06   UNKNOWN131068          LAUREL    IA  50141  SUBSTATION  ...    IMAGERY  2015/08/13 00:00:00      3    161.0 -999999.0          Y          N
131069 -1.035529e+07  5.125052e+06          JASPER          NEWTON    IA  50208  SUBSTATION  ...    IMAGERY  2018/11/28 00:00:00      3    161.0     161.0          Y          Y
131070 -1.071809e+07  5.914971e+06   UNKNOWN131070          HAWLEY    MN  56549  SUBSTATION  ...    IMAGERY  2015/08/28 00:00:00      1     69.0      69.0          Y          Y
131071 -1.048157e+07  5.915785e+06   UNKNOWN131071      PINE RIVER    MN  56474  SUBSTATION  ...    IMAGERY  2015/08/28 00:00:00      3     69.0 -999999.0          Y          N

[54492 rows x 24 columns]
```

### Time estimate
30min

### DIFF
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -1,8 +1,11 @@
 import os
 
 import networkx as nx
+import numpy as np
 import pandas as pd
-from powersimdata.utility.distance import haversine
+from powersimdata.utility.distance import haversine, ll2uv
+from scipy.spatial import KDTree
+from tqdm import tqdm
 
 from prereise.gather.griddata.hifld import const
 from prereise.gather.griddata.hifld.data_access.load import (
@@ -25,6 +28,114 @@ def check_for_location_conflicts(substations):
         raise ValueError(
             f"There are {num_collisions} substations with duplicate lat/lon values"
         )
+
+
+def map_lines_to_substations_using_coords(substations, lines, **kwargs):
+    """Map lines to substations using coordinates.
+
+    :param pandas.DataFrame substations: data frame of substations.
+    :param pandas.DataFrame lines: data frame of all lines.
+    :param \\*\\*kwargs: optional arguments.
+    :return: (*tuple*) -- lines and substations data frame.
+    """
+    # Optional parameters
+    rounding = 3 if "rounding" not in kwargs else kwargs["rounding"]
+    drop_zero_distance_line = (
+        True
+        if "drop_zero_distance_line" not in kwargs
+        else kwargs["drop_zero_distance_line"]
+    )
+
+    # Round coordinates of substations and lines' endpoints
+    print(
+        "assigning substations to lines' endpoints by mapping their rounded "
+        f"({rounding} digits) coordinates"
+    )
+    subcoord2subid = (
+        substations.round(rounding).groupby(["LATITUDE", "LONGITUDE"]).groups
+    )
+    lines_coord = lines["COORDINATES"].map(
+        lambda x: list(np.round([x[0], x[-1]], rounding))
+    )
+    line2sub = pd.DataFrame(
+        lines_coord.to_list(), columns=["FROM", "TO"], index=lines_coord.index
+    ).applymap(tuple)
+
+    # Map coordinates of lines' endpoints to coordinates of substations and add id of
+    # matched substations to data frame. Values can be single id, a list of id or NaN
+    # (no match)
+    end_sub = {"FROM": "SUB_1", "TO": "SUB_2"}
+    for e, e_sub in end_sub.items():
+        line2sub[e_sub] = line2sub[e].apply(
+            lambda x: list(subcoord2subid[x]) if x in subcoord2subid else np.NaN
+        )
+
+    # Remove zero-distance lines
+    if drop_zero_distance_line:
+        idx = line2sub["FROM"].compare(line2sub["TO"]).index
+        print(
+            f"dropping {lines.shape[0] - len(idx)} lines having same endpoints "
+            "coordinates after rounding"
+        )
+        line2sub = line2sub.loc[idx]
+
+    # Find closest neighbor(s) of unmapped lines' endpoints
+    print("finding closest substation to unmapped lines' endpoint(s)")
+    subcoord = list(subcoord2subid)
+    missing_points = set().union(
+        *[
+            set(line2sub.loc[line2sub[e_sub].isna(), e].apply(lambda x: (x[1], x[0])))
+            for e, e_sub in end_sub.items()
+        ]
+    )
+    tree = KDTree([ll2uv(p[0], p[1]) for p in subcoord])
+    endpoint2neighbor = {
+        (p[1], p[0]): subcoord2subid[subcoord[tree.query(ll2uv(p[0], p[1]))[1]]]
+        for p in tqdm(missing_points, total=len(missing_points))
+    }
+    filled_subs = [
+        line2sub.loc[line2sub[e_sub].isna(), e].map(endpoint2neighbor).map(list)
+        for e, e_sub in end_sub.items()
+    ]
+    line2sub.update(pd.concat(filled_subs, axis=1).rename(columns=end_sub))
+
+    # pick unique substation at each end. Order of preference:
+    # SUBSTATION --> TAP --> RISER --> DEAD END
+    all2one = {
+        tuple(i): None for j in end_sub.values() for i in line2sub[j] if len(i) > 1
+    }
+    for a in all2one:
+        type2id = substations.loc[list(a)].reset_index().groupby("TYPE").first()["ID"]
+        for t in ["SUBSTATION", "TAP", "RISER", "DEAD END"]:
+            try:
+                all2one[a] = type2id.loc[t]
+                break
+            except KeyError:
+                continue
+    one2all = {v: k for k, v in all2one.items()}
+
+    # Build substations data frame
+    new_substations = substations.copy()
+    new_substations = substations.loc[
+        set(pd.concat([line2sub[e] for e in end_sub.values()]).explode())
+    ]
+    new_substations["OTHER_SUB"] = (
+        new_substations.reset_index()["ID"]
+        .apply(lambda x: list(set(one2all[x]) - {x}) if x in one2all else None)
+        .values
+    )
+
+    # Build lines data frame
+    new_lines = lines.copy()
+    new_lines = new_lines.loc[line2sub.index]
+    for e_sub in end_sub.values():
+        primary_sub = line2sub[e_sub].apply(
+            lambda x: all2one[tuple(x)] if len(x) > 1 else x[0]
+        )
+        new_lines[e_sub] = substations.loc[primary_sub.values, "NAME"].values
+        new_lines[f"{e_sub}_ID"] = primary_sub.values
+
+    return new_lines, new_substations
 
 
 def filter_substations_with_zero_lines(substations):
@@ -322,8 +433,26 @@ def create_transformers(bus):
     return pd.DataFrame(bus_pairs, columns=["from_bus_id", "to_bus_id"])
 
 
-def build_transmission():
-    """Main user-facing entry point."""
+def build_transmission(method="sub2line", kwargs={"rounding": 3}):
+    """Build transmission network
+
+    :param str method: method used to build network. Default method is *sub2line*
+        where all substations are considered and lines are filtered accordingly. Other
+        method is *line2sub* where all lines are considered and a list of substations
+        is compiled.
+    :param dict kwargs: keyword arguments to be passed to functions.
+    :raises TypeError:
+        if ``method`` is not a str.
+        if ``kwargs`` is not a dict.
+    :raises ValueError: if ``method`` is unknown.
+    """
+    if not isinstance(method, str):
+        raise TypeError("method must be a str")
+    if not isinstance(kwargs, dict):
+        raise TypeError("kwargs must be a dict")
+    if method not in ["sub2line", "line2sub"]:
+        raise ValueError("Unknown method to build transmission network")
+
     # Load input data
     hifld_substations = get_hifld_electric_substations(const.blob_paths["substations"])
     hifld_substations.set_index("ID", inplace=True)
@@ -338,11 +467,19 @@ def build_transmission():
     substations = filter_substations_with_zero_lines(hifld_substations)
     check_for_location_conflicts(substations)
 
-    # Filter lines based on substations
-    lines = filter_lines_with_unavailable_substations(hifld_lines)
-    lines = filter_lines_with_no_matching_substations(lines, substations)
-    lines = filter_lines_with_nonmatching_substation_coords(lines, substations)
-    lines = filter_lines_with_identical_substation_names(lines)
+    if method == "sub2line":
+        print("filter lines based on substations")
+        print("---------------------------------")
+        lines = filter_lines_with_unavailable_substations(hifld_lines)
+        lines = filter_lines_with_no_matching_substations(lines, substations)
+        lines = filter_lines_with_nonmatching_substation_coords(lines, substations)
+        lines = filter_lines_with_identical_substation_names(lines)
+    elif method == "line2sub":
+        print("filter substations based on lines")
+        print("---------------------------------")
+        lines, substations = map_lines_to_substations_using_coords(
+            substations, hifld_lines, **kwargs
+        )
 
     # Add voltages to lines with missing data
     augment_line_voltages(lines, substations)


danielolsen:
Is there any path dependence between this PR and the #210-#218-#219 chain? I _think_ they're all independent, but maybe we want to ensure that the others still work under both methods of coming up with lines/substations?

rouille:
> Is there any path dependence between this PR and the #210-#218-#219 chain? I _think_ they're all independent, but maybe we want to ensure that the others still work under both methods of coming up with lines/substations?

Yes, I believe they are all independent. We don't have to merge this PR now, we can go over it once we have the whole chain and we want to compare both tracks.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,163 @@ def check_for_location_conflicts(substations):
...
+
+    from_sub_missing = line2sub["FROM_SUB"].isnull().values
+    to_sub_missing = line2sub["TO_SUB"].isnull().values
+
+    unique_endpoint_coord = set(
+        line2sub["FROM"].unique().tolist() + line2sub["TO"].unique().tolist()
+    )
+
+    n_two_sub_missing = line2sub[from_sub_missing & to_sub_missing].shape[0]
+    n_one_sub_missing = line2sub[from_sub_missing ^ to_sub_missing].shape[0]
```
danielolsen, line 80: Do we need to index and then query the shape to count the number missing, or can we just sum the boolean vector after the boolean algebra?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,163 @@ def check_for_location_conflicts(substations):
...
+    n_unique_orphan_endpoint = len(unique_endpoint_coord - set(subcoord2subid))
+    n_uncommected_sub = len(set(subcoord2subid) - unique_endpoint_coord)
```
danielolsen, line 82: `n_uncommected_sub` -> `n_unconnected_sub`?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,163 @@ def check_for_location_conflicts(substations):
...
+    )
+
+    lines_coord_from = lines["COORDINATES"].map(
+        lambda x: (round(x[0][0], rounding), round(x[0][1], rounding))
+    )
+    lines_coord_to = lines["COORDINATES"].map(
+        lambda x: (round(x[-1][0], rounding), round(x[-1][1], rounding))
+    )
+    line2sub = lines_coord_from.to_frame(name="FROM").join(
+        lines_coord_to.to_frame(name="TO"), how="inner"
```
danielolsen, line 57: Do we need to specify `how="inner"`? I think the indices of the two should be identical, so the `how` parameter shouldn't matter.

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: yeah that would be cleaner

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,163 @@ def check_for_location_conflicts(substations):
...
+        and *OTHER_SUB_2_ID* when multiple location are associated to the same endpoint.
+    """
+    print("finding closest neighbor to unconnected lines")
+
+    # find closest neighbor of unconnected lines
+    subcoord2subid = substations.groupby(["LONGITUDE", "LATITUDE"]).groups
+    subcoord = list(subcoord2subid)
+    missing = line2sub[
+        line2sub["FROM_SUB"].isnull() | line2sub["TO_SUB"].isnull()
+    ].index
```
danielolsen, line 113: I think we can fit this on one line as:
```
missing = line2sub.query("FROM_SUB.isnull() or TO_SUB.isnull()").index
```

rouille: indeed

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -303,8 +462,26 @@ def create_buses(lines):
...
-def build_transmission():
-    """Main user-facing entry point."""
+def build_transmission(method="sub2line", kwargs={"rounding": 3}):
+    """Build transmission network
+
+    :param str method: method used to build network. Default method is *sub2line*
+        where all substations are considered and lines are filtered accordingly. Other
+        method is *line2sub* where all lines are considered and a list of substations
+        is compiled.
+    :param dict kwargs: keyword arguments to be passed to functions.
```
danielolsen, line 472: Can we be more specific here, that `kwargs` is only used when `method == "line2sub"`?

rouille: Sure. You don't expect any other keyword arguments for other function(s), e.g. (voltage, topology, etc.) in the future?

rouille: If not, perhaps, I should only have a `rounding` optional argument with `3` as default value instead of unpacking a dictionary

danielolsen: If we have multiple functions' optional keyword arguments contained in one big user-facing dict, then I think we'll need to change the call signature of the lower-level functions, right? So `map_lines_to_substations_using_coords(substations, lines, rounding=3)` would become `map_lines_to_substations_using_coords(substations, lines, **kwargs)`, and then internal to the function we'd have a dict of defaults including `rounding=3` that can supplement if the user-suppled `kwargs` doesn't specify, and then within the function we'd look up `kwargs["rounding"]`. Otherwise, if `map_lines_to_substations_using_coords` gets keyword arguments for other functions (of if they get `rounding`), it'll throw an error.

Either pattern should work I think. If we want to design `build_transmission` using the one-big-dict way, then let's future-proof the first lower-level function that ingests that dict, and illustrate a pattern for other lower-level functions. Otherwise, let's rename `kwargs` to something like `line2sub_kwargs`, or just pass `line2sub_rounding`.

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: I went with the user-facing dictionary

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,162 @@ def check_for_location_conflicts(substations):
...
+    lines_coord_to = lines["COORDINATES"].map(
+        lambda x: (round(x[-1][0], rounding), round(x[-1][1], rounding))
+    )
+    line2sub = lines_coord_from.to_frame(name="FROM").join(
+        lines_coord_to.to_frame(name="TO")
+    )
+
+    for endpoint in ["FROM", "TO"]:
+        line2sub[f"{endpoint}_SUB"] = [
+            i if isinstance(i, float) else list(i)
```
danielolsen, line 63: Is this ternary operator based on the logic that if a line endpoint isn't present in `subcoord2subid`, we'll get `NaN` which is a float, rather than an `int` index?

danielolsen:
Running the latest code, I'm getting an error:
```python
>>> lines, substations = build_transmission(method="line2sub", kwargs={"rounding": 2})
filter substations based on lines
---------------------------------
initial number of substations: 63965
substations with same location after rounding: 13769

initial number of lines: 71554
zero distance lines after rounding: 11589

there are 1117 lines with two substations missing
there are 1848 lines with one substation missing
there are 2042 unique orphan endpoint
there are 7622 substations unconnected

finding closest neighbor to unconnected lines
100%|██████████████████████████████████████| 2965/2965 [02:07<00:00, 23.31it/s]
  File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\hifld\data_process\transmission.py", line 511, in build_transmission
    lines, substations = assign_substations_to_lines(
  File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\hifld\data_process\transmission.py", line 167, in assign_substations_to_lines
    line2sub.loc[i, f"OTHER_{endpoint}_SUB"] = list(
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 723, in __setitem__
    iloc._setitem_with_indexer(indexer, value, self.name)
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1730, in _setitem_with_indexer
    self._setitem_with_indexer_split_path(indexer, value, name)
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1785, in _setitem_with_indexer_split_path
    raise ValueError(
ValueError: Must have equal len keys and value when setting with an iterable
```

Maybe this is because of something upstream?

rouille:
> Running the latest code, I'm getting an error:
> 
> ```python
> >>> from prereise.gather.griddata.hifld.data_process.transmission import build_transmission
> >>> lines, substations = build_transmission(method="line2sub", kwargs={"rounding": 2})
> dropping 6892 substations of 70857 total due to LINES parameter equal to 0
> filter substations based on lines
> ---------------------------------
> initial number of substations: 63965
> substations with same location after rounding: 13769
> 
> initial number of lines: 71554
> zero distance lines after rounding: 11589
> 
> there are 1117 lines with two substations missing
> there are 1848 lines with one substation missing
> there are 2042 unique orphan endpoint
> there are 7622 substations unconnected
> 
> finding closest neighbor to unconnected lines
> 100%|██████████████████████████████████████| 2965/2965 [02:07<00:00, 23.31it/s]
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\hifld\data_process\transmission.py", line 511, in build_transmission
>     lines, substations = assign_substations_to_lines(
>   File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\hifld\data_process\transmission.py", line 167, in assign_substations_to_lines
>     line2sub.loc[i, f"OTHER_{endpoint}_SUB"] = list(
>   File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 723, in __setitem__
>     iloc._setitem_with_indexer(indexer, value, self.name)
>   File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1730, in _setitem_with_indexer
>     self._setitem_with_indexer_split_path(indexer, value, name)
>   File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1785, in _setitem_with_indexer_split_path
>     raise ValueError(
> ValueError: Must have equal len keys and value when setting with an iterable
> ```
> 
> Maybe this is because of something upstream?

It works for me:
```
>>> from prereise.gather.griddata.hifld.data_process.transmission import map_lines_to_substations_using_coords, build_transmission
>>> lines, substations  = build_transmission(method="line2sub")
filter substations based on lines
---------------------------------
initial number of substations: 63965
substations with same location after rounding: 4307

initial number of lines: 71554
zero distance lines after rounding: 3314

there are 1197 lines with two substations missing
there are 2413 lines with one substation missing
there are 2632 unique orphan endpoint
there are 9259 substations unconnected

finding closest neighbor to unconnected lines
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3610/3610 [04:10<00:00, 14.44it/s]
2467 line voltages can't be found via neighbor consensus
296 line voltages can't be found via neighbor minimum
>>> 
```

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: After the mapping, we have in the `FROM_SUB` and `TO_SUB` some NA (`float` type) and some `pandas.Int64Index`. I cast the `Int64Index` to `list` since later I working with some `set`, `itertools`, etc.

rouille:
> > Running the latest code, I'm getting an error:
> > ```python
> > >>> from prereise.gather.griddata.hifld.data_process.transmission import build_transmission
> > >>> lines, substations = build_transmission(method="line2sub", kwargs={"rounding": 2})
> > dropping 6892 substations of 70857 total due to LINES parameter equal to 0
> > filter substations based on lines
> > ---------------------------------
> > initial number of substations: 63965
> > substations with same location after rounding: 13769
> > 
> > initial number of lines: 71554
> > zero distance lines after rounding: 11589
> > 
> > there are 1117 lines with two substations missing
> > there are 1848 lines with one substation missing
> > there are 2042 unique orphan endpoint
> > there are 7622 substations unconnected
> > 
> > finding closest neighbor to unconnected lines
> > 100%|██████████████████████████████████████| 2965/2965 [02:07<00:00, 23.31it/s]
> > Traceback (most recent call last):
> >   File "<stdin>", line 1, in <module>
> >   File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\hifld\data_process\transmission.py", line 511, in build_transmission
> >     lines, substations = assign_substations_to_lines(
> >   File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\hifld\data_process\transmission.py", line 167, in assign_substations_to_lines
> >     line2sub.loc[i, f"OTHER_{endpoint}_SUB"] = list(
> >   File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 723, in __setitem__
> >     iloc._setitem_with_indexer(indexer, value, self.name)
> >   File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1730, in _setitem_with_indexer
> >     self._setitem_with_indexer_split_path(indexer, value, name)
> >   File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1785, in _setitem_with_indexer_split_path
> >     raise ValueError(
> > ValueError: Must have equal len keys and value when setting with an iterable
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe this is because of something upstream?
> 
> It works for me:
> 
> ```
> >>> from prereise.gather.griddata.hifld.data_process.transmission import map_lines_to_substations_using_coords, build_transmission
> >>> lines, substations  = build_transmission(method="line2sub")
> dropping 6892 substations of 70857 total due to LINES parameter equal to 0
> filter substations based on lines
> ---------------------------------
> initial number of substations: 63965
> substations with same location after rounding: 4307
> 
> initial number of lines: 71554
> zero distance lines after rounding: 3314
> 
> there are 1197 lines with two substations missing
> there are 2413 lines with one substation missing
> there are 2632 unique orphan endpoint
> there are 9259 substations unconnected
> 
> finding closest neighbor to unconnected lines
> 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3610/3610 [04:10<00:00, 14.44it/s]
> 2467 line voltages can't be found via neighbor consensus
> 296 line voltages can't be found via neighbor minimum
> >>> 
> ```

Using your exact same call (passing kwargs={"rounding": 2}), it still works for me:
```
>>> lines, substations  = build_transmission(method="line2sub", kwargs={"rounding": 2})
filter substations based on lines
---------------------------------
initial number of substations: 63965
substations with same location after rounding: 13769

initial number of lines: 71554
zero distance lines after rounding: 11589

there are 1117 lines with two substations missing
there are 1848 lines with one substation missing
there are 2042 unique orphan endpoint
there are 7622 substations unconnected

finding closest neighbor to unconnected lines
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2965/2965 [02:52<00:00, 17.22it/s]
2661 line voltages can't be found via neighbor consensus
238 line voltages can't be found via neighbor minimum
>>> 
```

danielolsen:
@rouille, there must be something funny about my environment, when I spin up a docker container and `pipenv sync` everything runs properly.

danielolsen:
Oh this is strange: Using Python 3.9.6 on Windows and Pandas 1.2.5, 1.3.1, or 1.3.2 (latest), I get the error shown. If I downgrade to Pandas 1.1.5 (as specified in the Pipfile), it works properly. This seems potentially related to Pandas Issue#32372.

If I change l.167 from `line2sub.loc[i, f"OTHER_{endpoint}_SUB"] = list(` to `line2sub.at[i, f"OTHER_{endpoint}_SUB"] = list(`, and l.170 from `line2sub.loc[i, f"{endpoint}_SUB"] = all2one[key]` to `line2sub.at[i, f"{endpoint}_SUB"] = all2one[key]`, then the example runs without error on both Pandas 1.1.5 and 1.3.2, so it seems like that change could help future-proof our code. `.at` seems to be preferred over `.loc` when you know you're setting just a single value.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,162 @@ def check_for_location_conflicts(substations):
...
+
+    # pick unique substation at each end. Order of preference:
+    # SUBSTATION --> TAP --> RISER --> DEAD END
+    all2one = {
+        tuple(i): None
+        for j in ["FROM_SUB", "TO_SUB"]
+        for i in line2sub[j]
+        if len(i) > 1
+    }
+    for all in all2one:
```
danielolsen, line 148: Should we rename this to avoid overwriting the python built-in `all()`? It may not make a difference right now, but could interfere with this uses of `all()` within this functions namespace in a potential future refactor.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Do you think it would be clearer if we did something like:
```python
line2sub[f"{endpoint}_SUB"] = line2sub[f"{endpoint}"].apply(
    lambda x: list(subcoord2subid[x]) if x in subcoord2subid else pd.NA
)
```
?

danielolsen: Also I guess we can simplify `f"{endpoint}"` to just `endpoint`.

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: good point

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: When I use your implementation I have got the following error later (l. 135)
`*{tuple(i) for j in ["FROM_SUB", "TO_SUB"] for i in line2sub[j]}`
TypeError: 'NAType' object is not iterable

BainanXia: I think we need to change `pd.NA` into something iterable, like `(pd.NA,)` or `[pd.NA]` or `[]`.

rouille: The funny thing is that in the previous implementation the NaN automatically generated by `map` were `float` and it was fine

rouille: I think the difference is:
```
>>> import pandas as pd
>>> type(pd.NA)
<class 'pandas._libs.missing.NAType'>
```
and
```
>>> import numpy as np
>>> type(np.NAN)
<class 'float'>
```

danielolsen: I don't know why the downstream behavior is different, since conceptually I think we are doing the same thing:
1. Loop through every entry in `line2sub[f"{endpoint}"]`.
2. If the entry is present in `subcoord2subid`, return the result wrapped in a `list`.
3. If not, return `float("nan")` or `pd.NA`.

danielolsen: You can't `tuple` `float("nan")` either:
```python
>>> tuple(float("nan"))
TypeError: 'float' object is not iterable
```

BainanXia: I think @rouille 's post tells the issue. @danielolsen Yeah, we can convert that into tuple, but we can construct a tuple with it: 
image

rouille: This works:
```
line2sub[f"{endpoint}_SUB"] = line2sub[endpoint].apply(
            lambda x: list(subcoord2subid[x]) if x in subcoord2subid else np.NaN
```

danielolsen: It's as simple as `np.NaN` vs `pd.NA`? That's spooky.

rouille: I got the same issue with the `pandas.Int64Index` vs `list`

rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

rouille:
> Oh this is strange: Using Python 3.9.6 on Windows and Pandas 1.2.5, 1.3.1, or 1.3.2 (latest), I get the error shown. If I downgrade to Pandas 1.1.5 (as specified in the Pipfile), it works properly. This seems potentially related to Pandas Issue#32372.
> 
> If I change l.167 from `line2sub.loc[i, f"OTHER_{endpoint}_SUB"] = list(` to `line2sub.at[i, f"OTHER_{endpoint}_SUB"] = list(`, and l.170 from `line2sub.loc[i, f"{endpoint}_SUB"] = all2one[key]` to `line2sub.at[i, f"{endpoint}_SUB"] = all2one[key]`, then the example runs without error on both Pandas 1.1.5 and 1.3.2, so it seems like that change could help future-proof our code. `.at` seems to be preferred over `.loc` when you know you're setting just a single value.

Done

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +30,161 @@ def check_for_location_conflicts(substations):
...
+                        line2sub.loc[i, endpoint]
+                    ]
+                else:
+                    subid = subcoord2subid[
+                        subcoord[
+                            find_closest_neighbor(line2sub.loc[i, endpoint], subcoord)
+                        ]
+                    ]
+                    line2sub.loc[i, f"{endpoint}_SUB"] = list(subid)
+                    endpoint2neighbor[line2sub.loc[i, endpoint]] = list(subid)
```
danielolsen, line 128: I took a stab at reimplementing this in a way that seemed simpler/clearer to me, since it seems we can in order:
1. Find substation(s) closest to all endpoints with no substation listed (`endpoint2neighbor`)
2. Look up the closest substations for each missing endpoint (new `filled_subs` variable)
3. Update the `line2sub` dataframe with the information from (2)
```python
end_sub = {"FROM": "FROM_SUB", "TO": "TO_SUB"}
missing_points = set().union(
    *[set(line2sub[e].loc[line2sub[e_sub].isna()]) for e, e_sub in end_sub.items()]
)
endpoint2neighbor = {
    point: subcoord2subid[subcoord[find_closest_neighbor(point, subcoord)]]
    for point in tqdm(missing_points, total=len(missing_points))
}
filled_subs = [
    line2sub[e].loc[line2sub[e_sub].isna()].map(endpoint2neighbor).map(list)
    for e, e_sub in end_sub.items()
]
line2sub.update(pd.concat(filled_subs, axis=1).rename(columns=end_sub))
```

It saves a few lines, but not as many as I had originally hoped.

rouille: Works for me. We have the same kind of performance (2:40s for rounding=2)

rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +30,157 @@ def check_for_location_conflicts(substations):
...
+    line2sub.update(pd.concat(filled_subs, axis=1).rename(columns=end_sub))
+
+    # select all substations
+    new_substations = substations.copy()
+    new_substations = substations.loc[
+        set(
+            itertools.chain(
+                *{tuple(i) for j in ["FROM_SUB", "TO_SUB"] for i in line2sub[j]}
+            )
+        )
```
danielolsen, line 133: Is this equivalent to `set(pd.concat([line2sub[e] for e in ["FROM_SUB", "TO_SUB"]]).explode())`?

danielolsen (approved):
I've gone through the code and the logic makes sense. Running the two methods with default parameters yields 71,554 lines for the new method, compared to 67,389 lines for the old method. If we trust the line coordinates more than the substation names, this method seems to avoid tossing lines unnecessarily, while only making fairly reasonable assumptions.

A test or two would be nice, but since this new functionality is not being activated by default yet, I don't think that should hold up enabling us to play around with the new method.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +30,157 @@ def check_for_location_conflicts(substations):
...
+    ]
+
+    # pick unique substation at each end. Order of preference:
+    # SUBSTATION --> TAP --> RISER --> DEAD END
+    all2one = {
+        tuple(i): None
+        for j in ["FROM_SUB", "TO_SUB"]
+        for i in line2sub[j]
+        if len(i) > 1
+    }
```
danielolsen, line 143: Is there a reason that this is a dict instead of a set or other iterable?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +30,157 @@ def check_for_location_conflicts(substations):
...
+    for a in all2one:
+        type2id = substations.loc[list(a)].groupby("TYPE").groups
+        for t in ["SUBSTATION", "TAP", "RISER", "DEAD END"]:
+            try:
+                # pick first substations if multiple of same TYPE are found
+                all2one[a] = [
+                    type2id[t][0].item() if len(type2id[t]) > 1 else type2id[t].item()
```
danielolsen, line 150: Would the `GroupBy.first()` method simplify things here?

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: No specific reason. I decided to use a dictionary to relate the multiple substations at the same location to a unique one. If you think it would be cleaner to have two iterables instead, I can refactor the code.

danielolsen: It doesn't really matter; starting as a dict with dummy values and then filling them in vs. starting as a set and then iterating over that to define a dict should be basically equivalent. I just wanted to make sure I wasn't missing something.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,153 @@ def check_for_location_conflicts(substations):
         )
 
 
+def map_lines_to_substations_using_coords(substations, lines, **kwargs):
+    """Map lines to substations using coordinates
+
+    :param pandas.DataFrame substations: data frame of substations.
+    :param pandas.DataFrame lines: data frame of all lines.
+    :param int \\*\\*kwargs: optional arguments.
```
BainanXia, line 37: Just to confirm, according to the discussion below, we are going to use a user facing dictionary as a collections of all `kwargs`, then here we just look for `rounding` right? Then I presume this is a `dict` instead of `int`, no?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,153 @@ def check_for_location_conflicts(substations):
...
+    :return: (*pandas.DataFrame*) -- indices: lines id, columns: lines endpoint
+        coordinates (*FROM* and *TO*) and id of substations found at these coordinates (*FROM_SUB* and *TO_SUB*)
```
BainanXia, line 39: Should we cut this long line off?

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: not sure how the `int` sneaked in

danielolsen: Probably originally `:param int rounding:`

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: I can but it still fits within the 89 character length (enforce by `black` for the code but not the docstring)

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Used `:param **kwargs:`

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: I shortened the docstrings

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,152 @@ def check_for_location_conflicts(substations):
...
+
+    lines_coord_from = lines["COORDINATES"].map(
+        lambda x: (round(x[0][0], rounding), round(x[0][1], rounding))
+    )
+    lines_coord_to = lines["COORDINATES"].map(
+        lambda x: (round(x[-1][0], rounding), round(x[-1][1], rounding))
+    )
+    line2sub = lines_coord_from.to_frame(name="FROM").join(
+        lines_coord_to.to_frame(name="TO")
+    )
```
BainanXia, line 60: Maybe we could do following here:
```python
lines_coord = lines.COORDINATES.map(lambda x: list(np.round([x[0],x[-1]], rounding)))
line2sub = pd.DataFrame(line_coord.to_list(), columns=["FROM", "TO"], index=line_coord.index)
```

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,152 @@ def check_for_location_conflicts(substations):
...
+
+    for endpoint in ["FROM", "TO"]:
+        line2sub[f"{endpoint}_SUB"] = line2sub[endpoint].apply(
+            lambda x: list(subcoord2subid[x]) if x in subcoord2subid else np.NaN
```
BainanXia, line 64: If converting to `list` is not essential, we could do `.apply(lambda x: subcoord2subid.get(x, np.NaN)` here.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,152 @@ def check_for_location_conflicts(substations):
...
+        f"there are {n_unique_orphan_endpoint} unique orphan endpoint\n"
+        f"there are {n_unconnected_sub} substations unconnected\n"
+    )
+
+    return line2sub
+
+
+def assign_substations_to_lines(substations, lines, line2sub):
+    """Locate closest substation of unconnected lines and update information
+    information of lines. Note that the ``line2sub`` data frame is modified in-place.
```
BainanXia, line 97: `information` is duplicated here.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,152 @@ def check_for_location_conflicts(substations):
...
+
+    :param pandas.DataFrame substations: data frame of substations.
+    :param pandas.DataFrame lines: data frame of all lines.
+    :param pandas.DataFrame line2sub: indices: table mapping lines to substations
+        as returned by :func:`map_lines_to_substations_using_coords`.
```
BainanXia, line 102: what does `indices` mean here?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,152 @@ def check_for_location_conflicts(substations):
...
+    :return: (*pandas.DataFrame*) -- original lines data frame with updated
+        *SUB_1* and *SUB_2* columns, new *SUB_1_ID*, *SUB_2_ID* along with
+        *OTHER_SUB_1_ID* and *OTHER_SUB_2_ID* when multiple location are
+        associated to the same endpoint.
+    """
+    print("finding closest neighbor to unconnected lines")
+
+    # find closest neighbor of unconnected lines
+    subcoord2subid = substations.groupby(["LONGITUDE", "LATITUDE"]).groups
```
BainanXia, line 111: Should we `round` here as well?

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: The problem is that the above gives me the lat/lon in `line2sub` as `list`. And then, I have the following error:
```
  File "/Users/brdo/CEM/PreREISE/prereise/gather/griddata/hifld/data_process/transmission.py", line 70, in <lambda>
    lambda x: list(subcoord2subid[x]) if x in subcoord2subid else np.NaN
TypeError: unhashable type: 'numpy.ndarray'
```

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: it should not be there

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: I would say no. We can only gain at using the more precise located substations to find the closest neighbor.

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: Yeah, we need tuple here. Googling a bit and I found a way to convert all elements into `tuple` using: `line2sub.applymap(tuple)`

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: But this will give us a different group of substations which is not consistent with `map_lines_to_substations_using_coords` right? Would it be a problem that we use one grouping of substations when dealing with some lines and use a different grouping of substations when dealing with the rest of the lines?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -27,6 +29,152 @@ def check_for_location_conflicts(substations):
...
+    line2sub["OTHER_FROM_SUB"] = None
+    line2sub["OTHER_TO_SUB"] = None
+    for i in line2sub.index:
+        for endpoint in ["FROM", "TO"]:
+            if len(line2sub.loc[i, f"{endpoint}_SUB"]) > 1:
+                key = tuple(line2sub.loc[i, f"{endpoint}_SUB"])
+                line2sub.at[i, f"OTHER_{endpoint}_SUB"] = list(
+                    set(key) - set(all2one[key])
+                )
+                line2sub.at[i, f"{endpoint}_SUB"] = all2one[key]
```
BainanXia, line 160: Not sure whether this is all we need here:
```python
type_rank = {"SUBSTATION": 1, "TAP": 2, "RISER": 3, "DEAD END": 4}
for c in ["FROM_SUB", "TO_SUB"]:
    line2sub.loc[:, c] = line2sub[c].apply(lambda x: sorted(x, key=lambda y: type_rank[substations.loc[y, "TYPE"]]))

```

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: I don't think it matters. In `map_lines_to_substations_using_coords`, we identify the substations that have the exact same coordinates than the endpoints of the lines for a given rounding. In `assign_substations_to_lines` we assign substations to endpoints that have not been matched. If we don't round we will likely have a unique closest neighbor, if we round we will likely have multiple closest neighbor (with same location) and the algorithm will pick the first one it encounters in the list (see `powersimdata.utility.distance.find_closest_neighbor`). What do you think is better?

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: I like the idea. I would still need to save the other substations to populate the `OTHER_FROM_SUB` and `OTHER_TO_SUB` columns. I will see how I can do that with your implementation

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: Done

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: I see your point. I was thinking the following situation:
Suppose we have raw sub list consists of: 
`coord1`, `sub_a`, "Substation"
`coord2`, `sub_b`, "Tap"
`coord3` `sub_c`,  "Deadend"
and line list consists of: 
`coord12`, `line_a`, 
`coord4`, `line_b`. 
After `groupby` with rounding, we have
`coord12`, [`sub_a`, `sub_b`], ["Substation", "Tap"]
`coord3`, `sub_c`, "Deadend"
Then we match `line_a` with `sub_a` via exact match of `coord12` and no match for `line_b`
Next we find nearest neighbor of `line_b` in the original sub list, and get `sub_b` with `coord2` (suppose dist(`coord4`, `coord2`) is the minimum). Then we will assign `sub_b` to `line_b` instead of `sub_a` in the other case. Not sure whether it matters and I don't have preferences between these two. Just bring this up for discussion.

rouille: I guess I don't see the point of rounding and potentially having multiple closes neighbor if we don't have a selection criteria (voltage, number of lines connected to the substations, etc) in place. Otherwise, we take the closest one based on an surrealistic precision (what is implemented) or the one that appears first in the list as it is done in the closest neighbor algorithm (what you propose).

BainanXia: My point in the above example is all matched lines will be assigned to `sub_a` if we follow consistent criteria in both cases (location matched directly + missing ones matched by nearest neighbor), however if we follow the current implementation, i.e. find nearest neighbor without rounding, we will have `sub_b` assigned to some of the lines. 

However, `sub_b` is unlikely to be an active candidate in the final substation list, given it is grouped together with `sub_a` which is a "SUBSTATION". Also I'm not sure at the end of day whether we will come up with universal substation IDs to replace the original ones, if so, are we going to assign different substation IDs to `OTHER_{endpoint}_SUB` as well?

rouille: I understand your the difference between the 2 implementations. That said, there is no guarantee that if we round the coordinates leading to `sub_a = sub_b` that `sub_a` will be picked up. It will depend of their respective position in the substation data frame that is traversed row-wise to find the closest neighbor.

You are right that in the current implementation a line that is matched to a neighbor substation will have `OTHER_{endpoint}_SUB` set to `None`

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -114,8 +115,9 @@ def assign_substations_to_lines(substations, lines, line2sub):
             for e, e_sub in end_sub.items()
         ]
     )
+    tree = KDTree([ll2uv(p[0], p[1]) for p in subcoord])
```
BainanXia, line 118: Shall we refactor `find_closest_neighbor` in this way so that any function calls `find_closest_neighbor` will benefit from the performance improvement?

danielolsen: I think the performance improvement comes when we have many points for which we want to find the closest neighbor in another set of points. If we build the tree each time, we may not get any performance improvement.

rouille: yes that is the problem. In fact I first refactored `find_closest_neighbor` to use `KDTree` but it was very slow because of the reason mentioned by @danielolsen, i.e., the tree was built for each iteration

danielolsen: It could be nice to have a new function `find_closest_neighbors`, to replace any repetitive calls to `find_closest_neighbor` or more manual `haversine` loops.

rouille: In fact we could write a new function that takes a list of coordinates that will be associated to a closest neighbor in a second list of points. It would return a list of id in the second list (neighbors).

danielolsen: It would be a thin wrapper around converting lat/lon to unit vector via `ll2uv`, building a `KDTree`, and querying it. Simple but handy.

BainanXia: Agree.

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: @danielolsen, what do you think? For lines that have no substations at their endpoint after `map_lines_to_substations_using_coords`, should we:
* Find closest neighbor in the original substation data frame (current implementation) or
* Find the closest location in the rounded substation data frame and pick one of the substation. `KDTree` returns a list of closest neighbors if they have same location.

danielolsen: If we're saying that substations with coordinates that are close enough (i.e. the same after rounding) will be considered as one substation in the output, then I don't think it makes sense to differentiate between them when assigning lines to them. Whether we assign lines based on rounded coordinates (somehow picking the ID of the one substation that will remain), or assign lines based on the unrounded coordinates and then reassign any lines mapped to 'other' subs back to the main sub, the results should be equivalent, right?

BainanXia: > assign lines based on the unrounded coordinates and then reassign any lines mapped to 'other' subs back to the main sub

@rouille In the current implementation, we don't reassign lines mapped to "other" subs back to the main sub on this track, do we?

rouille: Not sure I understand. let me quickly summarize what the code is doing. We have two functions:
* `map_lines_to_substations_using_coords` that does the following
    * Round coordinates of substations and lines' endpoints
    * Map coordinates of lines' endpoints to coordinates of substations
    * Return a data frame where each line (index) can have at its endpoints a substation, a list of substation a Nan (no match):
```
>>> from prereise.gather.griddata.hifld.data_access.load import get_hifld_electric_power_transmission_lines, get_hifld_electric_substations
>>> from prereise.gather.griddata.hifld.data_process.transmission import map_lines_to_substations_using_coords
>>> sub = get_hifld_electric_substations("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Substations_Jul2020.csv")
>>> line = get_hifld_electric_power_transmission_lines("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Power_Transmission_Lines_Jul2020.geojson.zip")
>>> 
>>> line2sub = map_lines_to_substations_using_coords(sub, line, rounding = 3)
initial number of substations: 70857
substations with same location after rounding: 4464

initial number of lines: 71554
zero distance lines after rounding: 3314

there are 1172 lines with two substations missing
there are 2386 lines with one substation missing
there are 2584 unique orphan endpoint
there are 15946 substations unconnected
>>> line2sub.head(n=25)
                  FROM                  TO        FROM_SUB          TO_SUB
0   (48.188, -122.113)  (48.215, -122.028)  [55014]         [64253]       
1   (47.924, -122.104)  (47.925, -122.101)  [55018]         [56524]       
2   (47.838, -117.352)  (47.838, -117.41)   [55247]         [64264]       
3   (47.647, -121.907)  (47.648, -121.907)  [55295]         [64254]       
4   (46.873, -122.695)  (46.882, -122.7)    [55308]         [64255]       
5   (47.842, -122.3)    (47.842, -122.297)  [55359]         [64256]       
6   (47.799, -122.351)  (47.8, -122.351)    [55363]         [64257]       
7   (47.282, -119.555)  (47.282, -119.555)  [55376, 64258]  [55376, 64258]
8   (47.674, -122.158)  (47.675, -122.158)  [55436]         [64259, 64260]
9   (47.674, -122.158)  (47.675, -122.158)  [55436]         [64259, 64260]
10  (47.668, -122.196)  (47.668, -122.196)  [55437, 64261]  [55437, 64261]
11  (46.974, -122.827)  (46.976, -122.825)  [64262]         [55460]       
12  (46.977, -122.824)  (46.976, -122.825)  [64263]         [55460]       
13  (44.072, -122.967)  (44.071, -122.972)  [56648]         [56649]       
14  (45.628, -122.782)  (45.628, -122.786)  [56684]         [64265]       
15  (45.611, -122.773)  (45.614, -122.769)  [56685]         [61413, 61414]
16  (45.485, -122.749)  (45.486, -122.748)  [56717]         [60296]       
17  (44.292, -121.503)  (44.27, -121.493)   [56733]         [63254]       
18  (45.067, -117.864)  (45.05, -117.809)   [56736]         [64266]       
19  (45.904, -119.32)   (45.909, -119.323)  [64267]         [64314, 64315]
20  (45.904, -119.32)   (45.909, -119.323)  [64267]         [64314, 64315]
21  (45.917, -119.307)  (45.865, -119.3)    NaN             [64268]       
22  (45.699, -121.522)  (45.699, -121.522)  [56848, 60264]  [56848, 60264]
23  (45.699, -121.522)  (45.699, -121.521)  [56848, 60264]  [64269]       
24  (44.297, -120.886)  (44.298, -120.884)  [56944]         [64270]       
``` 
* `assign_substations_to_lines` that does:
    * Find closest substation for each line having one or two NaN value(s). CURRENTLY (this is what is discussed here) we look into the original substation data frame where each entry as a unique location. Only one substation is associated to the missing endpoint.
    * For line having multiple substation at their endpoint (e.g. line `7`), we keep the substation that has a `TYPE` in the following order of preference: SUBSTATION --> TAP --> RISER --> DEAD END. The other substations are saved and will are added  in the `OTHER_SUB_1_ID` and `OTHER_SUB_2_ID` of the returned lines data frame (see below)
    * A line data frame is returned with additional `SUB_1_ID`, `SUB_2_ID`, `OTHER_SUB_1_ID` and `OTHER_SUB_2_ID`. The `SUB_1` and `SUB_2` columns (name of substations) is overwritten with the substations' name that has been matched using the `SUB_1_ID`, `SUB_2_ID` columns. This data frame has the same number of rows than the original one.
    * We return a substation data frame with entries matching what can be found in `SUB_1_ID`, `SUB_2_ID`. Potentially, we could include substations in `OTHER_SUB_1_ID` and `OTHER_SUB_2_ID`.

danielolsen: Thanks, this is clearer. There is a lot going on in the `assign_substations_to_lines` function and it was a little hard to keep straight in my head. If I understand properly, then for the returns of this function:
- within the lines data frame, each line maps to exactly one substation (either the one it was already mapped to, or a newly-mapped one), and if there were other substations that the line could have been mapped to, those are listed in the OTHER columns.
- the substation data frame only contains the substations that at least one line is mapped to.

Yeah?

rouille: Correct.

BainanXia: @rouille Very clear and it is consistent with my understanding. The line data frame is clean. Each line has two substations attached to both endpoints. I think the question is more for substation data frame, do we want any rows to be entries listed in `OTHER_SUB` columns in the line data frame, i.e. should there be lines connect to substations listed in `OTHER_SUB` columns?

rouille: It would not hurt.

BainanXia: Previously, I thought for the list of substations (`groupby` with rounding), we pick one according to the ranking as the unique substation entry and treat the rest ones in the list as functional affiliations. Hence, all the lines should have such unique entries as endpoints rather than affiliations. If that's not the case, I believe it's still okay. @danielolsen What do you think?

danielolsen: I think I see what @BainanXia is getting at: if a line _already_ was mapped to a substation ID, and then that substation was 'collapsed' into a group of co-located substations, and the original substation was not chosen as the 'primary' substation, then I think we would want to re-map that endpoint from the 'secondary' substation to the 'primary' one. That way, every 'primary' substation ID maps to an entry in the substation data frame.

rouille: I understand now. I will implement that

rouille: This has been implemented in 71f13d9. This is what is happening now
* Call `map_lines_to_substations_using_coords`:
```
>>> from prereise.gather.griddata.hifld.data_access.load import get_hifld_electric_power_transmission_lines, get_hifld_electric_substations
>>> from prereise.gather.griddata.hifld.data_process.transmission import map_lines_to_substations_using_coords, build_transmission
>>> 
>>> sub = get_hifld_electric_substations("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Substations_Jul2020.csv")
>>> line = get_hifld_electric_power_transmission_lines("https://besciences.blob.core.windows.net/datasets/hifld/Electric_Power_Transmission_Lines_Jul2020.geojson.zip")
>>> 
>>> line2sub = map_lines_to_substations_using_coords(sub, line, rounding = 3)
```
* Looking at line `87795` in data frame mapping lines to substations after rounding:
```
>>> line2sub.loc[87795]
FROM        (43.805, -95.175)
TO          (43.805, -95.175)
FROM_SUB       [42737, 52083]
TO_SUB                [42737]
Name: 87795, dtype: object
```
Since `FROM` = `TO`, then `FROM_SUB` = `TO_SUB` unless the location of substation `42737` matches exactly the location of the `TO` end of line `87795`.
* Looking at substation `42737`, we have:
```
>>> sub.loc[42737, "LATITUDE"]
43.805099
>>> sub.loc[42737, "LONGITUDE"]
-95.174541
```
* Looking at the coordinates of line `87795`:
```
>>> line.loc[87795, "COORDINATES"]
[[43.804914316000065, -95.17504414599995], [43.80509324500008, -95.17512502199997], [43.805211787000076, -95.17460879499998], [43.805099, -95.174541]]
```
The `FROM` coordinates (first list) is slightly different than the `TO` coordinates (last list) and the latter matches exactly the coordinates of substation `42737`

I think this is what we want.

danielolsen: Playing around with this:
```python
>>> lines, substations  = build_transmission(method="line2sub", kwargs={"rounding": 2})
[output truncated for brevity]
>>> len(lines)
71554
>>> len(substations)
54487
>>> lines["SUB_1_ID"].isin(substations.index).all()
True
>>> lines["SUB_2_ID"].isin(substations.index).all()
True
>>> primary_substations = set(lines["SUB_1_ID"].unique()) | set(lines["SUB_2_ID"].unique())
>>> len(primary_substations)
43285
>>> len(substations.loc[primary_substations].groupby(["LATITUDE", "LONGITUDE"]))
43285
>>> len(substations.loc[primary_substations, ["LATITUDE", "LONGITUDE"]].round(2).groupby(["LATITUDE", "LONGITUDE"]))
42671
```
It seems that there are no 'primary' substations with the same lat/lon, but there about 600 that overlap within the rounding precision at rounding=2, or about 180 when we round to 3 decimal places (50708 total, 50529 after grouping).

BainanXia: If I'm understanding correctly, here is what's going on:

1. Map lines to substations after rounding, some endpoints of lines are assigned a list of substations since they are grouped due to rounding.
2. For those endpoints above, look at the coordinates before rounding and check if there is an exact match in the list, if so, drop the rest elements in the list, if not, pick the primary substation according to the ranking then drop the rest elements in the list (later).
3. For the rest endpoints with no match in 1, assign the closest neighbor to it.
4. Recreate the list of substations based on the resulting line data frame, any substations matched to a line (with rounding) are kept and the rest ones are dropped.

Then I'm wondering whether we could simplify the process as following:

1. Map endpoints of lines to substations based on exact match of coordinates .
2. Map no-exact-match endpoints of lines to substations with rounding.
3. Map remaining endpoints of lines to nearest substations.

Further thoughts regarding new substation data frame:

1. Shall we keep row entries to be primary substations, i.e. substations with lines directly connecting to them. In this way` len(primary_substations)` == `len(substations)`
2. Move `OTHER_SUB` columns from line data frame to substation data frame, since this might be cleaner without duplicated entries (currently, if multiple endpoints are mapped to a same group of substations with rounding, they will have duplicated value in the `OTHER_SUB` columns.

All draft thoughts. Open to discussions.

rouille: @BainanXia, I believe that the first implementation and the second will lead to the same outcome. 

Regarding the difference in length between the output `substations` and @danielolsen's `primary_substations`, it comes from the fact that we include substations listed in `OTHER_FROM_SUB` and `OTHER_TO_SUB` into the output substations. Would you prefer to get to not include them?

@BainanXia, I don't understand your second point. How will we connect the line to substations listed in the `OTHER_FROM_SUB` and `OTHER_TO_SUB` columns if these are moved into the substations data frame?

BainanXia: @rouille 
- Agree. The two implementations should lead to the same outcome. I bring it up just for simpler logic flow.
- Yes. Following our current setup in the `usa_tamu` model, all substations (row entries) are primary substations. I lean to not including them as separate row entries in the data frame.
- I was thinking instead of `OTHER_FROM_SUB` and `OTHER_TO_SUB` in line data frame, we just maintain an `OTHER_SUB` column in substation data frame.

danielolsen: @BainanXia I think there is still value in reporting the 'secondary' substation rows. E.g. if we want to augment our network with underground lines that we don't think are present in the HIFLD data set, it is helpful to know which 'primary' substations have a secondary `RISER` substation.

BainanXia: @danielolsen Agree. What I'm proposing is to keep them in the substation data frame instead of line data frame. What do you think?

danielolsen: Ah, right, I had meant to reply to that point as well and agree with it. Secondary substation IDs seem more related to substations themselves than lines which connect to them.

PR #218, opened 2021-09-08
feat: add function to create transformers within substations

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Closes #206. ~This is a draft PR right now, since it relies on code from #210. It could also become part of #210.~

### What the code is doing
Using buses, grouped by substation, for each substation with more than one bus, a transformer is added between the highest-voltage bus in the substation and each other bus. This is an arbitrary topology that seems reasonable enough, although it is by no means the best one.

### Testing
A new unit test is added which covers the bases of: no transformers within a substation, one transformer within a substation, multiple transformers within a substation.

### Usage Example
```python
>>> from prereise.gather.griddata.hifld.data_process.transmission import (
...     build_transmission,
...     create_buses,
...     create_transformers,
... )
>>> lines, substations = build_transmission()
dropping 6892 substations of 70857 total due to LINES parameter equal to 0
dropping 759 lines with one or more substations listed as 'NOT AVAILABLE' out of a starting total of 71554
dropping 3143 lines with one or more substations not found in substations table out of a starting total of 70795
Evaluating endpoint location mismatches... (this may take several minutes)
dropping 120 lines with one or more substations with non-matching coordinates out of a starting total of 67652
dropping 143 lines with matching SUB_1 and SUB_2 out of a starting total of 67532
2119 line voltages can't be found via neighbor consensus
310 line voltages can't be found via neighbor minimum
>>> bus, bus2sub = create_buses(lines)
>>> transformers = create_transformers(bus, bus2sub)
>>> transformers
      from_bus_id  to_bus_id
0               0          1
1               2          3
2               5          6
3              12         13
4              20         21
...           ...        ...
5394        58235      58236
5395        58239      58240
5396        58246      58247
5397        58258      58259
5398        58262      58263

[5399 rows x 2 columns]
```

### Time estimate
15 minutes.

PR #219, opened 2021-09-08
feat: add functions to estimate branch parameters (impedance and rating)

danielolsen:

### Purpose
Add functions to estimate the impedance and rating of each branch (transformers and lines). This PR depends on #218.

### What the code is doing
In **const.py**, we add assumed line and transformer parameters. @YifanLi86, do we have documentation of where these values come from? Are the assumptions you provided last December still good?

In **transmission.py**:
- We add a new function `get_mst_edges`, which either generates these edges from scratch and then caches them locally (the pickle file is about 35 KB), or loads them from the local cache, using a subset of the lines and substations dataframes as a crude hash key in an attempt to only load the solved edges for the same starting network.
- Within `build_transmission`, we call the `get_mst_edges` function, and then expand the information returned into a data frame with information required to estimate line impedances and ratings.
- We add new functions `calculate_branch_mileage` (lines only), estimate_branch_impedance (lines and transformers), and estimate_branch_rating (lines and transformers).
- Within `check_for_location_conflicts`, we add a bit more printing, so that it's clear to the user than all voltages are assigned by the heuristics, now that the minimum spanning tree edges are added.

### Testing
Tested manually.

### Usage Example/Visuals
```python
>>> import pandas as pd
...     build_transmission,
...     calculate_branch_mileage,
...     create_buses,
...     create_transformers,
...     estimate_branch_impedance,
...     estimate_branch_rating,
... )
>>> # build_transmission has been augmented to include minimum spanning tree
Reading cached minimum spanning tree
2140 line voltages can't be found via neighbor consensus
No more missing voltages remain after neighbor minimum
>>> bus, bus2sub = create_buses(lines)
>>> transformers = create_transformers(bus, bus2sub)
>>> # New usage example code starts here
>>> lines["type"] = "Line"
>>> lines["length"] = lines.apply(calculate_branch_mileage, axis=1)
>>> transformers["type"] = "Transformer"
>>> branch = pd.concat([lines, transformers])
>>> branch["x"] = branch.apply(lambda x: estimate_branch_impedance(x, bus), axis=1)
>>> branch["rateA"] = branch.apply(estimate_branch_rating, axis=1)
```

### Time estimate
1 hour.


YifanLi86:
@danielolsen For reference, I think we can mentioned these are industry generalized/standard values commonly used in studies like MTEP and RIIA (not exactly the same, have been averaged out so not CEII). https://www.misoenergy.org/about/media-center/miso-report-concludes-much-higher-levels-of-renewables-integration-are-achievable/; https://www.misoenergy.org/planning/planning/. They are developed based on various sources like (https://www.hitachiabb-powergrids.com/us/en/offering/product-and-system/high-voltage-switchgear-and-breakers/gas-insulated-transmission-line) with some engineering judgement and general knowledge.

danielolsen:
@YifanLi86 could you please come up with a couple of sentences for each of these sets of values? Something that can be included in the repo as documentation. We probably don't need to go through the full data intake procedure for these simple tables, but the [data intake template](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/.github/ISSUE_TEMPLATE/data-intake.md) should give a good idea of the sort of information we usually add.

YifanLi86:
> @YifanLi86 could you please come up with a couple of sentences for each of these sets of values? Something that can be included in the repo as documentation. We probably don't need to go through the full data intake procedure for these simple tables, but the [data intake template](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/.github/ISSUE_TEMPLATE/data-intake.md) should give a good idea of the sort of information we usually add.

Check this out to see if this is helpful..
[Data Intake.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7130977/Data.Intake.docx)

danielolsen:
> 
> 
> > @YifanLi86 could you please come up with a couple of sentences for each of these sets of values? Something that can be included in the repo as documentation. We probably don't need to go through the full data intake procedure for these simple tables, but the [data intake template](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/.github/ISSUE_TEMPLATE/data-intake.md) should give a good idea of the sort of information we usually add.
> 
> Check this out to see if this is helpful..
> [Data Intake.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7130977/Data.Intake.docx)

Could you be more specific with which sources are used for which tables?

YifanLi86:
> > > @YifanLi86 could you please come up with a couple of sentences for each of these sets of values? Something that can be included in the repo as documentation. We probably don't need to go through the full data intake procedure for these simple tables, but the [data intake template](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/.github/ISSUE_TEMPLATE/data-intake.md) should give a good idea of the sort of information we usually add.
> > 
> > 
> > Check this out to see if this is helpful..
> > [Data Intake.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7130977/Data.Intake.docx)
> 
> Could you be more specific with which sources are used for which tables?

I splited it into rating and reactance. Not sure if this is adequate, as there are documents that useful for both. Let me know how this looks.

[Data Intake.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7131064/Data.Intake.docx)

danielolsen:
This is more helpful, but there's still a lot of ambiguity. Could you please break it down by which specific sources were used in developing:
- Line reactance per mile for lines > 50 miles
- Line ratings for lines <= 50 miles
- Equation structure/values for line ratings as a function of surge impedance loading
- Transformer reactances by voltage
- Transformer ratings

For the links that are general information on MISO's transmission planning, or for the Eastern Interconnection Reliability Assessment Group, could you point to particular document(s)? In general, could you point to where in each document (e.g. table name/number) information comes from?

The static rating for all transformers seems potentially problematic, especially when it's applied to some of the higher voltage transformers for which I would assume higher ratings are often appropriate.

YifanLi86:
@danielolsen I agree, these data sources are pretty ambiguity. A lot of things are actually just general industry guidelines, e.g., I know 345kV MVA can vary a lot by conductor type, but 1793MVA is just the number everyone is using when TOs developing newprojects without much reference. Same thing for transformer, apprently its rating can vary a lot, however everyone seems just using this value regardless of its kV. I am not fully comfortable with this either, but it might be tricky to track down everything to the exact data source. Let me go over it again and try to get more specific. I will get back a little bit later.

danielolsen:
> @danielolsen I agree, these data sources are pretty ambiguity. A lot of things are actually just general industry guidelines, e.g., I know 345kV MVA can vary a lot by conductor type, but 1793MVA is just the number everyone is using when TOs developing newprojects without much reference. Same thing for transformer, apprently its rating can vary a lot, however everyone seems just using this value regardless of its kV. I am not fully comfortable with this either, but it might be tricky to track down everything to the exact data source. Let me go over it again and try to get more specific information. I will get back a little bit later.

Thanks! I don't think we need to track down the _original_ source of a given number, but it would be helpful if we can point to at least one specific external source for each of the numbers we're adding (or multiple sources if we've picked a representative value). Then if we find additional or updated sources, it's more straightforward to be able to update our 'consensus' values.

danielolsen:
If _physical_ transformers are usually modeled in industry as being all 800 MVA, would it be reasonable to equivalently model `n` transformers in parallel when they're connected to lines at 345 kV and above? So the rating would be `800*n` MVA, and the impedance would be `x/n` where `x` is the per-unit impedance for a single transformer that's looked up via the `transformer_reactance` table in **const.py**.

YifanLi86:
> If _physical_ transformers are usually modeled in industry as being all 800 MVA, would it be reasonable to equivalently model `n` transformers in parallel when they're connected to lines at 345 kV and above (since for these lines, a short line would have a rating of >800 MVA)? So the rating would be `800*n` MVA, and the impedance would be `x/n` where `x` is the per-unit impedance for a single transformer that's looked up via the `transformer_reactance` table in **const.py**.

Yes you are right, that is exactly what people would do in modeling. I think this is partly because transformers most of the time are not limited by thermal effect. I will get you more documents soon.

danielolsen:
> > If _physical_ transformers are usually modeled in industry as being all 800 MVA, would it be reasonable to equivalently model `n` transformers in parallel when they're connected to lines at 345 kV and above (since for these lines, a short line would have a rating of >800 MVA)? So the rating would be `800*n` MVA, and the impedance would be `x/n` where `x` is the per-unit impedance for a single transformer that's looked up via the `transformer_reactance` table in **const.py**.
> 
> Yes you are right, that is exactly what people would do in modeling. I think this is partly because transformers most of the time are not limited by thermal effect. I will get you more documents soon.

Great! Do the assumed per-unit reactances already take this sort of parallel arrangement into account? If we keep the default rating as a starting point for all transformers, then for the current values if a transformer has 345 kV as the top end, it'll be modeled as 3 parallel transformers, 500 kV will get 4, and 765 kV will get 7. If we accidentally double-scale the impedance, it could have a pretty large impact for higher voltage transformers.

YifanLi86:
> > > If _physical_ transformers are usually modeled in industry as being all 800 MVA, would it be reasonable to equivalently model `n` transformers in parallel when they're connected to lines at 345 kV and above (since for these lines, a short line would have a rating of >800 MVA)? So the rating would be `800*n` MVA, and the impedance would be `x/n` where `x` is the per-unit impedance for a single transformer that's looked up via the `transformer_reactance` table in **const.py**.
> > 
> > 
> > Yes you are right, that is exactly what people would do in modeling. I think this is partly because transformers most of the time are not limited by thermal effect. I will get you more documents soon.
> 
> Great! Do the assumed per-unit reactances already take this sort of parallel arrangement into account? If we keep the default rating as a starting point for all transformers, then for the current values if a transformer has 345 kV as the top end, it'll be modeled as 3 parallel transformers, 500 kV will get 4, and 765 kV will get 7. If we accidentally double-scale the impedance, it could have a pretty large impact for higher voltage transformers.

Good point...I believe these should be already taking care of the parallel issue. These transformer data are actually averaged out from ERAG cases (the case itself is CEII but average numbers are not).

prereise/gather/griddata/hifld/const.py
```diff
@@ -118,3 +118,63 @@
...
+    (161, 345): 0.03293,
+    (161, 500): 0.06978,
+    (230, 345): 0.02085,
+    (230, 500): 0.01846,
+    (230, 765): 0.01616,
+    (345, 500): 0.01974,
+    (345, 765): 0.01625,
+    (500, 765): 0.00436,
+}
+transformer_rating = 800
```
rouille, line 180: Should we add comments that give units for impedance, reactance, rating?

YifanLi86:
Please see here for all the original data source that points to the values that have been used. One thing I think we need to update is to increase the actual route length factor from 1.2 to 1.3, which seems more commonly used now.
[data source.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7156680/data.source.docx)

prereise/gather/griddata/hifld/const.py
danielolsen: Done, for the tables I'm confident in the units for.

rouille:
General comments. The original line data frame has all its column name in upper case. The new added ones (`type`, `length`, `start` and `end`) are in lower case. I would try to use the same convention across columns (using upper case for the new ones or lowering all the columns of all the data frames in the `load` module)

danielolsen:
> Please see here for all the original data source that points to the values that have been used. One thing I think we need to update is to increase the actual route length factor from 1.2 to 1.3, which seems more commonly used now.
> [data source.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7156680/data.source.docx)

Thanks! I have a couple more questions:
- When you say `Average out based on our synthetic grid data and cut by 75% quantile`, are you referring to the dataset directly from TAMU (as per the link shown), the 'base' values in the **branch.csv** file in our repo/Zenodo, or the `Grid.branch` values of a particular `Scenario`? If we're referring to the TAMU dataset or our **branch.csv** file, which version of the dataset?
- When you say `Compared our data to ERAG and find relatively small mismatch so eventually decided to go with our data`, are you also comparing against the 75% quantile there, or an average? Either way, can those reference values from ERAG be shared?
- Are we taking the 75th percentile value, or cutting off the top quartile and then averaging? What's the significance of the 75th percentile here?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -322,6 +328,105 @@ def create_transformers(bus):
...
+
+
+def calculate_branch_mileage(branch):
+    """Estimate distance of a line.
+
+    :param pandas.Series branch: data for a single line.
+    :return: (*float*) -- distance (in miles) for that line.
+    """
+    coordinates = branch.loc["COORDINATES"]
+    return sum([haversine(a, b) for a, b in zip(coordinates[:-1], coordinates[1:])])
```
rouille, line 371: I think `Haversine` is expected `a` and `b` to (lat, lon) but the elements of `coordinates` are (lon, lat)

YifanLi86:
> > Please see here for all the original data source that points to the values that have been used. One thing I think we need to update is to increase the actual route length factor from 1.2 to 1.3, which seems more commonly used now.
> > [data source.docx](https://github.com/Breakthrough-Energy/PreREISE/files/7156680/data.source.docx)
> 
> Thanks! I have a couple more questions:
> 
> * When you say `Average out based on our synthetic grid data and cut by 75% quantile`, are you referring to the dataset directly from TAMU (as per the link shown), the 'base' values in the **branch.csv** file in our repo/Zenodo, or the `Grid.branch` values of a particular `Scenario`? If we're referring to the TAMU dataset or our **branch.csv** file, which version of the dataset?
> * When you say `Compared our data to ERAG and find relatively small mismatch so eventually decided to go with our data`, are you also comparing against the 75% quantile there, or an average? Either way, can those reference values from ERAG be shared?
> * Are we taking the 75th percentile value, or cutting off the top quartile and then averaging? What's the significance of the 75th percentile here?

1. The branch.csv file;
2. Yes, and probably not to be safe;
3. Here is the method illustration:

image

danielolsen:
> General comments. The original line data frame has all its column name in upper case. The new added ones (`type`, `length`, `start` and `end`) are in lower case. I would try to use the same convention across columns (using upper case for the new ones or lowering all the columns of all the data frames in the `load` module)

There will definitely be more massaging of column names before we return the final CSV to be used to create a Grid. Do you think there is value in maintaining a distinction between columns from the source data (all upper case) and columns that we've calculated/estimated (all lower case)?

prereise/gather/griddata/hifld/const.py
```diff
@@ -118,3 +118,67 @@
...
+    138: 0.0026,
+    161: 0.0021,
+    230: 0.0011,
+    345: 0.0005,
+    500: 0.0002,
+    765: 0.0001,
+}
+line_rating_short = {  # MVA
+    69: 86,
+    110: 181,
```
BainanXia, line 135: Should this be 100 instead of 110 here?

danielolsen:
@YifanLi86 visualizing the distribution of impedances is useful, but I'm still not quite certain what `Average out... and cut by 75% quantile` means as a formula/algorithm. Is it just taking the value at the 75th percentile? Or it is excluding upper-quartile outliers and then calculating an average? If so, is there any significance to 75% as a cutoff? And should we exclude lower-end outliers too?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -322,6 +328,105 @@ def create_transformers(bus):
...
+    os.makedirs(cache_dir, exist_ok=True)
+    cache_key = repr(lines[["SUB_1_ID", "SUB_2_ID"]].to_numpy().tolist()) + repr(
+        substations[["LATITUDE", "LONGITUDE"]].to_numpy().tolist()
+    )
+    cache_hash = hashlib.md5(cache_key.encode("utf-8")).hexdigest()
+    try:
+        with open(os.path.join(cache_dir, f"mst_{cache_hash}.pkl"), "rb") as f:
+            print("Reading cached minimum spanning tree")
+            mst_edges = pickle.load(f)
+    except Exception:
```
rouille, line 422: I guess we can use a `FileNotFoundError`

danielolsen: What if the file is there, but corrupted?

prereise/gather/griddata/hifld/const.py
danielolsen: That'll teach me to type manually, rather than copy/paste.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -322,6 +328,105 @@ def create_transformers(bus):
     return pd.DataFrame(bus_pairs, columns=["from_bus_id", "to_bus_id"])
 
 
+def estimate_branch_impedance(branch, bus):
+    """Estimate branch impedance using transformer voltages or line voltage and length.
+
+    :param pandas.Series branch: data for a single branch (line or transformer). All
+        branches require 'type' attributes, lines require 'VOLTAGE' and 'length',
+        transformers require 'from_bus_id' and 'to_bus_id'.
```
BainanXia, line 336: Missing `bus` parameter in the doc string here.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: You're right. Do you think it would make sense to correct this when we first load the line data, rather than remember to reverse them within `data_process`? I think everywhere else in our codebase we generally expect (lat, lon).

YifanLi86:
> @YifanLi86 visualizing the distribution of impedances is useful, but I'm still not quite certain what `Average out... and cut by 75% quantile` means as a formula/algorithm. Is it just taking the value at the 75th percentile (and if so, what's the logic)? Or it is excluding upper-quartile outliers and then calculating an average? If so, is there any significance to 75% as a cutoff? And should we exclude lower-end outliers too?

Oh I see what you mean. Yes, it is cutting the top AND bottom 25% "outliers", and average out the remaining. The 75% is not strict science, just from observations in the charts and past practices that seems works best - no exact literature that demonstrates it is the best cutoff to use.

prereise/gather/griddata/hifld/data_process/transmission.py
rouille: `powersimdata/utility/distance.find_closest_neighbor` that I use in #217 uses (lon, lat). I can reverse it if you decide to modify the `COORDINATES` column

danielolsen: I think that would be cleaner than having to remember that in this part of the codebase, we have (lon, lat), where everywhere else we have (lat, lon).

BainanXia: Making it consistent across the code base is definitely cleaner. Just keep in mind we will need to refactor the functions and scripts where we call the corresponding functions as well.

rouille: I think we are just talking to flip lon and lat in the `COORDINATES` column of the lines data frame

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -322,6 +328,105 @@ def create_transformers(bus):
...
+        index of the first connected component (int),
+        index of the second connected component (int),
+        a dictionary with keys containing ``start``, ``end`` and ``weight``, defines
+        the ``from substation ID``, ``to substation ID`` and the distance of the line.
+    """
+    cache_dir = os.path.join(os.path.dirname(__file__), "cache")
+    os.makedirs(cache_dir, exist_ok=True)
+    cache_key = repr(lines[["SUB_1_ID", "SUB_2_ID"]].to_numpy().tolist()) + repr(
+        substations[["LATITUDE", "LONGITUDE"]].to_numpy().tolist()
+    )
```
BainanXia, line 416: Given we are going to encode it anyways, do we need `to_numpy()` and `tolist()` here?

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I made https://github.com/Breakthrough-Energy/PreREISE/pull/223, which I think can go in upstream of this PR and #217 to avoid subsequent refactor PRs.

prereise/gather/griddata/hifld/const.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Calling `repr` on a data frame or numpy array seems to truncate for large arrays:
```python
>>> import numpy as np
>>> a = np.arange(100)
>>> df = pd.DataFrame({"a": a})
>>> print(repr(df))
     a
0    0
1    1
2    2
3    3
4    4
..  ..
95  95
96  96
97  97
98  98
99  99

[100 rows x 1 columns]
```

numpy will happily display 100-length arrays, even 1,000-length arrays, but starts to truncate output somewhere between there and 10,000:
```python
>>> a = np.arange(10000)
>>> print(repr(a))
array([   0,    1,    2, ..., 9997, 9998, 9999])
```

BainanXia: Does it imply it will give us same cache_key for different big data frames or it is simply an issue for display?

danielolsen: Yes.
```python
>>> import numpy as np
>>> import pandas as pd
>>> a = np.arange(100)
>>> df1 = pd.DataFrame({"a": a})
>>> df2 = pd.DataFrame({"a": a})
>>> df2.loc[5:94] = -999
>>> df2
     a
0    0
1    1
2    2
3    3
4    4
..  ..
95  95
96  96
97  97
98  98
99  99

[100 rows x 1 columns]
>>> repr(df1) == repr(df2)
True
>>> df1.equals(df2)
False
```

danielolsen:
@YifanLi86 I've implemented what I understand your method to be, but the values I'm getting don't quite match with the constants that you provided (generally a bit greater):
```python
>>> from powersimdata import Grid
>>> from powersimdata.utility.distance import haversine
>>> line_length_factor = 1.2
>>> grid = Grid("USA")
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
>>> lines = grid.branch.query(
...     "branch_device_type == 'Line' and (from_lat != to_lat or from_lon != to_lon)"
... ).copy()
>>> lines["length"] = line_length_factor * lines.apply(
...     lambda x: haversine((x.from_lat, x.from_lon), (x.to_lat, x.to_lon)),
...     axis=1
... )
>>> lines["x_per_mile"] = lines["x"] / lines["length"]
>>> lines["voltage"] = lines["from_bus_id"].map(grid.bus["baseKV"])
>>> lower_quartile_cutoff = lines.groupby("voltage")["x_per_mile"].quantile(0.25)
>>> upper_quartile_cutoff = lines.groupby("voltage")["x_per_mile"].quantile(0.75)
>>> middle_half_lines = lines.loc[
...     (lines["voltage"].map(lower_quartile_cutoff) <= lines["x_per_mile"])
...     & (lines["x_per_mile"] <= lines["voltage"].map(upper_quartile_cutoff))
... ]
>>> middle_half_lines.groupby("voltage")["x_per_mile"].mean()
voltage
69.0     0.011171
100.0    0.007372
115.0    0.004533
138.0    0.003376
161.0    0.002480
230.0    0.001219
345.0    0.000475
500.0    0.000209
765.0    0.000094
Name: x_per_mile, dtype: float64
```

Is there something I'm doing differently?

YifanLi86:
> @YifanLi86 I've implemented what I understand your method to be, but the values I'm getting don't quite match with the constants that you provided (my values are generally a bit larger):
> 
> ```python
> >>> from powersimdata import Grid
> >>> from powersimdata.utility.distance import haversine
> >>> line_length_factor = 1.2
> >>> grid = Grid("USA")
> Reading bus.csv
> Reading plant.csv
> Reading gencost.csv
> Reading branch.csv
> Reading dcline.csv
> Reading sub.csv
> Reading bus2sub.csv
> Reading zone.csv
> >>> lines = grid.branch.query(
> ...     "branch_device_type == 'Line' and (from_lat != to_lat or from_lon != to_lon)"
> ... ).copy()
> >>> lines["length"] = line_length_factor * lines.apply(
> ...     lambda x: haversine((x.from_lat, x.from_lon), (x.to_lat, x.to_lon)),
> ...     axis=1
> ... )
> >>> lines["x_per_mile"] = lines["x"] / lines["length"]
> >>> lines["voltage"] = lines["from_bus_id"].map(grid.bus["baseKV"])
> >>> lower_quartile_cutoff = lines.groupby("voltage")["x_per_mile"].quantile(0.25)
> >>> upper_quartile_cutoff = lines.groupby("voltage")["x_per_mile"].quantile(0.75)
> >>> middle_half_lines = lines.loc[
> ...     (lines["voltage"].map(lower_quartile_cutoff) <= lines["x_per_mile"])
> ...     & (lines["x_per_mile"] <= lines["voltage"].map(upper_quartile_cutoff))
> ... ]
> >>> middle_half_lines.groupby("voltage")["x_per_mile"].mean()
> voltage
> 69.0     0.011171
> 100.0    0.007372
> 115.0    0.004533
> 138.0    0.003376
> 161.0    0.002480
> 230.0    0.001219
> 345.0    0.000475
> 500.0    0.000209
> 765.0    0.000094
> Name: x_per_mile, dtype: float64
> ```
> 
> Is there something I'm doing differently?

hmmm....let me go back to verify what I did before. I will get back soon.

danielolsen:
I'm also getting different values for transformer impedances:
```python
>>> from powersimdata import Grid
>>> from powersimdata.utility.distance import haversine
>>> grid = Grid("USA")
>>> transformers = grid.branch.query("branch_device_type == 'Transformer'").copy()
>>> transformers["voltage_tuple"] = transformers.apply(
...     lambda x: tuple(sorted(grid.bus.loc[[x.from_bus_id, x.to_bus_id], "baseKV"])),
...     axis=1,
... )
>>> lower_quartile_cutoff = transformers.groupby("voltage_tuple")["x"].quantile(0.25)
>>> upper_quartile_cutoff = transformers.groupby("voltage_tuple")["x"].quantile(0.75)
>>> middle_half = transformers.loc[
...     (transformers["voltage_tuple"].map(lower_quartile_cutoff) <= transformers["x"])
...     & (transformers["x"] <= transformers["voltage_tuple"].map(upper_quartile_cutoff))
... ]
>>> average_impedances = middle_half.groupby("voltage_tuple")["x"].mean()
>>> interesting_voltage_pairs = [(v1 != v2) and (v1 >= 69) for v1, v2 in average_impedances.index]
>>> average_impedances.loc[interesting_voltage_pairs]
voltage_tuple
(69.0, 100.0)     0.047467
(69.0, 115.0)     0.225843
(69.0, 138.0)     0.138248
(69.0, 161.0)     0.119146
(100.0, 115.0)    0.254849
(100.0, 138.0)    0.147409
(100.0, 161.0)    0.133274
(100.0, 230.0)    0.058539
(100.0, 345.0)    0.026318
(115.0, 138.0)    0.130176
(115.0, 161.0)    0.107011
(115.0, 230.0)    0.053260
(115.0, 345.0)    0.026548
(115.0, 500.0)    0.014705
(138.0, 161.0)    0.113119
(138.0, 230.0)    0.057111
(138.0, 345.0)    0.026793
(161.0, 230.0)    0.055645
(161.0, 345.0)    0.025749
(161.0, 500.0)    0.014462
(230.0, 345.0)    0.026506
(230.0, 500.0)    0.014171
(345.0, 500.0)    0.014385
(345.0, 765.0)    0.007809
(500.0, 765.0)    0.008651
Name: x, dtype: float64
```

prereise/gather/griddata/hifld/const.py
```diff
@@ -118,3 +118,67 @@
...
+    (161, 230): 0.06539,
+    (161, 345): 0.03293,
+    (161, 500): 0.06978,
+    (230, 345): 0.02085,
+    (230, 500): 0.01846,
+    (230, 765): 0.01616,
+    (345, 500): 0.01974,
+    (345, 765): 0.01625,
+    (500, 765): 0.00436,
+}
```
rouille, line 183: Why not all the combinations are in the dictionary? Like 69-100, 161-765, etc.

danielolsen: These came from an external document, we're working on validating/refining this table (see https://github.com/Breakthrough-Energy/PreREISE/pull/219#issuecomment-920286423). If anything combination is missing, the 'nearest' one should be chosen automatically.

rouille: Sorry, I did not really follow the discussion

YifanLi86:
@danielolsen I just went it over, it has been a while and some back and forth on the method to use, actually below is the exact process utilized to generate the reactance values listed above:

1. Transmission reactance: the 25% quantile method has been tested, but we eventually chose to directly average out all lines > 10miles, from original csv inputs. The work is done in excel before, I have rewrote the code which will generate the exact numbers: (be aware if we switch to 1.3 length factor, these values will be changed as well)
> 
> ```python
> >>> import pandas as pd
> >>> from powersimdata.utility.distance import haversine
> >>> line_length_factor = 1.2
> >>> branch = pd.read_csv('C:/Users/YifanLi/Desktop/Packages/PowerSimData-0.4.4/PowerSimData-0.4.4/powersimdata/network/usa_tamu/data/branch.csv')
> >>> bus = pd.read_csv('C:/Users/YifanLi/Desktop/Packages/PowerSimData-0.4.4/PowerSimData-0.4.4/powersimdata/network/usa_tamu/data/bus.csv')
> >>> sub = pd.read_csv('C:/Users/YifanLi/Desktop/Packages/PowerSimData-0.4.4/PowerSimData-0.4.4/powersimdata/network/usa_tamu/data/sub.csv')
> >>> bus2sub = pd.read_csv('C:/Users/YifanLi/Desktop/Packages/PowerSimData-0.4.4/PowerSimData-0.4.4/powersimdata/network/usa_tamu/data/bus2sub.csv')
> >>> bus2sub_copy = bus2sub.copy()
> >>> bus_copy = bus.copy()
> >>> sub_copy = sub.copy()
> >>> branch_copy = branch.copy()
> >>> bus2sub_copy['lat'] = bus2sub_copy['sub_id'].map(pd.Series(sub.lat.values,index=sub.sub_id).to_dict())
> >>> bus2sub_copy['lon'] = bus2sub_copy['sub_id'].map(pd.Series(sub.lon.values,index=sub.sub_id).to_dict())
> >>> bus_copy['lat'] = bus_copy['bus_id'].map(pd.Series(bus2sub_copy.lat.values,index=bus2sub_copy.bus_id).to_dict())
> >>> bus_copy['lon'] = bus_copy['bus_id'].map(pd.Series(bus2sub_copy.lon.values,index=bus2sub_copy.bus_id).to_dict())
> >>> branch_copy['from_bus_lat'] = branch_copy['from_bus_id'].map(pd.Series(bus_copy.lat.values,index=bus_copy.bus_id).to_dict())
> >>> branch_copy['from_bus_lon'] = branch_copy['from_bus_id'].map(pd.Series(bus_copy.lon.values,index=bus_copy.bus_id).to_dict())
> >>> branch_copy['to_bus_lat'] = branch_copy['to_bus_id'].map(pd.Series(bus_copy.lat.values,index=bus_copy.bus_id).to_dict())
> >>> branch_copy['to_bus_lon'] = branch_copy['to_bus_id'].map(pd.Series(bus_copy.lon.values,index=bus_copy.bus_id).to_dict())
> >>> branch_copy["length"] = 1.2 * branch_copy.apply(lambda x: haversine((x.from_bus_lat, x.from_bus_lon), (x.to_bus_lat,x.to_bus_lon)),axis=1)
> >>> branch_copy["x_per_mile"] = branch_copy["x"]/branch_copy["length"]
> >>> branch_filtered = branch_copy[(branch_copy.length > 10) & (branch_copy.branch_device_type == 'Line')]
> >>> branch_filtered["from_bus_kV"] = branch_filtered["from_bus_id"].map(pd.Series(bus.baseKV.values,index=bus.bus_id).to_dict())
> >>> branch_filtered["to_bus_kV"] = branch_filtered["to_bus_id"].map(pd.Series(bus.baseKV.values,index=bus.bus_id).to_dict())
> >>> x_per_mile_by_kV = branch_filtered.groupby('from_bus_kV')['x_per_mile'].mean()
> >>> print(x_per_mile_by_kV)
```
2. For transformer and phase shifter reactance, it was originally using the quantile method as well, but there seems to have some negative reactance, so the current number is actually from ERAG.

So decisions to be made:

for the 10mile cutoff for lines, it is not a science or no rigid support. Open to suggestions to adjust the method. Or I can go back to try to benchmark to industry cases to double check whichever method (either quantile or cutoff, or anything else with calibrated threshold) that makes better sense, given that we have more data available right now.

for the xfmr and ps, if we want to refer to a source that can be fully published, I will have to go back and do the same thing, benchmarking based on the better data we have now, rather than using ERAG directly.

Anyway, I suspect no matter what method we will be using, it will still might be quite different than real system for a certain local area.

Thoughts?

**Comment by danielolsen** [comment]
Thanks! Just as a heads-up, when we load a Grid object, it automatically adds the `from_lat`, `from_lon`, `to_lat`, and `to_lon` attributes to the original data in the **branch.csv** file, so you don't need to do it manually:
```python
>>> from powersimdata import Grid
>>> from powersimdata.utility.distance import haversine
>>> line_length_factor = 1.2
>>> grid = Grid("USA")
>>> lines = grid.branch.query("branch_device_type == 'Line'").copy()
>>> lines["length"] = line_length_factor * lines.apply(
...     lambda x: haversine((x.from_lat, x.from_lon), (x.to_lat, x.to_lon)),
...     axis=1,
... )
>>> lines["x_per_mile"] = lines["x"] / lines["length"]
>>> lines["voltage"] = lines["from_bus_id"].map(grid.bus["baseKV"])
>>> lines.query("length > 10").groupby("voltage")["x_per_mile"].mean()
voltage
69.0     0.009577
100.0    0.006326
115.0    0.003883
138.0    0.002645
161.0    0.002149
230.0    0.001138
345.0    0.000471
500.0    0.000208
765.0    0.000094
Name: x_per_mile, dtype: float64
```
These values do match with the document, once we round to 4 decimal places, but we should probably specify with a certain number of significant digits instead to avoid distorting the higher-voltage impedances too much (or directly calculate the values in code and use them directly, for maximum transparency).

What about the transformers, what was the process used to generate the values in the document?

No matter which set of values we choose, it is of course not going to match local results perfectly, but we can choose the approach that best matches external benchmarks and justify our decision that way.

YifanLi86:
> Thanks! Just as a heads-up, when we load a Grid object, it automatically adds the `from_lat`, `from_lon`, `to_lat`, and `to_lon` attributes to the original data in the **branch.csv** file, so you don't need to do it manually:
> 
> ```python
> >>> from powersimdata import Grid
> >>> from powersimdata.utility.distance import haversine
> >>> line_length_factor = 1.2
> >>> grid = Grid("USA")
> >>> lines = grid.branch.query("branch_device_type == 'Line'").copy()
> >>> lines["length"] = line_length_factor * lines.apply(
> ...     lambda x: haversine((x.from_lat, x.from_lon), (x.to_lat, x.to_lon)),
> ...     axis=1,
> ... )
> >>> lines["x_per_mile"] = lines["x"] / lines["length"]
> >>> lines["voltage"] = lines["from_bus_id"].map(grid.bus["baseKV"])
> >>> lines.query("length > 10").groupby("voltage")["x_per_mile"].mean()
> voltage
> 69.0     0.009577
> 100.0    0.006326
> 115.0    0.003883
> 138.0    0.002645
> 161.0    0.002149
> 230.0    0.001138
> 345.0    0.000471
> 500.0    0.000208
> 765.0    0.000094
> Name: x_per_mile, dtype: float64
> ```
> 
> These values do match with the document, once we round to 4 decimal places, but we should probably specify with a certain number of significant digits instead to avoid distorting the higher-voltage impedances too much (or directly calculate the values in code and use them directly, for maximum transparency).
> 
> What about the transformers, what was the process used to generate the values in the document?
> 
> No matter which set of values we choose, it is of course not going to match local results perfectly, but we can choose the approach that best matches external benchmarks and justify our decision that way.

Thanks Daniel! Ok just for everyone's information, conversation has been going on in slack, for transformer and phase shifter, the current data is a summary of ERAG data which is not convenient to share or refer to. After discussion, the work going forward looks like: 1. try to find a method to derive these data from our grid data first, and benchmark it using data we have in compliance to all data requirements; 2. try to derive general numbers from PLEXOS WECC data we have now (pending legal check and check with EE for permission to release); 3. wait to get more EI data and reiterate the analysis including EI, since the Gridlab MISO data is not authorized to use outside of the Gridlab MISO project. So the help it can provide is very limited.

rouille (approved):
Thanks

YifanLi86 (approved):

Issue #13, opened 2018-10-20
fix test_call in call module


kasparm:
call.launch_scenario_performance('western_scenarioUnitTest02', 16)

PR #220, opened 2021-09-09
UCI Vehicle Charging Translation Update 2

josephxu1:

### Purpose
To add the file that contains code for the immediate charging scenario.

### What the code is doing
How is the purpose executed?
This code generates charging profiles by iterating through NHTS census data and doing calculations with them.

### Testing
The code has not yet been tested but we will test it by running it with the same inputs as the original MATLAB code and comparing outputs.

### Where to look
The new code is located at PreREISE/prereise/gather/demanddata/transportation_electrification/immediate.py

The file named "immediate.py" (this is the only new file in this PR)

### Usage Example/Visuals

### Time estimate
Should not take more than a week or two.


danielolsen:
Could you please post a minimum-working-example of how you would use the main user-facing function(s), and what they return?

rouille:
> Could you please post a minimum-working-example of how you would use the main user-facing function(s), and what they return?

@josephxu1, you can edit the PR description and write fill out the section (Puropose, What is the code is doing, etc) and, in particular under the **Testing** section, give a code snippet and corresponding outputs as @danielolsen mentioned

josephxu1:
> > Could you please post a minimum-working-example of how you would use the main user-facing function(s), and what they return?
> 
> @josephxu1, you can edit the PR description and write fill out the section (Puropose, What is the code is doing, etc) and, in particular under the **Testing** section, give a code snippet and corresponding outputs as @danielolsen mentioned

@danielolsen @rouille I will ask Kate for an example input/output case and edit the Pull Request to have that as soon as possible.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
+# DELETED: comm_type, LOAD, vmtsc
+
+import mileage
+import charging
```
danielolsen, line 4: Imports should be absolute, not relative.

danielolsen, line 4: What is this `charging` module? It doesn't seem to be present in this branch.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
...
+    newdata = mileage.remove_ldt(mileage.load_data(census_region))
+    # add new columns to newdata to store data that is not in NHTS data
+    newdata.insert(28, "trip start battery charge", [0] * len(newdata))
+    newdata.insert(29, "trip end battery charge", [0] * len(newdata))
+    newdata.insert(30, "charging power", [0] * len(newdata))
+    newdata.insert(31, "charging time", [0] * len(newdata))
+    # jx: never actually used... delete?
+    newdata.insert(32, "charging consumption", [0] * len(newdata))  # needs double check
+    newdata.insert(33, "BEV could be used", [0] * len(newdata))
+    newdata.insert(34, "trip number", [0] * len(newdata))
```
danielolsen, line 31: If all we're doing is initializing new columns full of zeros, we can do simply: `newdata["trip start battery charge"] = 0`, or better yet we can avoid pre-initializing at all and just add the data to the data frame once it's calculated.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
...
+    battery_strategy = 1
+
+    # 1-anytrip number, 2-last trip
+    trip_strategy = 1
+
+    kwh = kwhmi * veh_range
+
+    for day_iter in range(len(input_day)):
+
+        electricload = np.zeros(4800)
```
danielolsen, line 55: What is the purpose of this number?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
...
+
+    TRANS_charge = np.zeros(24 * len(input_day))
+
+    data_day = mileage.get_data_day(newdata)
+
+    actual_vmt = 0
+    potential_vmt = 0
+
+    # from grid to battery efficiency
+    charging_efficiency = 0.9
```
rouille, line 43: It can probably be moved into the `const` module

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
...
+                    newdata.loc[i]["trip number"] = trip_num
+                else:
+                    newdata.loc[i]["trip start battery charge"] = kwh
+                    trip_num = 1
+                    # trip number
+                    newdata.loc[i]["trip number"] = trip_num
+
+                # battery electric vehicle
+                # pure electric
+                ER = 1
```
rouille, line 93: it can probably be moved into the `const` module

prereise/gather/demanddata/transportation_electrification/immediate.py
josephxu1: The charging module contains the helper functions that Meg and I have previously mentioned. This will be in Pull Request soon. Expect Monday next week.

prereise/gather/demanddata/transportation_electrification/immediate.py
josephxu1: Fixed

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
+# DELETED: comm_type, LOAD, vmtsc
+
```
danielolsen, line 2: There are several comments like this that seem to pertain more to the code development process than the code itself, and can probably be removed.

josephxu1: Yup this is absolutely correct. If you see comments like this please ignore them haha, they are more for me than anything.

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,140 @@
...
+    df = pd.DataFrame(raw_data, columns=const.nhts_census_column_names)
+
+    return df
+
+
+def remove_ldt(data: pd.DataFrame) -> pd.DataFrame:
+    """Remove light duty trucks from data loaded from nths_census.mat.
+
+    :param pd.DataFrame data: the data returned from load_data.
+    :return: (*pd.DataFrame*) -- the data loaded from load_data with all rows
```
rouille, line 77: load_data --> :func:\`load_data\`

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
...
+                    newdata.loc[i]["depleting time"] = newdata.loc[i][
+                        "Travel time (hour decimal)"
+                    ]
+
+                else:
+                    # 0 means the day trip could not be used in battery electric vehicle
+                    newdata.loc[i]["BEV could be used"] = 0
+
+                # fuel consumption is 0
+                newdata.loc[i]["fuel consumption"] = 0
```
danielolsen, line 114: If this is the same for every entry, can it be assigned to the whole column outside of the loop?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,194 @@
...
+                    trip_num = 1
+                    # trip number
+                    newdata.loc[i]["trip number"] = trip_num
+
+                # battery electric vehicle
+                # pure electric
+                ER = 1
+
+                # 1 is the safety coefficient
+                if newdata.loc[i]["total vehicle miles traveled"] < veh_range * 1:
```
danielolsen, line 96: If this `1` value is meaningful and could be changed, we should give it a name and put it into `const.py`.

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,140 @@
...
+    input_day: list[int],
+) -> np.array[float]:
+    """load data and use the parameters to calculate total_daily_vmt.
+
+    :param int census_region: the type of census
+    :param int comm_type: the type of Commute
+    :param list locationstrategy: strategy for each location
+    :param list input_day: day of the week for each day in the year derived from
+        first_func
+    :return: (*list*) -- daily_vmt_total each row is a year of entries for each vehicle
```
rouille, line 99: list --> np.array for the returned variable type. Also, the name (`daily_vmt_total`) of the should not be mentioned in the description. For an array I would use element in place of row

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,140 @@
...
+    return pd.date_range(
+        start=f"{model_year}-01-01", end=f"{model_year}-12-31", freq="D"
+    )
+
+
+def get_input_day(model_year_dti: pd.DatetimeIndex) -> np.ndarray:
+    """Determine whether each day of the model year is a weekend (1) or weekday (2)
+
+    :param pd.DatetimeIndex model_year_dti: a DatetimeIndex encompassing the model year.
+    :return: (*np.ndarray*) array of 1s and 2s indicating weekend/weekday designations for the model year.
```
rouille, line 18: missing -- after closing parenthesis

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,140 @@
+import numpy as np
+import pandas as pd
+from scipy.io import loadmat
+
+from prereise.gather.demanddata.transportation_electrification import const
+
+
+def get_model_year_dti(model_year: int) -> pd.DatetimeIndex:
```
rouille, line 8: missing docstring

prereise/gather/demanddata/transportation_electrification/immediate.py
josephxu1: This should actually be deleted.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,191 @@
...
+                        + newdata.loc[i - 1]["depleting time"]
+                    )
+                    # trip number
+                    trip_num += 1
+                    newdata.loc[i]["trip number"] = trip_num
+                else:
+                    newdata.loc[i]["trip start battery charge"] = kwh
+                    trip_num = 1
+                    # trip number
+                    newdata.loc[i]["trip number"] = trip_num
```
danielolsen, line 86: It seems that most of what we're doing here is assigning trip numbers based on grouping by vehicle number, and then assigning `"trip start battery charge"` based on the value within the trip number. The former can be accomplished outside of a loop using something like `newdata.loc["trip number"] = newdata.groupby("sample vehicle number").rank(method="first")`.

prereise/gather/demanddata/transportation_electrification/immediate.py
josephxu1: I have made the variable safety_coefficient and stored it in const.py. I will further talk to Kate and ask for her opinion, but this should be good for now.

prereise/gather/demanddata/transportation_electrification/immediate.py
danielolsen: Is there an implicit assumption here that the data entries are in chronological order?

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,140 @@
...
+    data.loc[data["Day of Week"].isin(list(range(2, 7))), "If Weekend"] = 2
+
+    data_day = get_data_day(data)
+    daily_vmt_total = np.zeros(len(input_day))
+
+    for day_iter in range(len(input_day)):
+        for i in range(n):
+            if data_day[i] == input_day[day_iter]:
+                daily_vmt_total[day_iter][0] += data[i][12]
+                daily_vmt_total[day_iter][1] += 1
```
danielolsen, line 138: It seems that these loops are still performing weekday/weekend grouping, without any month grouping.

Either way, this should be refactored to a two-step process, rather than nested loops:
1. Group trips (by either weekday/weekend alone, or by weekday/weekend and month), then count trips and sum mileage.
2. Map the collected information from (1) to the `daily_vmt_total` output based on the attributes of the days we want to return.

With nested loops, it seems we're performing the summing across all `data_day` entries _for each day we want to return data for_, so we're performing two orders of magnitude more computation than we need to.

prereise/gather/demanddata/transportation_electrification/immediate.py
danielolsen: It seems like maybe what we're doing is calculating electrical load over two days by the hundredth of an hour, before downsampling to hourly resolution. If so, we should be able to do this in a more automated way using the `resample` method of a time-indexed data frame: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.resample.html

prereise/gather/demanddata/transportation_electrification/mileage.py
danielolsen: Some more context: formatting the docstring this way results in Sphinx automatically creating a link to the referenced function within the documentation. See https://breakthrough-energy.github.io/docs/powersimdata.utility.html#powersimdata.utility.distance.angular_distance for an example.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,184 @@
+# DELETED: comm_type, LOAD, vmtsc
+
+import prereise/gather/demanddata/transportation_electrification/mileage
+import prereise/gather/demanddata/transportation_electrification/charging
```
danielolsen, line 4: The syntax for these imports needs to be corrected, they currently throw a SyntaxError.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,184 @@
...
+                    trip_num = 1
+                    # trip number
+                    newdata.loc[i]["trip number"] = trip_num
+
+                # battery electric vehicle
+                # pure electric
+                ER = 1
+
+                # 1 is the safety coefficient
+                if newdata.loc[i]["total vehicle miles traveled"] < veh_range * const.safety_coefficient:
```
danielolsen, line 90: There will also be an import error here since we haven't imported `const` into this namespace.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,184 @@
...
+            if data_day[i] == input_day[day_iter]:
+                if (
+                    i > 0
+                    and newdata.loc[i]["sample vehicle number"]
+                    == newdata.loc[i - 1]["sample vehicle number"]
+                ):
+                    newdata.loc[i]["trip start battery charge"] = (
+                        newdata.loc[i - 1]["trip end battery charge"]
+                        + newdata.loc[i - 1]["depleting time"]
+                    )
```
danielolsen, line 75: This style of assignment for pandas dataframes (`df.loc[a][b] = `) is discouraged, see https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#why-does-assignment-fail-when-using-chained-indexing for more details. Dataframe values should be equivalently (and more safely) assigned using `df.loc[a, b] = `.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,184 @@
...
+        trip_num = 1  # trip number for current vehicle
+        # charging start and end points
+        start_point = 0
+        end_point = 0
+
+        # flag to see if the trip is the first for the vheicle, '1' yes, '0' no
+        firstrip = 0
+
+        # meh: day2 = data_day, month2 = data_month
+        for i in range(len(newdata)):
```
danielolsen, line 65: Looping over every day of the outputs you want to generate, and then within that loop looping over every trip record in the input data, is liable to be extremely slow. Are there calculations within this inner loop that can be done once outside of the outer loop, rather than repeated within?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,184 @@
...
+):
+    """Immediate charging function
+
+    :param int census_region: any of the 9 census regions defined by US census bureau.
+    :param int model_year: year that is being modelled/projected to, 2017, 2030, 2040, 2050.
+    :param int veh_range: 100, 200, or 300, represents how far vehicle can travel on single charge.
+    :param int kwhmi: fuel efficiency, should vary based on vehicle type and model_year.
+    :param int power: charger power, EVSE kW.
+    :param int location_strategy: where the vehicle can charge-1, 2, 3, 4, or 5; 1-home only, 2-home and work related, 3-anywhere if possibile, 4-home and school only, 5-home and work and school.
+    :return: (*numpy.ndarray*) -- charging profiles.
```
danielolsen, line 19: Is the output of this function one profile, or multiple? It looks like the result is a 1-D array that's 8760 or 8784 entries long, depending on whether it's a leap year or not.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,184 @@
...
+
+    kwh = kwhmi * veh_range
+
+    for day_iter in range(len(input_day)):
+
+        electricload = np.zeros(4800)
+        # initializes electricload_old
+
+        # vmt, trip amount, vehicle amount
+        info1 = [0] * 3
```
danielolsen, line 53: What is the purpose of this list? It seems that the second and third entries get modified, but never actually used.

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,121 @@
...
+    #home and school only
+    elif location_strategy == 4:
+        #21 is 'go to school as student'
+        if dwell_location == 1 or dwell_location == 21:
+            return 1
+        else:
+            return 0
+    #home and work and school
+    elif location_strategy == 5:
+        if dwell_location == 1 or dwell_location == 11 or dwell_location == 12 or dwell_location == 21:
```
danielolsen, line 88: `if dwell_location in {1, 11, 12, 21}:`

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,121 @@
...
+
+#consider battery SOC
+def get_battery_SOC(battery_SOC, kwh):
+    '''Determines if the vehicle needs to charge
+
+    :param float battery_SOC: vehicle battery at end of trip
+    :param float kwh: kwhmi * veh_range, amount of energy needed to charge vehicle.
+    :return: (int) -- a flag that determines if battery needs to be charged (1 yes, 0 no)
+    '''
+    return int(battery_SOC<kwh)
```
danielolsen, line 120: These functions with effectively boolean returns would be clearer if returned as such, rather than converting to `int`s.

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,121 @@
...
+    #charging_charge to the battery
+    return charging_power * charging_time * charging_efficiency
+
+
+
+#charge according to location
+def get_location(location_strategy, dwell_location):
+    '''Determines if the vehicle can be charged given location strategy and dwelling location
+
+    :param int location_strategy: where the vehicle can charge-1, 2, 3, 4, or 5; 1-home only, 2-home and work related, 3-anywhere if possibile, 4-home and school only, 5-home and work and school.
```
danielolsen, line 58: All lines should wrap to 88 characters or less, unless they're too long to be wrapped conveniently (e.g. a long URL).

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,121 @@
...
+    :param float charging_power: charging power.
+    :param float kwh: kwhmi * veh_range, amount of energy needed to charge vehicle.
+    :param float battery_SOC: battery charge at end of trip.
+    :param float charging_efficiency: grid to battery efficiency.
+    :return: (*float*) -- charging time in decimal format.
+    '''
+    if charging_power == 0:
+        charging_time = 0
+    else:
+        charging_time = (kwh - battery_SOC)/charging_power/charging_efficiency
```
danielolsen, line 37: I think this would be clearer as `charging_time = (kwh - battery_SOC)/(charging_power * charging_efficiency)`.

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,121 @@
...
+def get_charging_consumption(charging_power, charging_time, charging_efficiency):
+    '''Calculates power consumption of the charging session
+
+    :param float charging_power: charging power.
+    :param float charging_time: charging time in decimal format
+    :param float charging_efficiency: grid to battery efficiency.
+    :return: (*float*) -- power consumption of the charging session.
+    '''
+    #charging_charge to the battery
+    return charging_power * charging_time * charging_efficiency
```
danielolsen, line 50: This function seems unnecessary, simply multiplying the three values together should be clearer.

danielolsen:
Sorry for all the weird target-changing; I think the target needs to be `develop` in order for new commits to trigger the automated workflow run, even though this will eventually get merged to the transportation-electrification feature branch.

josephxu1:
> Sorry for all the weird target-changing; I think the target needs to be `develop` in order for new commits to trigger the automated workflow run, even though this will eventually get merged to the `transportation_electrification` feature branch.

Sorry, have I not been doing that? I was under the impression that all the changes I'm committing were to the develop branch

danielolsen:
@josephxu1 this isn't anything you need to worry about. Your commits have been to the branch named `develop` in your fork, but ultimately we will be merging the commits into the `transportation_electrification` branch in our repo, since this feature will likely have multiple PRs and we don't want to merge to our `develop` until everything is done. However, it seems like Github doesn't run the automated checks on code formatting, testing, etc. unless the PR target is our `develop` branch, so I've changed it back to that for the purpose of reviewing.

prereise/gather/demanddata/transportation_electrification/immediate.py
josephxu1: Right now we only use info1[0] (vmt) because it's used for the output. But you'll notice that we use it in a commented out section. This is just for Kate and I to test the code.

>  

        # only used as "test" output variables
        # output_load.append(outputelectricload)
        # info_out.append(info1)

prereise/gather/demanddata/transportation_electrification/charging.py
josephxu1: I addressed this but I am going to leave this issue unresolved for now because I think the conditions right now are a little bit confusing to say the least. For example

> elif location_strategy == 2:
          #13 is to 'attend business meeting', 14 is 'other work related'. 
          #here only consider the regular work, which are 11 and 12, 'go to work' and 'return to work'
          return dwell_location < 13
>
is definitely not the final code we want going into your codebase. A more clear solution will be needed.

prereise/gather/demanddata/transportation_electrification/immediate.py
keforres: We can reduce the loops by running the trip calculations for two representative days: (1) Weekday and (2) Weekend. Once we have the profiles for those two days, we can replicate the profiles across the year. With this approach, the code runs through the trip dataset twice, instead of 365/366 times. (Weight factors for day of week and month can be applied afterwards.)

danielolsen: For the innermost loop, it seems there are some things that may only need to happen once, e.g. assigning trip numbers and calculating starting/end SoCs for a given vehicle number (unless these will also differ by weekday/weekend?). Then it seems like we could use that data to come up with a profile for weekdays and weekends, and then look up these pre-built profiles as you suggested.

Where do the month and day-of-week weighting factors come from? Should they be being applied within this function, or will that be a follow-up step?

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,101 @@
+def get_charging_power(power, trip_strategy, location_strategy, kwh, trip_data):
+    '''Determines the charging power. Returns either param power or 0.
```
rouille, line 2: docstrings should start with 3 double quotes and end with three double quotes:
```
"""

"""
```

danielolsen:
Closing for now following conversations with @isond, since we're expecting a new PR for the rewritten module.

PR #223, opened 2021-09-15
refactor: switch (lon, lat) points to (lat, lon)

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Switch [(lon, lat), (lon, lat), ..] points to [(lat, lon), (lat, lon), ...] when we load coordinates, to match our convention in the rest of the codebase. See conversation at https://github.com/Breakthrough-Energy/PreREISE/pull/219#discussion_r707615437

### What the code is doing
It's a fancy one-liner, applied to each coordinates list in the data frame. The inner-most `list`/`zip` pair collects the `[(lon, lat), (lon, lat), ..]` points as `[(lon, lon, ...), (lat, lat, ...)]`, then this 2-length list is reversed, and the outer-most `list`/`zip` re-expands to `[(lat, lon), (lat, lon), ...]`. A simple example:
```python
>>> a = [(1, 2), (3, 4), (5, 6)]
>>> list(zip(*list(zip(*a))[::-1]))
[(2, 1), (4, 3), (6, 5)]
```

### Testing
Tested manually. Before:
```python
>>> from prereise.gather.griddata.hifld import const
>>> from prereise.gather.griddata.hifld.data_access import load
>>> lines = load.get_hifld_electric_power_transmission_lines(const.blob_paths["transmission_lines"])
>>> lines.iloc[0]["COORDINATES"]
[...]

PR #224, opened 2021-09-20
refactor: change transmission processing code to expect (lat, lon) instead of (lon, lat)

danielolsen:

### Purpose
This is a follow-up to #223. Now that we have points as (lat, lon) instead of (lon, lat), we need to change the code that uses coordinates.

### What the code is doing
Undoing the list reversal `[::-1]`.

### Testing
Manually.

### Usage Example
Before, all lines are dropped because they're evaluated as 'mismatches', since the (lat, lon) location is very far away from the (lon, lat) location:
```python
dropping 67652 lines with one or more substations with non-matching coordinates out of a starting total of 67652
dropping 0 lines with matching SUB_1 and SUB_2 out of a starting total of 0
```
After, the previous behavior is restored:
```python
2119 line voltages can't be found via neighbor consensus
310 line voltages can't be found via neighbor minimum
```

### Time estimate
2 minutes.


rouille (approved):
Good catch

BainanXia (approved):
Thanks!

Issue #222, opened 2021-09-14
Parse EPA and EIA data into one table of generator data to be used in the HIFLD Grid


danielolsen:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish I could parse data from the EIA and EPA, and align them, to be able to estimate all relevant information for generators (units) or power plants (collections of units).

### Describe your proposed implementation

The EPA's AMPD dataset has hourly samples of (power, heat) for units, which can be used to fit a heat rate curve. The EPA's NEEDS dataset has linear heat-rate estimates as a fall-back. See the EPA Power Sector Modeling Platform v6 documentation for the [November 2018](https://www.epa.gov/airmarkets/documentation-epas-power-sector-modeling-platform-v6-november-2018-reference-case) and [January 2020](https://www.epa.gov/airmarkets/documentation-epas-power-sector-modeling-platform-v6-january-2020-reference-case) versions.

The EIA data has other information we want, like minimum generation level and more details on generator types (e.g. standalone CT vs. standalone ST vs. units that are part of a combined-cycle plant).

The data are not well aligned between EIA and EPA data sets, because the EIA cares about what can make electricity (generators) while the EPA cares about what can make emissions (boilers and CTs), and they both refer to them as 'units', even though the mapping is not always 1-to-1. There is a project that aims to aid in mapping: https://github.com/USEPA/camd-eia-crosswalk, if we can figure out how to use the information effectively.

There is also an open question of at which level we want to be aggregating units, especially for combined-cycle plants. As an example, the West Phoenix plant in Arizona has 3x CT/ST pairs: 1x (91 MW, 45 MW) + 2x (185 MW, 201 MW). Do we want to try to create: a) one plant, containing all generators, b) two plants, one for the smaller CC pair and one for the combination of the identical larger two, c) three plants, one for each CC pair, using each pair's fitted heat rate curve, d) three plants, one for the smaller CC pair, and two identical ones for the larger CC pairs, where we use data from both to fit a single heat rate curve, e) six individual plants, f) something else?

a) Benefits: simple. Drawbacks: overly simplified, different CC units have different characteristics.
b) Benefits: better resolution than (a). Drawbacks: we have to assume that the two larger CCs are identical based on their nameplate capacities (we could also look at things like operational date, but would have to make some assumptions about how close is close enough).
c) Benefits: better resolution than (b). Drawbacks: less data to fit curves for each of the two larger generators
[...]

PR #225, opened 2021-09-21
feat: add heat rate parameter estimation

danielolsen:

### Purpose
Add code to estimate heat rate curves from EPA AMPD data. See #222 for more context.

### What the code is doing
- Within **load.py**, we add a new function `get_eia_epa_crosswalk` to get the EIA/EPA crosswalk CSV as a dataframe with columns interpreted properly. We also modify the `get_epa_ampd` function to only use a subset of columns, to be able to cut down on the large memory requirements for loading all of the EPA data into memory.
- Within **generators.py**, we add a new primary function `estimate_heat_rate_curve`. Within this function, we check if there is an EIA/EPA crosswalk mapping available, and if there are enough non-null data points available to fit a curve to (if no to either, return NA). If so, we use the small helper function `estimate_quadratic_curve` to estimate quadratic curve parameters, with each bounded to be non-negative (based on an intuitive understanding of generator physical properties). Within `build_plant`, we prepare inputs to this primary function and call it.
- Within **setup.py**, we ensure that the files in the `prereise/gather/griddata/hifld/data` folder get copied when PreREISE is installed (unrelated fix).

### Testing
Tested manually.

### Usage Example/Visuals
```python
>>> from prereise.gather.griddata.hifld.data_process.generators import build_plant
...     build_transmission,
...     create_buses,
... )
>>> bus = create_buses(lines)
>>> generators = build_plant(bus, substations)
```

We have a lot of generators whose coefficients can't be mapped this way, but many of them we wouldn't expect to be mapped this way (based on technology), and the by-capacity coverage is 84%:
```python
>>> columns_of_interest = [
...     "Plant Code",
...     "Generator ID",
...     "Technology",
...     "lat",
...     "lon",
...     "interconnect",
...     "sub_id",
...     "bus_id",
...     "Pmax",
...     "Pmin",
...     "c0",
...     "c1",
...     "c2",
... ]
>>> generators[columns_of_interest].isna().sum()
Plant Code          0
Generator ID        0
Technology          0
lat                 0
lon                 0
interconnect        0
sub_id            383
bus_id            383
Pmax               72
Pmin             3975
c0              18076
c1              18076
c2              18076
dtype: int64
>>> no_ampd_expected_techs = {
...     "Batteries",
...     "Conventional Hydroelectric",
...     "Flywheels",
...     "Geothermal",
...     "Hydroelectric Pumped Storage",
...     "Nuclear",
...     "Offshore Wind Turbine",
...     "Onshore Wind Turbine",
...     "Solar Photovoltaic",
...     "Solar Thermal with Energy Storage",
...     "Solar Thermal without Energy Storage",
... }
>>> (
...     generators.loc[~generators["c0"].isna(), "Pmax"].sum()
...     / generators.loc[~generators["Technology"].isin(no_ampd_expected_techs), "Pmax"].sum()
... )
0.8412704125821295
```
Of the generator types with a lot of capacity missing, the type with the worst coverage (natural gas open-loop CT) still has 80% coverage by capacity, so we should be able to reasonably extrapolate to fill in these generators.

One of the discussions of fitting quadratic vs. linear heat rate curves was that the fitted curves might not have significant quadratic components, but we see a good mix of linear-only fits and meaningful quadratics. Out of about 3,700 generators, there are 1300-1400 with effectively zero quadratic component (depending on rounding precision), but the median quadratic coefficients is 3e-3, with a fairly long tail:
```python
>>> gens_with_coefficients = generators.loc[~generators["c0"].isna()]
>>> len(gens_with_coefficients)
3739
>>> (gens_with_coefficients["c2"] >= 1e-3).sum()
2209
>>> gens_with_coefficients["c2"].median()
0.0030797879396536106
>>> gens_with_coefficients["c2"].mean()
0.036075623706129496
>>> gens_with_coefficients["c2"].hist(range=(0,0.2), bins=200,cumulative=True)
<AxesSubplot:>
>>> plt.xlabel("c2")
Text(0.5, 0, 'c2')
>>> plt.ylabel("Count (cumulative)")
Text(0, 0.5, 'Count (cumulative)')
>>> plt.show()
```
c2

### Broader context
As discussed in #222, we may eventually want to aggregate some combined units together into one unit, since the CT and ST parts aren't really operationally independent. I think we should be able to refactor this new feature to do so fairly easily, by adding another column to the `generators` dataframe to represent generator groupings, and then refactoring `estimate_heat_rate_curve` to take a group of generator samples as input (summing the sample points by timestamp as applicable), rather than a single generator's samples, if/when we decide that we want to group and how the groups should be defined.

### Time estimate
30 minutes.

### DIFF
--- a/prereise/gather/griddata/hifld/data_access/load.py
+++ b/prereise/gather/griddata/hifld/data_access/load.py
@@ -7,6 +7,7 @@ from urllib.request import urlopen
 from zipfile import ZipFile
 
 import pandas as pd
+from tqdm import tqdm
 
 from prereise.gather.griddata.hifld.const import abv2state  # noqa F401
 
@@ -22,12 +23,14 @@ def get_eia_form_860(path):
     return data.query("State in @abv2state")
 
 
-def get_epa_ampd(path, year=2019):
+def get_epa_ampd(path, year=2019, cache=False):
     """Read a collection of zipped CSV files from the EPA AMPD dataset and keep readings
     from plants located in contiguous states.
 
     :param str path: path to folder. Either local or URL.
     :param str year: year of data to read (will be present in filenames)
+    :param bool cache: Whether to locally cache the EPA AMPD data, and read from the
+        cache when it's available.
     :return: (*pandas.DataFrame*) -- readings from operational power plant in contiguous
         states.
     """
@@ -43,19 +46,43 @@ def get_epa_ampd(path, year=2019):
     # Trim trailing slashes as necessary to ensure that join works
     path = path.rstrip("/\\")
 
-    data = pd.concat(
-        [
-            pd.read_csv(
-                path_sep.join(
-                    [path, f"{year}{state.lower()}{str(month_num).rjust(2, '0')}.zip"]
+    # Build cache path if necessary
+    if cache:
+        cache_dir = os.path.join(os.path.dirname(__file__), "cache")
+        os.makedirs(cache_dir, exist_ok=True)
+
+    data = {state: {} for state in abv2state}
+    for state in tqdm(abv2state):
+        for month_num in range(1, 13):
+            filename = f"{year}{state.lower()}{str(month_num).rjust(2, '0')}.zip"
+            if cache:
+                try:
+                    df = pd.read_csv(
+                        os.path.join(cache_dir, filename),
+                        usecols=heat_rate_estimation_columns,
+                    )
+                except Exception:
+                    df = pd.read_csv(
+                        path_sep.join([path, filename]),
+                        usecols=heat_rate_estimation_columns,
+                    )
+                    df.to_csv(
+                        os.path.join(cache_dir, filename),
+                        compression={
+                            "method": "zip",
+                            "archive_name": filename.replace(".zip", ".csv"),
+                        },
+                    )
+            else:
+                df = pd.read_csv(
+                    path_sep.join([path, filename]), usecols=heat_rate_estimation_columns,
                 )
-            )
-            for state in abv2state
-            for month_num in range(1, 13)
-        ]
+            data[state][month_num] = df
+    joined = pd.concat(
+        [data[state][month_num] for state in abv2state for month_num in range(1, 13)]
     )
 
-    return data
+    return joined
 
 
 def get_epa_needs(path):
--- a/prereise/gather/griddata/hifld/data_access/load.py
+++ b/prereise/gather/griddata/hifld/data_access/load.py
@@ -24,6 +24,30 @@ def get_eia_form_860(path):
     return data.query("State in @abv2state")
 
 
+def get_eia_epa_crosswalk(path):
+    """Read a CSV file mapping EIA plants IDs to EPA plant IDs, keep non-retired plants.
+
+    :param str path: path to file. Either local or URL.
+    :return: (*pandas.DataFrame*) -- filtered data frame.
+    """
+    crosswalk_match_exclude = {"CAMD Unmatched", "Manual CAMD Excluded"}  # noqa: F841
+    data = (
+        pd.read_csv(path)
+        .query(
+            "MATCH_TYPE_GEN not in @crosswalk_match_exclude and CAMD_STATUS != 'RET'"
+        )
+        .astype(
+            {
+                "MOD_EIA_PLANT_ID": int,
+                "MOD_CAMD_UNIT_ID": "string",
+                "MOD_CAMD_GENERATOR_ID": "string",
+                "MOD_EIA_GENERATOR_ID_GEN": "string",
+            }
+        )
+    )
+    return data
+
+
 def get_epa_ampd(path, years={2019}, cache=False):
     """Read a collection of zipped CSV files from the EPA AMPD dataset and keep readings
     from plants located in contiguous states.
--- a/prereise/gather/griddata/hifld/data_access/load.py
+++ b/prereise/gather/griddata/hifld/data_access/load.py
@@ -9,7 +9,8 @@ from zipfile import ZipFile
 import pandas as pd
 from tqdm import tqdm
 
-from prereise.gather.griddata.hifld.const import abv2state  # noqa F401
+from prereise.gather.griddata.hifld.const import abv2state  # noqa: F401
+from prereise.gather.griddata.hifld.const import heat_rate_estimation_columns
 
 
 def get_eia_form_860(path):
@@ -75,7 +76,8 @@ def get_epa_ampd(path, year=2019, cache=False):
                     )
             else:
                 df = pd.read_csv(
-                    path_sep.join([path, filename]), usecols=heat_rate_estimation_columns,
+                    path_sep.join([path, filename]),
+                    usecols=heat_rate_estimation_columns,
                 )
             data[state][month_num] = df
     joined = pd.concat(
--- a/prereise/gather/griddata/hifld/data_access/load.py
+++ b/prereise/gather/griddata/hifld/data_access/load.py
@@ -82,7 +82,7 @@ def get_epa_ampd(path, year=2019, cache=False):
         [data[state][month_num] for state in abv2state for month_num in range(1, 13)]
     )
 
-    return joined
+    return joined.astype({"UNITID": "string"})
 
 
 def get_epa_needs(path):
--- a/prereise/gather/griddata/hifld/data_access/load.py
+++ b/prereise/gather/griddata/hifld/data_access/load.py
@@ -24,12 +24,12 @@ def get_eia_form_860(path):
     return data.query("State in @abv2state")
 
 
-def get_epa_ampd(path, year=2019, cache=False):
+def get_epa_ampd(path, years={2019}, cache=False):
     """Read a collection of zipped CSV files from the EPA AMPD dataset and keep readings
     from plants located in contiguous states.
 
     :param str path: path to folder. Either local or URL.
-    :param str year: year of data to read (will be present in filenames)
+    :param iterable years: years of data to read (will be present in filenames).
     :param bool cache: Whether to locally cache the EPA AMPD data, and read from the
         cache when it's available.
     :return: (*pandas.DataFrame*) -- readings from operational power plant in contiguous
@@ -52,36 +52,42 @@ def get_epa_ampd(path, year=2019, cache=False):
         cache_dir = os.path.join(os.path.dirname(__file__), "cache")
         os.makedirs(cache_dir, exist_ok=True)
 
-    data = {state: {} for state in abv2state}
-    for state in tqdm(abv2state):
-        for month_num in range(1, 13):
-            filename = f"{year}{state.lower()}{str(month_num).rjust(2, '0')}.zip"
-            if cache:
-                try:
-                    df = pd.read_csv(
-                        os.path.join(cache_dir, filename),
-                        usecols=heat_rate_estimation_columns,
-                    )
-                except Exception:
+    data = {year: {state: {} for state in abv2state} for year in years}
+    for year in sorted(years):
+        for state in tqdm(abv2state):
+            for month_num in range(1, 13):
+                filename = f"{year}{state.lower()}{str(month_num).rjust(2, '0')}.zip"
+                if cache:
+                    try:
+                        df = pd.read_csv(
+                            os.path.join(cache_dir, filename),
+                            usecols=heat_rate_estimation_columns,
+                        )
+                    except Exception:
+                        df = pd.read_csv(
+                            path_sep.join([path, filename]),
+                            usecols=heat_rate_estimation_columns,
+                        )
+                        df.to_csv(
+                            os.path.join(cache_dir, filename),
+                            compression={
+                                "method": "zip",
+                                "archive_name": filename.replace(".zip", ".csv"),
+                            },
+                        )
+                else:
                     df = pd.read_csv(
                         path_sep.join([path, filename]),
                         usecols=heat_rate_estimation_columns,
                     )
-                    df.to_csv(
-                        os.path.join(cache_dir, filename),
-                        compression={
-                            "method": "zip",
-                            "archive_name": filename.replace(".zip", ".csv"),
-                        },
-                    )
-            else:
-                df = pd.read_csv(
-                    path_sep.join([path, filename]),
-                    usecols=heat_rate_estimation_columns,
-                )
-            data[state][month_num] = df
+                data[year][state][month_num] = df
     joined = pd.concat(
-        [data[state][month_num] for state in abv2state for month_num in range(1, 13)]
+        [
+            data[year][state][month_num]
+            for year in sorted(years)
+            for state in abv2state
+            for month_num in range(1, 13)
+        ]
     )
 
     return joined.astype({"UNITID": "string"})
--- a/prereise/gather/griddata/hifld/data_process/generators.py
+++ b/prereise/gather/griddata/hifld/data_process/generators.py
@@ -1,6 +1,7 @@
 import pandas as pd
 from powersimdata.utility.distance import haversine
 from scipy.optimize import curve_fit
+from scipy.stats import linregress
 
 from prereise.gather.griddata.hifld import const
 from prereise.gather.griddata.hifld.data_access import load
@@ -98,7 +99,7 @@ def estimate_heat_rate_curve(
     :param int min_unique_x: minimum number of unique x points needed to fit a curve to.
     :param int min_points: minimum number points needed to fit a curve to.
     :return: (*tuple*) -- if possible, quadratic coefficients:
-        c0 (constant), c1 (linear), and c2 (quadratic); respectively. Otherwise, return
+        h0 (constant), h1 (linear), and h2 (quadratic); respectively. Otherwise, return
         NA values for each
     """
 
@@ -179,6 +180,70 @@ def filter_suspicious_heat_rates(generators):
     ] = float("nan")
 
 
+def augment_missing_heat_rates(generators):
+    """Identify generators with NaN heat rate coefficients, replace with quadratic
+    coefficients from a linear regression against other generators of that type (for
+    fossil generators) or with assumed linear coefficients for non-fossil generators.
+
+    :param pandas.DataFrame generators: data frame of generators with fitted heat rates
+        (modified inplace).
+    """
+
+    def estimate_coefficients(generator, regressions):
+        """Estimate heat rate coefficients for a generator, by type and capacity.
+
+        :param pandas.Series generator: single generator attributes.
+        :param dict regressions: nested dictionary of fitted regressions:
+            first level is ('Technology', 'Prime Mover') attributes,
+            second level is {'h0', 'h1', 'h2'}, values are regressions against Pmax.
+        :return: (*pandas.Series*) -- heat rate parameters: (h0, h1, h2).
+        """
+        regression_key = tuple(generator[["Technology", "Prime Mover"]])
+        if regression_key not in regressions:
+            return pd.Series([float("nan")] * 3, index=["h0", "h1", "h2"])
+        type_regressions = regressions[regression_key]
+        estimates = pd.Series(
+            {
+                coef: (regression.intercept + regression.slope * generator["Pmax"])
+                for coef, regression in type_regressions.items()
+            },
+        )
+        return estimates
+
+    output_index = ["h0", "h1", "h2"]
+
+    # Augment heat rate for fossil generators
+    gens_with_coefficients = generators.loc[~generators["h0"].isnull()]
+    gen_groups = gens_with_coefficients.groupby(["Technology", "Prime Mover"])
+    regressions = {
+        technology: {
+            coef: linregress(group["Pmax"], group[coef]) for coef in output_index
+        }
+        for technology, group in gen_groups
+    }
+    gens_without_coefficients = generators.loc[generators["h0"].isnull()]
+    filled_coefficients = gens_without_coefficients.apply(
+        lambda x: (estimate_coefficients(x, regressions)), axis=1
+    )
+    generators.update(filled_coefficients)
+
+    # Augment heat rates for non-fossil generators
+    linear_heat_rate_assumptions = (
+        generators["Technology"]
+        .map(const.heat_rate_assumptions)
+        .to_frame(name="h1")
+        .apply(
+            lambda x: (
+                pd.Series([float("nan")] * 3, index=output_index)
+                if pd.isna(x["h1"])
+                else pd.Series([0, x["h1"], 0], index=output_index)
+            ),
+            axis=1,
+        )
+    )
+    generators.update(linear_heat_rate_assumptions)
+
+
 def build_plant(bus, substations, kwargs={}):
     """Use source data on generating units from EIA/EPA, along with transmission network
     data, to produce a plant data frame.
@@ -253,5 +318,6 @@ def build_plant(bus, substations, kwargs={}):
     )
     generators = generators.join(heat_rate_curve_estimates)
     filter_suspicious_heat_rates(generators)
+    augment_missing_heat_rates(generators)
 
     return generators
--- a/prereise/gather/griddata/hifld/data_process/generators.py
+++ b/prereise/gather/griddata/hifld/data_process/generators.py
@@ -130,6 +130,55 @@ def estimate_heat_rate_curve(
         return default_return
 
 
+def filter_suspicious_heat_rates(generators):
+    """Identify linearized heat rates which violate assumptions of 'reasonable' bounds,
+    and replace them with NaN values.
+
+    :param pandas.DataFrame generators: data frame of generators with fitted heat rates
+        (modified inplace).
+    """
+
+    def get_heat_rate_bounds(generator):
+        """Lookup heat rate bounds for a generator, by either type or type and capacity.
+
+        :param pandas.Series generator: single generator attributes.
+        :return: (*pandas.Series*) -- heat rate bounds ('lower', 'upper').
+        """
+        gen_type = tuple(generator[["Technology", "Prime Mover"]])
+        if gen_type in const.reasonable_heat_rates_by_type:
+            data = const.reasonable_heat_rates_by_type[gen_type]
+        key = (
+            gen_type + ("small",)
+            if generator["Pmax"]
+            < const.reasonable_heat_rates_size_cutoffs.get(gen_type, 0)
+            else gen_type + ("large",)
+        )
+        data = const.reasonable_heat_rates_by_type_and_size.get(key, (0, float("inf")))
+        return pd.Series(data, index=("lower", "upper"))
+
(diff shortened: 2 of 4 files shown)


danielolsen:
Now that #197 is merged, this is ready for a full review.

danielolsen:
There's a small fraction of the single-segment-slopes that are non-sensical, but most generators got meaningful values within a reasonable range.
```python
>>> gens_with_coefficients = generators.loc[~generators["c0"].isna()].fillna({"Pmin": 0})
>>> # Trust me on the formulation here
>>> single_segment_slope = gens_with_coefficients["c1"] + gens_with_coefficients["c2"] * (gens_with_coefficients["Pmax"] + gens_with_coefficients["Pmin"])
>>> len(single_segment_slope)
3739
>>> meaningful_slopes = single_segment_slope.loc[~single_segment_slope.isna() & (single_segment_slope != float("inf")) & (single_segment_slope != -float("inf"))]
>>> len(meaningful_slopes)
3683
>>> meaningful_slopes.hist(range=(0, 30), bins=100)
```
heat_rates
```python
>>> meaningful_slopes.max()
235.2872429237583
>>> meaningful_slopes.hist(range=(0, 30), bins=100,cumulative=True)
```
heat_rates_cumulative

There's a bit of a long tail, but it's relatively thin, and we can probably add more constraints on the fitted coefficients or clip the outputs of the fits as necessary. EPA Power Sector Modeling Platform v6 uses the following limits:
image

danielolsen:
We can separate out the components of the linearized heat rates that arise from the linear and the quadratic coefficients ('c1 component' and 'c2 component' for short) to examine where some of the anomalous values come from:
heat_rate_components
In the figure above, any values >20 have been clipped to 20 for the sake of readability.

There are a decent number of generators (~5.8%, or 2.4% of capacity) whose c1 components are effectively zero, which seems suspicious. Perhaps we need stricter filtering on the set of points required before we can try to fit a curve, e.g. a greater minimum number of points, or automatically discarding the highest and/or lowest X% of points before fitting the curve, to reduce the impact of outliers.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -96,13 +139,25 @@ def build_plant(bus, substations):
...
+    epa_ampd = load.get_epa_ampd(const.blob_paths["epa_ampd"])
+    crosswalk = load.get_eia_epa_crosswalk(const.eia_epa_crosswalk_path)
 
     # Data interpretation
     plants = plants.set_index("Plant Code")
     plants["Latitude"] = plants["Latitude"].map(floatify)
     plants["Longitude"] = plants["Longitude"].map(floatify)
     for col in ["Summer Capacity (MW)", "Winter Capacity (MW)", "Minimum Load (MW)"]:
         generators[col] = generators[col].map(floatify)
+    epa_ampd = epa_ampd.astype({"UNITID": "string"})
```
rouille, line 152: Should we move this in `get_epa_ampd` function in the `load` module?

danielolsen: Sure.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -132,5 +188,11 @@ def build_plant(bus, substations):
         ["Summer Capacity (MW)", "Winter Capacity (MW)"]
     ].max(axis=1)
     generators.rename({"Minimum Load (MW)": "Pmin"}, inplace=True, axis=1)
+    print("Fitting heat rate curves to EPA data... (this may take several minutes)")
+    heat_rate_curve_estimates = generators.apply(
+        lambda x: estimate_heat_rate_curve(x, epa_ampd_groupby, crosswalk_translation),
+        axis=1,
+    )
```
rouille, line 195: Should we have a way to use the `min_x_points` optional parameter through `build_plant` as we do with the optional `rounding` parameter in `build_transmission`

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -132,5 +191,11 @@ def build_plant(bus, substations):
...
+    generators = generators.join(heat_rate_curve_estimates)
```
rouille, line 199: To keep the user facing function slim we could move the above (what is under the **Add information** comment) into a `add_information_to_generators` function. Just a suggestion.

danielolsen:
The latest pushed version of this branch also includes a caching option for the EPA AMPD data, to avoid having to continually re-download the same hefty set of files.
```python
>>> import datetime as dt
>>> from prereise.gather.griddata.hifld import const
>>> from prereise.gather.griddata.hifld.data_access import load
>>>
>>> # Previous behavior, no caching
>>> start_time = dt.datetime.now()
>>> epa_ampd_no_cache = load.get_epa_ampd(const.blob_paths["epa_ampd"])
100%|██████████████████████████████████████████| 48/48 [02:55<00:00,  3.65s/it]
>>> print(dt.datetime.now() - start_time)
0:03:01.170025
>>> 
>>> # First time caching
>>> start_time = dt.datetime.now()
>>> epa_ampd_getting_cached = load.get_epa_ampd(const.blob_paths["epa_ampd"], cache=True)
100%|██████████████████████████████████████████| 48/48 [05:00<00:00,  6.27s/it]
>>> print(dt.datetime.now() - start_time)
0:05:07.708787
>>>
>>> # Subsequent caching
>>> start_time = dt.datetime.now()
>>> epa_ampd_from_cache = load.get_epa_ampd(const.blob_paths["epa_ampd"], cache=True)
100%|██████████████████████████████████████████| 48/48 [00:20<00:00,  2.33it/s]
>>> print(dt.datetime.now() - start_time)
0:00:26.705056
```
The cached-and-loaded files aren't identical...
```python
>>> epa_ampd_no_cache.equals(epa_ampd_from_cache)
False
```
But the differences are only in the float columns, and only due to small rounding:
```python
>>> (epa_ampd_no_cache != epa_ampd_from_cache).sum()
ORISPL_CODE                  0
UNITID                       0
GLOAD (MW)            25046059
HEAT_INPUT (mmBtu)    23736098
dtype: int64
>>> float_cols = ["GLOAD (MW)", "HEAT_INPUT (mmBtu)"]
>>> (epa_ampd_no_cache[float_cols] - epa_ampd_from_cache[float_cols]).abs().sum()
GLOAD (MW)            0.000000e+00
HEAT_INPUT (mmBtu)    2.542751e-09
dtype: float64
```

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Not a bad idea. I'd lean toward waiting until we have most of the logic built in, then we can better decide how to group and delegate.

prereise/gather/griddata/hifld/data_process/generators.py

danielolsen:
The latest version of this branch also includes the ability to set a minimum threshold number of EPA data points to use in curve-fitting (NB: zeroes are not counted). As we increase the threshold of the number of points required, we reduce the number of generators that we fit curves for, but we tighten the distribution of linearized heat rates. Interestingly, we also lower the median linearized heat rate (since we disproportionally drop the infrequently run generators, which will tend to be higher-cost?). I bet we could get broader coverage and better fits by using two years of data, rather than just one, at the expense of slower runtime.

| Cutoff | Curves (#) | Capacity (MW) | 5th pctile. | 10th | median | 90th  | 95th  |
|--------|------------|---------------|-------------|------|--------|-------|-------|
| 0      | 3696       | 667           | 0.00        | 2.66 | 9.39   | 11.16 | 12.49 |
| 24     | 3561       | 659           | 0.07        | 2.88 | 9.32   | 11.00 | 12.23 |
| 48     | 3457       | 650           | 0.48        | 2.95 | 9.21   | 10.90 | 12.15 |
| 96     | 3363       | 636           | 0.63        | 3.02 | 9.11   | 10.65 | 11.74 |
| 168    | 3221       | 622           | 1.00        | 3.12 | 9.00   | 10.47 | 11.46 |
| 336    | 3018       | 603           | 1.17        | 3.16 | 8.79   | 10.17 | 11.04 |

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -84,6 +85,51 @@ def map_generator_to_bus_by_sub(generator, bus_groupby):
...
+    eia_key = (generator["Plant Code"], generator["Generator ID"])
+    x_col = "GLOAD (MW)"
+    y_col = "HEAT_INPUT (mmBtu)"
+    try:
+        epa_key = crosswalk_translation.loc[eia_key].values[0]
+        samples = epa_ampd_groupby.get_group(epa_key)
+        filtered_samples = samples.query(
+            f"(not `{x_col}`.isnull()) and `{x_col}` != 0 "
+            f"and (not `{y_col}`.isnull()) and `{y_col}` != 0"
+        )
```
rouille, line 120: Is that shorter `query(f"{x_col} not in [0, @nan] and {y_col} not in [0, @nan]}")`?

danielolsen: When I try to run a toy example for this, I get `pandas.core.computation.ops.UndefinedVariableError: local variable 'nan' is not defined`

danielolsen: `f"{x_col} not in [0, @float('nan')]"` doesn't work either: `ValueError: "__pd_eval_local_float" is not a supported function`.

BainanXia: Not sure using `@pd.NA` instead will work or not.

danielolsen: `@pd.NA` resolves, but `float("nan")` does not evaluate to being in `[0, @pd.NA]`.

rouille: I did that:
```
>>> import pandas as pd
>>> df = pd.DataFrame({"toto": [3,4,9,10,11,float("nan"),12], "tata": [3,4,9,10,11,12,float("nan")]})
>>> nan = float("nan")
>>> df.query("toto not in [0, @nan] and tata not in [0, @nan]")
   toto  tata
0   3.0   3.0
1   4.0   4.0
2   9.0   9.0
3  10.0  10.0
4  11.0  11.0
```

danielolsen: It will work as long as all missing values are `float("nan")`, and none are pd.NA:
```python
>>> import pandas as pd
>>> df = pd.DataFrame({"a": [0, float("nan"), 2, pd.NA]})
>>> df.query("a not in [0, @pd.NA]")
     a
1  NaN
2    2
>>> df.query("a not in [0, @nan]")
      a
2     2
3  <NA>
```

rouille: and `isna()` and `isnull()` catch all of them, I presume.

BainanXia: Learning from stack overflow: `float("nan") != float("nan")` and `pd.NA != pd.NA`.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/generators.py
rouille: That is confusing

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -132,5 +193,13 @@ def build_plant(bus, substations):
         ["Summer Capacity (MW)", "Winter Capacity (MW)"]
     ].max(axis=1)
     generators.rename({"Minimum Load (MW)": "Pmin"}, inplace=True, axis=1)
+    print("Fitting heat rate curves to EPA data... (this may take several minutes)")
```
rouille, line 196: Can we use `progress_apply` in place of `apply` below to get a progress bar? See https://stackoverflow.com/questions/18603270/progress-indicator-during-pandas-operations

danielolsen: I had tried it, but couldn't get it working, since `tqdm` was trying to import something from pandas that had been deprecated/removed.

rouille: Too bad

rouille (approved):
Looks good

danielolsen:
The latest version of this branch includes several notable improvements:
- In **const.py**:
  - Assumed fuel prices are added (from EIA's Annual Energy Outlook)
  - Assumed bounds for 'reasonable' heat rates are added (drawing from the EPA's Power System Platform v6)
  - An assumed heat rate for nuclear is added (from EIA's Electric Power Annual 2019)
- A new function `filter_suspicious_heat_rates` is added, to identify fossil generator heat rates which fall outside out assumed 'reasonable' bounds, and clear them out.
- A new function `augment_missing_heat_rates` is added, which takes the 'reasonable' heat rates (after `filter_suspicious_heat_rates` is run), and for each fossil generator type, builds three regressions (one for the constant component of the heat rate, one for the linear coefficient of the heat rate, and one for the quadratic component; each regressed against Pmax), then uses these regressions to estimate parameters for generators whose heat rate curves are missing. For non-fossil types, heat rates of zero are added, or a linear heat rate curve for nuclear plants.
- Within the highest-level function, if any heat rates are missing after `filter_suspicious_heat_rates` and `augment_missing_heat_rates` (i.e. if there's a generator class with no good heat rates, since there's no good data in the EPA dataset), these generators are dropped, and details of this dropping are printed to the user. With the current code, this results in 29% of generators being dropped, but only 2.5% of total capacity, and probably an even lower share of actually generated energy.
- Assumed fuel costs are applied to each generator to translate from heat rate curves to cost curves.

Usage example:
```python
>>> from prereise.gather.griddata.hifld.data_process.generators import build_plant
...     build_transmission,
...     create_buses,
... )
>>> bus = create_buses(lines)
>>> generators = build_plant(bus, substations)
Dropping generators whose heat rates can't be estimated: 6339 out of 21743, 28751.0 MW out of 1137952.0 total MW
Technology                                   Prime Mover
All Other                                    CA               96.0
                                             OT              140.2
                                             ST             1331.6
Coal Integrated Gasification Combined Cycle  CA              618.0
                                             CT              161.0
Landfill Gas                                 CA               47.1
                                             CT              109.1
                                             FC               15.0
                                             GT              269.6
                                             IC             1425.0
                                             ST              153.4
Municipal Solid Waste                        ST             2051.5
Natural Gas Internal Combustion Engine       IC             4890.6
Natural Gas with Compressed Air Storage      CE              110.0
Other Gases                                  CA              170.0
                                             GT              737.5
                                             IC               12.0
                                             OT               36.3
                                             ST             1608.3
Other Natural Gas                            FC              224.0
                                             OT                7.0
Other Waste Biomass                          CA                7.4
                                             CT               27.0
                                             FC               10.5
                                             GT               41.3
                                             IC              205.3
                                             ST              268.6
Petroleum Coke                               CA               59.0
                                             OT               17.6
                                             ST              714.6
Petroleum Liquids                            IC             4655.6
Wood/Wood Waste Biomass                      OT                7.1
                                             ST             8524.0
Name: Pmax, dtype: float64
```

None of the generators in the output dataframe have missing values for any heat rate or cost curve parameters, and all generators with zero marginal cost are of the expected type:
```python
>>> generators.isna().sum()[["h0", "h1", "h2", "c0", "c1", "c2"]]
h0    0
h1    0
h2    0
c0    0
c1    0
c2    0
dtype: int64
>>> generators.query("c1 == 0 and c2 == 0").value_counts("Technology")
Technology
Conventional Hydroelectric              3925
Solar Photovoltaic                      3902
Onshore Wind Turbine                    1315
Geothermal                               160
Hydroelectric Pumped Storage             153
Solar Thermal without Energy Storage      16
Solar Thermal with Energy Storage          3
Offshore Wind Turbine                      1
dtype: int64
```

danlivengood (approved):

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -22,12 +24,38 @@ def get_eia_form_860(path):
...
+        pd.read_csv(path)
+        .query(
+            "MATCH_TYPE_GEN not in @crosswalk_match_exclude and CAMD_STATUS != 'RET'"
+        )
+        .astype(
+            {
+                "MOD_EIA_PLANT_ID": int,
+                "MOD_CAMD_UNIT_ID": "string",
+                "MOD_CAMD_GENERATOR_ID": "string",
+                "MOD_EIA_GENERATOR_ID_GEN": "string",
```
BainanXia, line 44: Just curious, `str` won't work here?

danielolsen: `str` results in a columns of type `object` (older-style behavior), while `"string"` results in a `StringArray` (newer-style). I think the newer is preferred since it is more explicitly text data, while `object` was the previous catch-all.

https://pandas.pydata.org/pandas-docs/stable/user_guide/text.html

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -43,19 +71,50 @@ def get_epa_ampd(path, year=2019):
     # Trim trailing slashes as necessary to ensure that join works
     path = path.rstrip("/\\")
 
-    data = pd.concat(
+    # Build cache path if necessary
+    if cache:
+        cache_dir = os.path.join(os.path.dirname(__file__), "cache")
+        os.makedirs(cache_dir, exist_ok=True)
+
+    data = {year: {state: {} for state in abv2state} for year in years}
```
BainanXia, line 79: It might be clearer to use `dict()` instead of `{}` or we could do `dict.fromkeys(years, dict.fromkeys(abv2state, dict()))` as an alternative.

danielolsen: I would prefer `{}` over `dict()`, but `dict.fromkeys` does seem nice when we have all values the same. How would you feel about `dict.fromkeys(years, dict.fromkeys(abv2state, {}))`?

danielolsen: Wait, actually I think there's a problem with `dict.fromkeys` here. If the value to be passed is mutable, it's shared among all the keys!
```python
>>> seq = ["a", "b"]
>>> d = dict.fromkeys(seq, {})
>>> d
{'a': {}, 'b': {}}
>>> d["a"][1] = 2
>>> d
{'a': {1: 2}, 'b': {1: 2}}
```

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -84,25 +86,198 @@ def map_generator_to_bus_by_sub(generator, bus_groupby):
...
+    otherwise return NA.
+
+    :param pandas.Series generator: one generating unit from data frame.
+    :param pandas.GroupBy epa_ampd_groupby: data frame of EPA AMPD samples, grouped by
+        plant ID and unit ID.
+    :param pandas.Series crosswalk_translation: mapping of EIA keys to EPA keys.
+    :param int min_unique_x: minimum number of unique x points needed to fit a curve to.
+    :param int min_points: minimum number points needed to fit a curve to.
+    :return: (*tuple*) -- if possible, quadratic coefficients:
+        c0 (constant), c1 (linear), and c2 (quadratic); respectively. Otherwise, return
```
BainanXia, line 102: Not sure whether we should replace c0, c1, c2 to h0, h1, h2, given the latter is the `return_index`?

danielolsen: You're right. Initially I had the heat rates being fit as `c0, c1, c2`, and then I changed the heat rates to `h` and the cost curves to `c`. Fixed.

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: I guess we could do:
```python
from collections import defaultdict
data = defaultdict(lambda: defaultdict(dict))
```
That also lets us do assignment three-levels deep, like we're doing with `data[year][state][month_num] = df`:
```python
>>> from collections import defaultdict
>>> d = defaultdict(lambda: defaultdict(dict))
>>> d[1][2][3] = 4
>>> d
defaultdict(<function <lambda> at 0x0000025ECD3FCF70>, {1: defaultdict(<class 'dict'>, {2: {3: 4}})})
```

It's shorted, but I think less immediately clear what it's doing, compared to the dict comprehension.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -84,25 +86,198 @@ def map_generator_to_bus_by_sub(generator, bus_groupby):
...
+        if gen_type in const.reasonable_heat_rates_by_type:
+            bounds = const.reasonable_heat_rates_by_type[gen_type]
+        elif gen_type in const.reasonable_heat_rates_size_cutoffs:
+            if generator["Pmax"] < const.reasonable_heat_rates_size_cutoffs[gen_type]:
+                key = tuple((*gen_type, "small"))
+            else:
+                key = tuple((*gen_type, "large"))
+            bounds = const.reasonable_heat_rates_by_type_and_size[key]
+        else:
+            bounds = (0, float("inf"))
```
BainanXia, line 158: Not sure whether it will be cleaner if we do following here:
```python  
gen_type = tuple(generator[["Technology", "Prime Mover"]])
size = ("small",) if generator["Pmax"] < reasonable_heat_rates_size_cutoffs.get(gen_type, 0) else ("large",)
bounds = reasonable_heat_rates_by_type.get(
    gen_type,
    reasonable_heat_rates_by_type_and_size.get(gen_type+size, (0, float("inf")))
)
```

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -1,5 +1,7 @@
 import pandas as pd
 from powersimdata.utility.distance import haversine
+from scipy.optimize import curve_fit
+from scipy.stats import linregress
 
 from prereise.gather.griddata.hifld import const
```
BainanXia, line 6: Not in this PR, should we import constants explicitly instead of the entire module?

danielolsen: We could, but it would be pretty long.
```from prereise.gather.griddata.hifld.const import (
    foo,
    bar,
    baz,
    ...
)

File: prereise/gather/griddata/hifld/data_process/generators.py (hunk 9 above)
Comment (danielolsen) [reply prereise/gather/griddata/hifld/data_process/generators.py:158]: Nice, I forgot that you can add tuples together. Line length limits sort of kill the length savings, but we can still make it simpler via your `get` approach. A lightly revised version of what you have, to fit `black`...:
```python
gen_type = tuple(generator[["Technology", "Prime Mover"]])
if gen_type in const.reasonable_heat_rates_by_type:
    bounds = const.reasonable_heat_rates_by_type[gen_type]
size = (
    "small"
    if generator["Pmax"] < reasonable_heat_rates_size_cutoffs.get(gen_type, 0)
    else "large"
)
key = gen_type + (size,)
data = const.reasonable_heat_rates_by_type_and_size.get(key, (0, float("inf")))
return pd.Series(data, index=("lower", "upper"))
```
saves us one line, or
```
gen_type = tuple(generator[["Technology", "Prime Mover"]])
if gen_type in const.reasonable_heat_rates_by_type:
    bounds = const.reasonable_heat_rates_by_type[gen_type]
key = (
    gen_type + ("small",)
    if generator["Pmax"] < reasonable_heat_rates_size_cutoffs.get(gen_type, 0)
    else gen_type + ("large",)
)
data = const.reasonable_heat_rates_by_type_and_size.get(key, (0, float("inf")))
return pd.Series(data, index=("lower", "upper"))
```
saves us two.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -84,25 +86,198 @@ def map_generator_to_bus_by_sub(generator, bus_groupby):
...
+        regression_key = tuple(generator[["Technology", "Prime Mover"]])
+        if regression_key not in regressions:
+            return pd.Series([float("nan")] * 3, index=["h0", "h1", "h2"])
+        type_regressions = regressions[regression_key]
+        estimates = pd.Series(
+            {
+                coef: (regression.intercept + regression.slope * generator["Pmax"])
+                for coef, regression in type_regressions.items()
+            },
+            index=output_index,
```
BainanXia, line 211: Do we need this, given the `coef` is the same as `output_index`?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -131,6 +308,32 @@ def build_plant(bus, substations):
...
+    augment_missing_heat_rates(generators)
+    # Drop generators whose heat rates can't be estimated
+    to_be_dropped = generators.query("h1.isnull()")
+    print(
+        f"Dropping generators whose heat rates can't be estimated: {len(to_be_dropped)}"
+        f" out of {len(generators)}, {round(to_be_dropped['Pmax'].sum(), 0)} MW out of "
+        f"{round(generators['Pmax'].sum(), 0)} total MW"
+    )
+    print(to_be_dropped.groupby(["Technology", "Prime Mover"])["Pmax"].sum())
+    generators = generators.loc[~generators.index.isin(to_be_dropped.index)].copy()
```
BainanXia, line 333: Will this be equivalent
```python
generators.drop(to_be_dropped.index, inplace=True)
```

prereise/gather/griddata/hifld/data_access/load.py
BainanXia: Aha, good to know.

prereise/gather/griddata/hifld/data_access/load.py
BainanXia: Yeah, it reminds me the trouble of having mutable defaults. I don't like the nested `defaultdict` either which can be filled with arbitrary values. Let's keep the comprehension then.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: You're right, removed.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Yes! Much cleaner.

prereise/gather/griddata/hifld/data_process/generators.py
BainanXia: Either way, no strong preference, just checking what we did before.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: I went with the second one, which in the end only saves us one line, since `reasonable_heat_rates_size_cutoffs` is a pretty long variable name, but I think is cleaner.

BainanXia (approved):
Sorry for the late review. Very nice. Hopefully, we will get more reasonable OPF results due to this effort.

PR #221, opened 2021-09-14
feat: calculate pout based on hrrr data

goccert25:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
What is the larger goal of this change?
To improve the fidelity of REISE by increasing the granularity of the wind data

### What the code is doing
How is the purpose executed?
The code calculates power for wind data based on more granular data from HRRR

### Testing
How did you test this change (unit/functional testing, manual testing, etc.)? 
Unit tests and manually running and spot checking

### Time estimate
How long will it take for reviewers and observers to understand this code change?
30 minutes to 45 minutes

PR #228, opened 2021-09-23
feat: add options to use KDTree in connecting islands by MST

BainanXia:

### Purpose
Inspired by finding closest neighbor using KDTree data structure will improve the computational performance when working with big candidate sets, discussing with @rouille and @danielolsen, there might be a chance of improving the computational time of generating the MST using the same logic. In the original implementation, during the exhaustive search of the shortest link between two islands, we did loop through the list of potential candidates each time, now we build a KDTree for the bigger island in each pair in the combination, then query each point within the small island in the tree to find the closest neighbor.

In general, KDTree should outperform over the naive implementation for big island pairs, however, for small ones, considering the time to build the tree, it underperforms. Hence, considering the long tail island size distribution we have in this project, it is unclear how much the performance improvement would be (a full test on the real dataset is running in the meantime) . 

Nonetheless, we would like to add this capability for general purpose and future use. 

### What the code is doing
- In `topology.py`:
   - A new function, `min_dist_of_2_conn_comp_kdtree`, is implemented to provide another option for `connect_islands_with_minimum_cost` on the top of `min_dist_of_2_conn_comp_kdtree`.
   - `connect_islands_with_minimum_cost` is refactored to be capable to choose which algorithm to use when calculating the minimum distance between two islands, either `naive` or `kdtree`, all corresponding keyword arguments are passing through.
- In `transmission.py`
   - Optional parameters are added to make it possible for users to specify which algorithm under the hood will be used and all corresponding parameters are passing through.
   - Add all keyword arguments into the cache key so that identically results are cached properly.
- In both modules, we add two additional parameters, `island_size_upper_bound`, `island_size_lower_bound` to allow users to filer the results by specifying the range of interested island sizes. It can be done in the upper level module (`build_transmission`) then pass through to lower level modules, however, to keep the flexibility of calling lower level module independently, this feature is implemented in both locations in a slightly different way.

### Testing
Unit tests have been added for all new functions and existing tests are extended to cover new features. ~~Integration test has been running in the mean time and the results will be updated into this post shortly.~~ A full test has been run over the default track of initial filtering as shown in the last screenshot below: **the total run time with KDTree algorithm gives around 40 mins faster comparing with the original implementation.**

Note that the weight of the MST edges calculated from the KDTree method doesn't contain realistic meanings, hence those are ignored during tests and only the selected candidates are compared, i.e. the resultant MSTs are identical. 

### Where to look
```
/prereise/gather/griddata/hifld/data_process/topology.py
/prereise/gather/griddata/hifld/data_process/transmission.py
/prereise/gather/griddata/hifld/data_process/tests/test_topology.py
```

### Usage Example/Visuals
Including screenshots of test results here:
image
image
image
image
image

### Time estimate
30 min to go through the changes, longer to understand the logic.

### DIFF
--- a/prereise/gather/griddata/hifld/data_process/topology.py
+++ b/prereise/gather/griddata/hifld/data_process/topology.py
@@ -125,7 +125,7 @@ def connect_islands_with_minimum_cost(
         defined in the ``const`` module is used.
     :param func min_dist_method: the function used to calculate minimum distance
         between two connected components, defaults to *naive* which uses
-        :py:func:`min_dist_of_2_conn_comp` or *kdtree" which uses
+        :py:func:`min_dist_of_2_conn_comp` or *kdtree* which uses
         :py:func:`min_dist_of_2_conn_comp_kdtree`.
     :param function cost_metric: a function defines the cost metric to calculate the
         weight of lines, defaults to :py:func:`haversine`.
@@ -161,7 +161,7 @@ def connect_islands_with_minimum_cost(
         raise TypeError("island_size_upper_bound must be an integer")
     if island_size_lower_bound >= island_size_upper_bound:
         raise ValueError(
-            "island_size_lower_bound must be smaller than " "island_size_upper_bound"
+            "island_size_lower_bound must be smaller than island_size_upper_bound"
         )
     if state_neighbor is None:
         state_neighbor = abv_state_neighbor
@@ -196,8 +196,8 @@ def connect_islands_with_minimum_cost(
             # components to find a line with minimum cost the connects the two islands
             if min_dist_method == "kdtree":
                 if kdtree_kwargs is None:
-                    kdtree_kwargs = {}
-                if kdtree_kwargs is not None and not isinstance(kdtree_kwargs, dict):
+                    kdtree_kwargs = dict()
+                if not isinstance(kdtree_kwargs, dict):
                     raise TypeError("kdtree_kwargs must be a dictionary")
                 min_dist, sub1, sub2 = min_dist_of_2_conn_comp_kdtree(
                     nodes1, nodes2, **kdtree_kwargs
--- a/prereise/gather/griddata/hifld/data_process/topology.py
+++ b/prereise/gather/griddata/hifld/data_process/topology.py
@@ -1,3 +1,6 @@
+import hashlib
+import os
+import pickle
 from itertools import combinations, product
 
 import networkx as nx
@@ -8,7 +11,9 @@ from tqdm import tqdm
 from prereise.gather.griddata.hifld.const import abv_state_neighbor
 
 
-def min_dist_of_2_conn_comp(nodes1, nodes2, dist_metric, memory_efficient=False):
+def min_dist_of_2_conn_comp(
+    nodes1, nodes2, dist_metric=haversine, memory_efficient=False
+):
     """Calculate the minimum distance of two connected components based on the given
     distance metric.
 
@@ -18,7 +23,8 @@ def min_dist_of_2_conn_comp(nodes1, nodes2, dist_metric, memory_efficient=False)
     :param pandas.DataFrame nodes2: a data frame contains all the nodes of the second
         connected component with index being node ID and two columns being latitude
         and longitude in order.
-    :param func dist_metric: a function defines the distance metric.
+    :param func dist_metric: a function defines the distance metric, defaults to
+        :py:func:`haversine`.
     :param bool memory_efficient: run the function in a memory efficient way or not,
         defaults to False.
     :return: (*tuple*) -- a tuple contains three elements, the minimum distance
@@ -99,13 +105,9 @@ def find_adj_state_of_2_conn_comp(cc1, cc2, state_adj, subs):
 def connect_islands_with_minimum_cost(
     lines,
     subs,
-    island_size_lower_bound=0,
-    island_size_upper_bound=None,
     state_neighbor=None,
     min_dist_method="naive",
-    cost_metric=haversine,
-    memory_efficient=False,
-    kdtree_kwargs=None,
+    **kwargs,
 ):
     """Connect a group of islands defined by lines and substations into one big
     island with minimum cost.
@@ -116,10 +118,6 @@ def connect_islands_with_minimum_cost(
     :param pandas.DataFrame subs: substation data frame indexed by ``ID`` with two
         columns ``LATITUDE`` and ``LONGITUDE`` representing the geographical
         coordination of each entry.
-    :param int island_size_lower_bound: smallest island the function should consider
-        exclusively, defaults to 0.
-    :param int island_size_upper_bound: largest island the function should consider
-        exclusively, defaults to None.
     :param dict state_neighbor: a dictionary defines the adjacency relationship among
         states, defaults to None and the constant dictionary ``abv_state_neighbor``
         defined in the ``const`` module is used.
@@ -127,12 +125,9 @@ def connect_islands_with_minimum_cost(
         between two connected components, defaults to *naive* which uses
         :py:func:`min_dist_of_2_conn_comp` or *kdtree* which uses
         :py:func:`min_dist_of_2_conn_comp_kdtree`.
-    :param function cost_metric: a function defines the cost metric to calculate the
-        weight of lines, defaults to :py:func:`haversine`.
-    :param bool memory_efficient: run the function in a memory efficient way or not,
-        defaults to False.
-    :param dict kdtree_kwargs: keyword arguments to pass to
-        :py:func:`min_dist_of_2_conn_comp_kdtree` if specified by ``min_dist_method``.
+    :param \\*\\*kwargs: keyword arguments to pass to either
+        :py:func:`min_dist_of_2_conn_comp` or :py:func:`min_dist_of_2_conn_comp_kdtree`
+        specified by ``min_dist_method``.
     :return: (*tuple*) -- a pair of lists, the first one is a list of all potential
         line candidates among given connected components; the second one is a list
         of subsequence of the first entry, representing the chosen lines to form a
@@ -140,29 +135,14 @@ def connect_islands_with_minimum_cost(
         first connected component, index of the second connected component,
         a dictionary with keys containing ``start``, ``end`` and ``weight``, defines
         the ``from substation ID``, ``to substation ID`` and the weight of the line
-        calculated by either ``cost_metric`` or KDTree based on ``kdtree_kwargs``.
-    :raises TypeError:
-        if ``island_size_lower_bound`` is not int, and/or
-        if ``island_size_upper_bound`` is not int, and/or
-        if ``kdtree_kwargs`` is specified but not a dict.
-    :raises ValueError:
-        if ``island_size_lower_bound`` is greater or equal to
-        ``island_size_upper_bound``, and/or
-        if ``min_dist_method`` is unknown.
+        calculated by either ``dist_metric`` or KDTree configurations defined in
+        ``kwargs``.
+    :raises ValueError: if ``min_dist_method`` is unknown.
 
         .. note:: the indexes of connected components are defined by the size,
-            i.e. number of nodes, of the connected components in ascending order.
+            i.e. number of nodes, of the connected components in descending order.
     """
-    if island_size_upper_bound is None:
-        island_size_upper_bound = len(subs) + 1
-    if not isinstance(island_size_lower_bound, int):
-        raise TypeError("island_size_lower_bound must be an integer")
-    if not isinstance(island_size_upper_bound, int):
-        raise TypeError("island_size_upper_bound must be an integer")
-    if island_size_lower_bound >= island_size_upper_bound:
-        raise ValueError(
-            "island_size_lower_bound must be smaller than island_size_upper_bound"
-        )
+
     if state_neighbor is None:
         state_neighbor = abv_state_neighbor
 
@@ -171,14 +151,7 @@ def connect_islands_with_minimum_cost(
 
     # Get the full list of connected components of the resultant graph sored by size in
     # ascending order
-    all_cc = sorted(
-        [
-            cc
-            for cc in list(nx.connected_components(graph))
-            if island_size_lower_bound < len(cc) < island_size_upper_bound
-        ],
-        key=len,
-    )
+    all_cc = sorted(list(nx.connected_components(graph)), key=len, reverse=True)
 
     # Loop through all the combinations of connected components to identify potential
     # edges filtered by the adjacency relationship specified by the user
@@ -195,17 +168,11 @@ def connect_islands_with_minimum_cost(
             # Run an exhaustive search on the filtered substations of both connected
             # components to find a line with minimum cost the connects the two islands
             if min_dist_method == "kdtree":
-                if kdtree_kwargs is None:
-                    kdtree_kwargs = dict()
-                if not isinstance(kdtree_kwargs, dict):
-                    raise TypeError("kdtree_kwargs must be a dictionary")
                 min_dist, sub1, sub2 = min_dist_of_2_conn_comp_kdtree(
-                    nodes1, nodes2, **kdtree_kwargs
+                    nodes1, nodes2, **kwargs
                 )
             elif min_dist_method == "naive":
-                min_dist, sub1, sub2 = min_dist_of_2_conn_comp(
-                    nodes1, nodes2, cost_metric, memory_efficient=memory_efficient
-                )
+                min_dist, sub1, sub2 = min_dist_of_2_conn_comp(nodes1, nodes2, **kwargs)
             else:
                 raise ValueError("min_dist_method must be either `naive` or `kdtree`")
             edge_list.append(
@@ -216,3 +183,37 @@ def connect_islands_with_minimum_cost(
     cc_graph.add_edges_from(edge_list)
     cc_mst = nx.minimum_spanning_tree(cc_graph)
     return edge_list, sorted(cc_mst.edges(data=True))
+
+
+def get_mst_edges(lines, substations, **kwargs):
+    """Get the set of lines which connected the connected components of the lines graph,
+    either from a cache or by generating from scratch and caching.
+
+    :param pandas.DataFrame lines: data frame of lines.
+    :param pandas.DataFrame substations: data frame of substations.
+    :param \\*\\*kwargs: optional arguments for
+        :py:func:`connect_islands_with_minimum_cost`.
+    :return: (*list*) -- each entry is a 3-tuple:
+        index of the first connected component (int),
+        index of the second connected component (int),
+        a dictionary with keys containing ``start``, ``end`` and ``weight``, defines
+        the ``from substation ID``, ``to substation ID`` and the distance of the line.
+    """
+    cache_dir = os.path.join(os.path.dirname(__file__), "cache")
+    os.makedirs(cache_dir, exist_ok=True)
+    cache_key = (
+        repr(lines[["SUB_1_ID", "SUB_2_ID"]].to_numpy().tolist())
+        + repr(substations[["LATITUDE", "LONGITUDE"]].to_numpy().tolist())
+        + repr(kwargs)
+    )
+    cache_hash = hashlib.md5(cache_key.encode("utf-8")).hexdigest()
+    try:
+        with open(os.path.join(cache_dir, f"mst_{cache_hash}.pkl"), "rb") as f:
+            print("Reading cached minimum spanning tree")
+            mst_edges = pickle.load(f)
+    except Exception:
+        print("No minimum spanning tree available, generating...")
+        _, mst_edges = connect_islands_with_minimum_cost(lines, substations, **kwargs)
+        with open(os.path.join(cache_dir, f"mst_{cache_hash}.pkl"), "wb") as f:
+            pickle.dump(mst_edges, f)
+    return mst_edges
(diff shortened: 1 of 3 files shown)


prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -40,6 +41,40 @@ def min_dist_of_2_conn_comp(nodes1, nodes2, dist_metric, memory_efficient=False)
...
+    tree = KDTree([ll2uv(p[1], p[0]) for p in nodes1.values])
+    shortest_link = nodes2.apply(
+        lambda x: tree.query(ll2uv(*x[::-1]), **kwargs), axis=1
+    )
+    min_dist, sub1_ind = shortest_link.min()
+    sub1 = nodes1.index[sub1_ind]
+    sub2 = shortest_link.loc[shortest_link == (min_dist, sub1_ind)].index[0]
+    return (
+        (min_dist, sub2, sub1) if len(nodes1) < len(nodes2) else (min_dist, sub1, sub2)
+    )
```
danielolsen, line 75: After this block...
```python
if len(nodes1) < len(nodes2):
     nodes1, nodes2 = nodes2, nodes1
```
...wont `len(nodes1)` always be >= `len(nodes2)`? In the current test, the length of the test connected components are equal, so that first block doesn't kick in either way and the output indices are properly reversed if we reverse the inputs, but if we add another entry to one of the connected components, then the function returns identical outputs even if the ordering of the inputs is switched.

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -73,25 +116,53 @@ def connect_islands_with_minimum_cost(
     :param pandas.DataFrame subs: substation data frame indexed by ``ID`` with two
         columns ``LATITUDE`` and ``LONGITUDE`` representing the geographical
         coordination of each entry.
-    :param function cost_metric: a function defines the cost metric to calculate the
-        weight of lines, defaults to :py:func:`haversine`.
+    :param int island_size_lower_bound: smallest island the function should consider
+        exclusively, defaults to 0.
+    :param int island_size_upper_bound: largest island the function should consider
+        exclusively, defaults to None.
```
danielolsen, line 122: Does a default of `None` mean that all islands are considered?

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -73,25 +116,53 @@ def connect_islands_with_minimum_cost(
...
     :param dict state_neighbor: a dictionary defines the adjacency relationship among
         states, defaults to None and the constant dictionary ``abv_state_neighbor``
         defined in the ``const`` module is used.
+    :param func min_dist_method: the function used to calculate minimum distance
+        between two connected components, defaults to *naive* which uses
+        :py:func:`min_dist_of_2_conn_comp` or *kdtree" which uses
```
danielolsen, line 128: `*kdtree"` -> `*kdtree*`?

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -73,25 +116,53 @@ def connect_islands_with_minimum_cost(
...
+    :raises ValueError:
+        if ``island_size_lower_bound`` is greater or equal to
+        ``island_size_upper_bound``, and/or
+        if ``min_dist_method`` is unknown.
 
         .. note:: the indexes of connected components are defined by the size,
             i.e. number of nodes, of the connected components in ascending order.
     """
+    if island_size_upper_bound is None:
+        island_size_upper_bound = len(subs) + 1
```
danielolsen, line 157: Would `float("inf")` be clearer than `len(subs) + 1`?

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -73,25 +116,53 @@ def connect_islands_with_minimum_cost(
...
+    if not isinstance(island_size_lower_bound, int):
+        raise TypeError("island_size_lower_bound must be an integer")
+    if not isinstance(island_size_upper_bound, int):
+        raise TypeError("island_size_upper_bound must be an integer")
+    if island_size_lower_bound >= island_size_upper_bound:
+        raise ValueError(
+            "island_size_lower_bound must be smaller than " "island_size_upper_bound"
```
danielolsen, line 164: Weird string formatting, maybe a `black` artifact?

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -116,9 +194,20 @@ def connect_islands_with_minimum_cost(
             nodes2 = subs.loc[cc2].query("STATE in @cc2_adj")[["LATITUDE", "LONGITUDE"]]
             # Run an exhaustive search on the filtered substations of both connected
             # components to find a line with minimum cost the connects the two islands
-            min_dist, sub1, sub2 = min_dist_of_2_conn_comp(
-                nodes1, nodes2, cost_metric, memory_efficient=memory_efficient
-            )
+            if min_dist_method == "kdtree":
+                if kdtree_kwargs is None:
+                    kdtree_kwargs = {}
+                if kdtree_kwargs is not None and not isinstance(kdtree_kwargs, dict):
```
danielolsen, line 200: At this point, we know that `kdtree_kwargs is not None`, because of the block immediately before, right?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -588,6 +592,28 @@ def build_transmission(method="sub2line", kwargs={"rounding": 3}):
...
+        if keyword in kwargs:
+            get_mst_kwargs[keyword] = kwargs.pop(keyword)
+        if "island_size_lower_bound" in get_mst_kwargs:
+            island_size_lower_bound = get_mst_kwargs["island_size_lower_bound"]
+        else:
+            island_size_lower_bound = 0
+        if "island_size_upper_bound" in get_mst_kwargs:
+            island_size_upper_bound = get_mst_kwargs["island_size_upper_bound"]
+        else:
+            island_size_upper_bound = len(substations) + 1
```
danielolsen, line 615: I don't really understand the point of this section. Isn't the point of the current `build_transmission` `kwargs` parameter that it can take any combination of keyword arguments for different lower-level functions, and each function can be passed these combined keyword arguments and only care about the ones that are relevant to it, ignoring the rest?

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Yeah good catch, this is a bug. The intention here is creating the tree using the big island and looping through the nodes in the small island to query to tree. So we would like `nodes1` always represents the big island. Hence, in the situation we swap, we will need to swap the return value to keep output consistent.

BainanXia: Fixed in [d03295b](https://github.com/Breakthrough-Energy/PreREISE/pull/228/commits/d03294b40c6c1aee36b1e259296dbaec0364e8c9) and one more test to cover this track is added [65a2557](https://github.com/Breakthrough-Energy/PreREISE/pull/228/commits/65a2557ae356d27522379c83309c75f5f4102192).

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Yes, as long as i`sland_size_lower_bound == 0`.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Actually, I thought about it. If we use `float("inf")` instead of `None`, it's a little confusing to give an `int` type variable a `float` default value, also, we need to take care of it during the input checking in l.160-l.161 of `topology.py`.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Previously it doesn't fit into one line, then I reworded it, however, forgot to delete the auto added quotation mark. Fixed.

prereise/gather/griddata/hifld/data_process/topology.py
danielolsen: Good point.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Done.

prereise/gather/griddata/hifld/data_process/topology.py
BainanXia: Done.

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: Yeah, I agree with the logic. However, we still need to filter out the corresponding kwargs for specific lower level functions, right? Otherwise, it will throw errors like `TypeError: connect_islands_with_minimum_cost() got an unexpected keyword argument 'rounding'.`

danielolsen: I think we want to have the function definitions to have `**kwargs` all the way down to prevent this, i.e. `def connect_islands_with_minimum_cost(lines, subs, **kwargs)`.

BainanXia: I think this logic looks good for passing `kwargs` through, however, when some `kwargs` hit the lowest level function, we should have them explicitly defined.

danielolsen: I don't think that follows the current pattern in **transmission.py** though, since `map_lines_to_substations_using_coords` is a lowest-level function and still has `**kwargs`.

BainanXia: I'm saying it follows the pattern in other packages such as `pandas`. 

Then we should refactor `map_lines_to_substations_using_coords` some day in the future once we have a clear idea on what `kwargs` it takes.

danielolsen: Maybe I'm unclear on the design for the `build_transmission` then. Are we going to take `**kwargs` at the highest-level, and then separate this one dict into several lower-level dicts? If so, I think we could do this purely on keys, and then have a pattern of `dict.get(key, default value)` in the lower- (lowest-?) level functions.

BainanXia: That's another option, then how do we write doc strings? I think there should be a classic way of handling `kwargs` and we can simply follow that. The current implementation is my 2 cents after looking through pandas documentation a little bit. 

Maybe @jon-hagg has a better idea on this?
The issue we encounter here is, given a top level user facing function, which takes a set of `kwargs`, however, different subsets of `kwargs` should be passed to different middle/lower level functions. What would be the classic pattern of function definitions for top level function, middle level functions, and lower level functions to handle this situdation?

BainanXia: Btw, the two `if`s setting defaults for island size bounds here is on a different track, given they are also used in `transmission.py` as well.

danielolsen: You raise a good point about the docstrings.

I think one approach would be to define a dictionary `{str: set of str}` within build_transmission that dictates which of the kwargs go to which lower-level functions, then decompose the highest-level `**kwargs` to lower-level ones (making sure that there are no leftover keys) and pass those as appropriate... but it gets confusing when some of those highest-level kwargs are going to be used in the highest-level function as well. What do you think about moving everything from `mst_edges = get_mst_edges(...)` to `lines = pd.concat([lines, new_lines])` to a lower-level function? E.g. `lines, substations = filter_connected_components_and_connect(lines, substations, **specific_kwargs)`?

BainanXia: That's another work around which is cleaner. However, given it is a general implementation related question, I presume there should be a classic way to handle this. Let's put it on hold waiting for ideas from people with related experiences for now.

BainanXia: Regarding the refactor idea, I like it. The down side is the MST feature is tied up with this framework, which makes it less flexible to use it elsewhere and it also closes the window of further processing on the selected edges in the main function as well.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -588,6 +592,28 @@ def build_transmission(method="sub2line", kwargs={"rounding": 3}):
...
+    for keyword in {
+        "island_size_lower_bound",
+        "island_size_upper_bound",
+        "state_neighbor",
+        "min_dist_method",
+        "cost_metric",
+        "memory_efficient",
+        "kdtree_kwargs",
+    }:
+        if keyword in kwargs:
```
jon-hagg, line 606: If we stick with the kwargs pattern, I would change the signature to be the following, since the dict is mutable, and we are also modifying it here.

```
def build_transmission(method="sub2line", **kwargs):
 ```

File: prereise/gather/griddata/hifld/data_process/transmission.py (hunk 7 above)
Comment (jon-hagg) [reply prereise/gather/griddata/hifld/data_process/transmission.py:615]: Haven't fully thought this through, but one way to handle the various options is using classes, and constructing the top level instance based on the mst algorithm. Part of this could include also using a class to define/validate the kwargs that would apply to the algorithm. Something like

```
class Transmission:
    mst_builder: MstBuilder

    def __init__(self, mst_algo="naive", mst_options=None)
        # set defaults or validate the combination of options/algo

    def build_transmission(self, method):
        # use the mst_builder instance

class NaiveMst:
    options: NaiveMstOptions

    def build(self):
        # do stuff

class KDTreeMst:
    options: KDTreeMstOptions

    def build(self):
        # do stuff

class KDTreeMstOptions:
    def __init__(self, island_size_lower_bound, island_size_upper_bound, other_params):
        # set defaults or validate param types/values
        
    def build(self):
        # do stuff

builder = Transmission(KDTreeMst(KDTreeMstOptions()))


```

BainanXia: > Haven't fully thought this through, but one way to handle the various options is using classes, and constructing the top level instance based on the mst algorithm. Part of this could include also using a class to define/validate the kwargs that would apply to the algorithm. Something like
> 
> ```
> class Transmission:
>     mst_builder: MstBuilder
> 
>     def __init__(self, mst_algo="naive", mst_options=None)
>         # set defaults or validate the combination of options/algo
> 
>     def build_transmission(self, method):
>         # use the mst_builder instance
> 
> class NaiveMst:
>     options: NaiveMstOptions
> 
>     def build(self):
>         # do stuff
> 
> class KDTreeMst:
>     options: KDTreeMstOptions
> 
>     def build(self):
>         # do stuff
> 
> class KDTreeMstOptions:
>     def __init__(self, island_size_lower_bound, island_size_upper_bound, other_params):
>         # set defaults or validate param types/values
>         
>     def build(self):
>         # do stuff
> 
> builder = Transmission(KDTreeMst(KDTreeMstOptions()))
> ```

Not sure I understand what we are trying to solve by changing everything into class structure. In this situation, we don't have specific data structures with different set of functions associated with it. It fits well with a sequence of functions calls to process the data in a proposed way. Also, I don't see how class instance might help passing `kwargs` through a nested function call structure assuming we still have one user facing function.

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: Maybe we need some background story from @rouille or @danielolsen on the current implementation?

BainanXia: @jon-hagg If I'm understanding correctly, the idea of the design would be in function definition, any `kwarg` hits the lowest level function, it should be declared explicitly and everything else can be collapsed in `**kwargs`? Suppose we have the main user facing function `funcA(kwarg_a, **kwargs)` and it calls `funcB` and `funcC`, which take different entries in `kwargs` as keyword arguments, how do will filter out those arguments during the function call so that it won't throw `TypeError: funcB() got an unexpected keyword argument 'xxx'`?

jon-hagg: I guess there are two ways I see to handle it. The current way pops off each argument before passing to lower level functions with explicit keyword arguments defined. Another way would be having lower level functions just take arbitrary kwargs (in your example, it would be `funcB(**kwargs)` and same for `funcC`), and pop the ones they are interested in. 

My original comment was mainly about the mutable default argument, which is a pattern we avoid in general. I don't have a strong opinion on whether to declare it as `def foo(kwargs=None)` or `def foo(**kwargs)`.

danielolsen: There's some backstory here: https://github.com/Breakthrough-Energy/PreREISE/pull/217#discussion_r705552095

The main design goal I think was to enable the user to have a lot of flexibility from the highest-level `build_transmission` function, to be able to specify parameters for several lower-level functions, without cluttering the `build_transmission` docstring too bad. We do something similar in PostREISE for the map functions that can pass keyword arguments down to the lower-level function that builds the background states map.

If we stick with one big dict in the top-level function, then it seems the design candidates for how to handle it are:
a. Pass all kwargs to all lower-level functions. Pro: simplicity of code. Con: it's less clear which keyword arguments will be interpreted by each lower-level function.
b. Within the highest-level function, split the one big dictionary into smaller dictionaries, each of which is passed to its corresponding lower-level function. Pro: clarity of lower-level function docstrings. Con: the higher-level function gets a bit more complex.

prereise/gather/griddata/hifld/data_process/transmission.py
jon-hagg: Whether or not classes are the way to go here, I would say in general they _may_ be useful regardless of if there is a conventional data structure to implement. E.g. having an "options" class could store a group of related kwargs and validate them upon creation, keeping that logic separate from the rest of the function (which makes it testable independently, if applicable), and clearly shows what the grouping is.

In this case, my intuition is to use classes to represent the polymorphic aspect (naive vs kdtree, or various sets of kwargs). Constructing the object up front allows internal calls to operate within a larger scope, without having to have a bunch of pass-through parameters. It essentially moves some if/else logic to the top layer, rather than having it embedded in lower level functions. So not a fundamental reduction in complexity, but might improve readability and ease of development. 

Feel free to use any/all/none of the class approach. If we still want the user interface level to be function based, it could still use classes internally, or not, and construct the appropriate instances based on the function arguments.

BainanXia: If we keep the user interface level to be function based, but go for classes internally, we still need the user facing function to pass through the `kwargs` upon internal class instance creation, right?

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: The thing is, `KDTree` itself is not implemented by us, i.e. we can't pass arbitrary `kwargs` to it at the lower level function that creates the `KDTree`. Instead, we still need to filter out what `kwargs` should go into `KDTree` creation and `KDTree.query` at the end, right? Filtering at the top level function will give us a clean doc string structure for all lower level functions.

In any case, we will need convergence in the options we have been discussing on this track:
1. The current implementation: the user facing function takes `**kwargs`, then we filter which `kwargs` to pass to lower level functions before the function call. Lower level functions will have `kwargs` defined explicitly.
2. Use one big dict `kwargs` in the top-level function. Pass all `kwargs` to all lower level functions. All lower level functions will take `**kwargs`.
3. Use several smaller dicts in the top-level function. Pass specific dict to lower level functions.
4. Top-level function is the same as 2. But use classes internally, pass `kwargs` upon class instance creation.
5. All of class approach.

prereise/gather/griddata/hifld/data_process/transmission.py
jon-hagg: I think so, yes. Not sure if it's applicable here, but for example if we had like 10 layers of nested function calls, creating a class instance (specifically, the object graph) up front would save on a bunch of redundant parameters, and only require parsing kwargs into objects once. Since we have like 2-3 layers, the trade off is not so obvious.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I like option 1 the best, and option 3 second-best.

rouille: I like 1 the best

BainanXia: Done with [77d6ce7](https://github.com/Breakthrough-Energy/PreREISE/pull/228/commits/77d6ce7c94e54acced97d04da4d83870683c32da)
I did a bunch of refactor to make the `kwargs` passing logic consistent throughout the module based on option 1:
- A new function `filter_islands_and_connect_with_mst` is added following @danielolsen 's suggestion in this [thread](https://github.com/Breakthrough-Energy/PreREISE/pull/228#discussion_r718054690), which keeps `build_transmission` as purely a collection of calls to low/medium-level functions, with no lowest-level logic.
- `get_mst_edges` is moved to `topology` module.
- According to option 1, top level user facing function `build_transmission` only takes `**kwargs`, which is further split into two subsets feeding into `map_lines_to_substations_using_coords` and `filter_islands_and_connect_with_mst` respectively. The former function is refactored accordingly to reflect the change.

danielolsen (approved):
The code looks good, and the tests indeed still pass in the Docker container. After #221, native Windows installation/testing is difficult, since `pygrib` relies on some non-python dependencies which don't seem to be easily installable on Windows.

Issue #233, opened 2021-10-13
Improve handling of DC lines when loading HIFLD data


danielolsen:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish when we loaded the HIFLD lines data (i.e. `prereise.gather.griddata.hifld.data_access.load.get_hifld_electric_power_transmission_lines`):
- We filtered out DC lines based on the `TYPE` columns of the data frame (i.e. explicitly remove `"DC; OVERHEAD"` and `"DC; UNDERGROUND"`) to eventually become part of the `Grid.dcline` table instead of `Grid.branch`.
- We made sure that all real-world DC lines got added somehow. Filtering just on {`"DC; OVERHEAD"`, `"DC; UNDERGROUND"`} and looking at the substation names, I can identify:
  - the Pacific DC Intertie (Path 65): ID 200823 from `CELILO` to `NOT AVAILABLE`
  - the Intermountain Power Project DC Line (Path 27): ID 308464 from `INTERMOUNTAIN` to `ADELANTO`
  -  the TransBay Cable: ID 310053 from `PITTSBURG` to `TRANS BAY CABLE FACILITY`
  - Square Butte: ID 108354 from `ARROWHEAD` to `UNKNOWN133416`
  - CU: ID 150123 from `DICKINSON` to `UNDERWOOD`
- Looking at the map, I was able to find:
  - The Cross-Sound Cable: ID 157627 from `NEW HAVEN HARBOR` to `SHOREHAM POWER STATION` (ambiguously classified as `UNDERGROUND`)
  - The Hudson Project: ID 157629 from `HVDC B2B STATION HUDSON` to `WEST 49TH STREET` (accurately classified as `AC; UNDERGROUND`, since it connects to a B2B facility)
  - The Neptune Cable: ID 113313 from `SAYREVILLE` to `DUFFY AVENUE CONVERTER STATION` (erroneously classified as `AC; UNDERGROUND`)

I also see the Quebec/New England Transmission line, in two sections (ID 158515 from `NOT AVAILABLE` [Canada] to `MONROE` and ID 131914 from `UNKNOWN133682` [Ayer, MA] to `MONROE`).

~I don't see the Neptune cable, the Cross-Sound cable, or the Hudson Project, although they might be part of the more general {`"OVERHEAD"`, `"UNDERGROUND"`, or `"NOT AVAILABLE"`} type lines (16% of the total).~

[...]

Issue #226, opened 2021-09-21
Add single function that kicks off all HIFLD Grid building


danielolsen:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish there was a single user-facing function that could orchestrate all lower-level functions to build HIFLD grid data.

### Describe your proposed implementation
The functions inputs should only be pass-throughs to lower-level functions, e.g. the higher-level functions within `prereise.gather.griddata.hiflddata.data_process`.

danielolsen:
Closed via #236.

Issue #234, opened 2021-10-28
Improve transmission coverage of HIFLD dataset when source data are missing


danielolsen:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish we had a way to add transmission lines to areas where there it is known to be missing.

### Describe your proposed implementation

One relatively simple way to achieve this could be a user-supplied list of substation coordinates and transmission lines connecting them, which would get appended to the original HIFLD data before further processing. An algorithm to create synthetic information would be nice, but much more difficult.

### Additional context
See New York City:
![image](https://user-images.githubusercontent.com/7348392/139350733-d6a31fcb-6311-4ece-87e8-a287b6ef33b4.png)

danielolsen:
New York City is especially tough because so much of the electrical network is underground, so the location of substation and any transmission lines that connect them is unknown. On the data side, we may be able to use subway stations as a proxy, since their locations are more well-known and station density should roughly correlate with density of electrical consumption.

danielolsen:
See also: Chicago
![image](https://user-images.githubusercontent.com/7348392/156674999-5f89b07b-b056-4429-958d-61ad3554b1ab.png)

danielolsen:
Currently within #267, we add the ability to add arbitrary substations to the source data. In the same way, I believe we can add arbitrary transmission lines, once we determine which substations and transmission lines we can use as a proxy.

danielolsen:
Interestingly, there's more coverage of **substations** within these areas as compared to transmission lines. Even before adding proxy substations, we should be able to add proxy lines that connect to these real substations to distribute the cities' demand a bit better.
![image](https://user-images.githubusercontent.com/7348392/160169755-b05ba538-7117-4bb3-a843-f3d458310b76.png)
![image](https://user-images.githubusercontent.com/7348392/160170117-903c2cbc-c7f7-46f5-a287-d10cf8efbb5c.png)

BainanXia:
@danielolsen Does it imply the real lines connecting those real substations are actually underground?

danielolsen:
Probably. The HIFLD transmission line dataset has some underground lines, but the coverage seems much spottier than the overhead ones, either due to data availability or some strategic decision about which data should be shared (or something else entirely).

danielolsen:
Another fun one from looking at Texas:
[...]

PR #235, opened 2021-10-29
feat: add function to assign demand to buses proportional to population

### LINKED ISSUE #229: Build distribution of demand to HIFLD buses
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish there was a function that populated the `Pd` field for the bus table for the HIFLD grid.

### Describe your proposed implementation
The new function should either live in `prereise.gather.griddata.hifld.data_process.transmission` or in a new module in a similar location (`prereise.gather.griddata.hifld.data_process.demand`?). One route could be demand proportional to population, which could be obtained from the census or similar sources.

danielolsen:

### Purpose
Estimates population per substation, using data from simplemaps.com, and uses this information to add demand to buses. This is a re-implementation of the approach taken by the collaborators. Closes #229. 

### What the code is doing

We add a new module `prereise.gather.griddata.hifld.data_process.demand` with a single function `assign_demand_to_buses`. This function:
1. Loads population by ZIP and by county from CSVs in the repo.
2. Sorts substations by connected transmission capacity.
3. Distributes each ZIP code's population to the N highest-transmission-capacity non-generation substations within that ZIP, where N is an integer value less than the total (calculated using the `substation_load_share` fraction within **const.py**) but no less than one, unless there are no substations within that ZIP. These are considered as 'load substations'.
4. Determines which counties have no substations with ZIP-assigned demand ('load substations'), and for each of these, picks the substation in the county with greatest transmission capacity to add to the set of load substations.
5. Determines how much of each county's population is not yet assigned, and distributes this to the load substations within the county. Note: if a county has no substations at all, its population does not go anywhere else, and is effectively ignored. Less than 5% of population is ignored this way, so it should not have a large effect on the overall demand distribution.
6. Translates total population (from ZIP-assignment and county assignment) to demand using an assumption of demand-per-person.
7. Selects the lowest-voltage bus within each substation to assign demand to.

Besides **demand.py**, all other changes are data/documentation.

### Testing
Tested manually.

### Usage Example/Visuals
```python
import pandas as pd
from prereise.gather.griddata.hifld.data_process.demand import assign_demand_to_buses
from prereise.gather.griddata.hifld.data_process.generators import build_plant
from prereise.gather.griddata.hifld.data_process.transmission import (
    build_transmission,
    calculate_branch_mileage,
    create_buses,
    create_transformers,
    estimate_branch_impedance,
    estimate_branch_rating,
)

# Invoking highest-level `data_process` functions
lines, substations = build_transmission(method="line2sub")
bus = create_buses(lines)
generators = build_plant(bus, substations)

# This code has been demonstrated via other PRs, but hasn't been baked into the top-level data process functions yet
lines["type"] = "Line"
lines["length"] = lines.apply(calculate_branch_mileage, axis=1)
transformers = create_transformers(bus)
transformers["type"] = "Transformer"
branch = pd.concat([lines, transformers])
branch["x"] = branch.apply(lambda x: estimate_branch_impedance(x, bus["baseKV"]), axis=1)
branch["rateA"] = branch.apply(lambda x: estimate_branch_rating(x, bus["baseKV"]), axis=1)

# New code
assign_demand_to_buses(substations, branch, generators, bus)
```

A `"Pd"` column is added inplace to the `bus` dataframe:
```python
>>> # Estimated 2.01 kW per person within code, 315.8 million people assigned to buses
>>> bus["Pd"].sum() / 2.01e-3
315800868.4704517
>>> bus["Pd"].isna().sum()
0
```

We need the generators to be able to preferentially assign demand to non-generator buses, and we need branch capacities to preferentially assign demand to higher-capacity substations when multiple substations are available within an area (ZIP or county). Generating the transformers and estimating branch impedances and capacities should probably be added to `build_transmission` as part of fulfillment of #226.

I demonstrate using the "line2sub" method since I have more faith in line coordinates than line substation names, based on some exploration with DC lines: https://github.com/Breakthrough-Energy/PreREISE/issues/233#issuecomment-954154337. I suggest that we also switch to this as default as part of #226.

### Time estimate
1 hour. The code itself isn't too long, but it's some fairly dense pandas.

### DIFF
--- a/ATTRIBUTION.md
+++ b/ATTRIBUTION.md
@@ -447,4 +447,36 @@ The data are used to populate the power generation profile of U.S. transmission 
 ##### Note
 Public use (see https://hifld-geoplatform.opendata.arcgis.com/datasets/geoplatform::power-plants/about). Users are advised to read the data set's metadata thoroughly to understand appropriate use and data limitations.
 
+
+##### Source
+* Name: U.S. Zips
+* Author: Pareto Software, LLC
+* Description: Data on United States ZIP codes
+* Source: https://simplemaps.com
+* Exact source location: https://simplemaps.com/data/us-zips
+* Note: version 1.78, accessed 2021-10-27.
+
+##### Destination
+* Modifications to source file(s): None
+* Location: https://besciences.blob.core.windows.net/datasets/geo_data/uszips.csv
+
+##### General Purpose
+The data are used to estimate population served by each substation, to distribute demand.
+
+
+##### Source
+* Name: U.S. Counties
+* Author: Pareto Software, LLC
+* Description: Data on United States counties
+* Source: https://simplemaps.com
+* Exact source location: https://simplemaps.com/data/us-counties
+* Note: version 1.71, accessed 2021-10-27.
+
+##### Destination
+* Modifications to source file(s): None
+* Location: https://besciences.blob.core.windows.net/datasets/geo_data/uscounties.csv
+
+##### General Purpose
+The data are used to estimate population served by each substation, to distribute demand.
+
 ---
--- a/prereise/gather/griddata/hifld/data_process/demand.py
+++ b/prereise/gather/griddata/hifld/data_process/demand.py
@@ -1,0 +1,76 @@
+import pandas as pd
+
+from prereise.gather.griddata.hifld import const
+from prereise.gather.griddata.hifld.data_access.load import get_us_counties, get_us_zips
+
+
+def assign_demand_to_buses(substations, branch, plant, bus):
+    """Using data on population by county and ZIP code, assign demand to substations,
+    then to the lowest-voltage bus within each substation.
+    This demand parameter is added inplace as a 'Pd' column to the ``bus`` data frame.
+
+    :param pandas.DataFrame substations: table of substation data.
+    :param pandas.DataFrame branch: table of branch data.
+    :param pandas.DataFrame plant: table of plant data.
+    :param pandas.DataFrame bus: table of bus data.
+    """
+    # Load data
+    zip_data = get_us_zips(const.blob_paths["us_zips"])
+    county_data = get_us_counties(const.blob_paths["us_counties"])
+
+    # Determine each substation's transmission capacity, then sort for selection
+    filtered_branch = branch.query("SUB_1_ID != SUB_2_ID")
+    from_cap = filtered_branch.groupby("SUB_1_ID").sum()["rateA"]
+    to_cap = filtered_branch.groupby("SUB_2_ID").sum()["rateA"]
+    sub_cap = from_cap.combine(to_cap, lambda x, y: x + y, fill_value=0)
+    # Sort substations by their capacities for later ordered selection
+    sorted_subs = substations.loc[sub_cap.sort_values(ascending=False).index].copy()
+
+    # Determine for each ZIP, how much demand to assign to each load substation
+    # Assume here that generator substations don't have load attached to them
+    filtered_subs = sorted_subs.loc[~sorted_subs.index.isin(plant["sub_id"])]
+    subs_per_zip = filtered_subs.value_counts("ZIP")
+    zip_load_substations = subs_per_zip * const.substation_load_share
+    zip_load_substations = zip_load_substations.round().clip(lower=1)
+    zip_assigned_population = (zip_data["population"] / zip_load_substations).dropna()
+    # Select the N substations per ZIP with greatest transmission capacity
+    load_substations = pd.concat(
+        df.head(int(zip_load_substations[name]))
+        for name, df in filtered_subs.groupby("ZIP")
+    )
+    substations["pop_ZIP"] = load_substations["ZIP"].map(zip_assigned_population)
+
+    # Assign remaining county population to substations with load already,
+    # plus the most connected substation in any county without a load substation.
+    load_subs_from_zips = substations.query("pop_ZIP > 0")
+    load_subs_per_county = load_subs_from_zips.value_counts("COUNTYFIPS")
+    county_pop = county_data["population"]
+
+    # Select the one substation per missing county with greatest transmission capacity
+    counties_without_load_subs = set(county_pop.index) - set(load_subs_per_county.index)
+    subs_in_counties_without_load_subs = sorted_subs.loc[
+        sorted_subs["COUNTYFIPS"].isin(counties_without_load_subs)
+    ]
+    added_load_subs = pd.concat(
+        df.head(1)
+        for name, df in subs_in_counties_without_load_subs.groupby("COUNTYFIPS")
+    )
+    load_subs = pd.concat([load_subs_from_zips, added_load_subs])
+    load_subs_per_county = load_subs_per_county.reindex(county_pop.index).fillna(1)
+
+    # Distribute population remaining after ZIP distribution to identified load buses
+    distributed_pop = load_subs.groupby("COUNTYFIPS")["pop_ZIP"].sum()
+    remaining_pop = county_pop - distributed_pop.reindex(county_pop.index).fillna(0)
+    remaining_pop_per_sub = remaining_pop.clip(lower=0) / load_subs_per_county
+    # We may still miss some population, since there may be a county without any
+    # substations, but we should cover the vast majority.
+    substations["pop_county"] = load_subs["COUNTYFIPS"].map(remaining_pop_per_sub)
+
+    # Translate population to demand
+    total_pop = substations["pop_ZIP"].fillna(0) + substations["pop_county"].fillna(0)
+    sub_demand = total_pop * const.demand_per_person
+
+    load_buses = pd.concat(
+        df.head(1) for sub_id, df in bus.sort_values("baseKV").groupby("sub_id")
+    )
+    bus["Pd"] = load_buses["sub_id"].map(sub_demand).reindex(bus.index).fillna(0)
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -111,6 +111,8 @@ blob_paths = {
     "epa_needs": "https://besciences.blob.core.windows.net/datasets/EPA_NEEDS/needs-v620_06-30-21-2_active.csv",
     "substations": "https://besciences.blob.core.windows.net/datasets/hifld/Electric_Substations_Jul2020.csv",
     "transmission_lines": "https://besciences.blob.core.windows.net/datasets/hifld/Electric_Power_Transmission_Lines_Jul2020.geojson.zip",
+    "us_counties": "https://besciences.blob.core.windows.net/datasets/geo_data/uscounties.csv",
+    "us_zips": "https://besciences.blob.core.windows.net/datasets/geo_data/uszips.csv",
 }
 eia_epa_crosswalk_path = "https://raw.githubusercontent.com/Breakthrough-Energy/camd-eia-crosswalk/master/epa_eia_crosswalk.csv"
 
@@ -386,6 +388,7 @@ heat_rate_assumptions = {
     "Solar Thermal without Energy Storage": 0,
 }
 
+
 # These lines were manually identified based on a combination of: their 'TYPE'
 # classification, their substation names, and their geographical paths. The capacities
 # for each line were compiled from a variety of public sources.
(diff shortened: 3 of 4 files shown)


ATTRIBUTION.md
```diff
@@ -447,4 +447,36 @@ The data are used to populate the power generation profile of U.S. transmission
 ##### Note
 Public use (see https://hifld-geoplatform.opendata.arcgis.com/datasets/geoplatform::power-plants/about). Users are advised to read the data set's metadata thoroughly to understand appropriate use and data limitations.
 
+
+##### Source
+* Name: U.S. Zips
+* Author: simplemaps.com
```
BainanXia, line 453: Not sure whether we should put down the company name in the Author field: `Pareto Software, LLC`?

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    then to the lowest-voltage bus within each substation.
+    This information is added inplace as a 'Pd' column to the ``bus`` data frame.
+
+    :param pandas.DataFrame substations: table of substation data.
+    :param pandas.DataFrame branch: table of branch data.
+    :param pandas.DataFrame plant: table of plant data.
+    :param pandas.DataFrame bus: table of bus data.
+    """
+    # Load data
+    data_dir = os.path.join(os.path.dirname(inspect.getfile(hifld)), "data")
```
BainanXia, line 21: Can we simply do `data_dir = os.path.join(os.path.dirname(os.path.dirname(__file__), "data")` here?

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+
+import pandas as pd
+
+from prereise.gather.griddata import hifld
+from prereise.gather.griddata.hifld import const
+
+
+def assign_demand_to_buses(substations, branch, plant, bus):
+    """Using data on population by county and ZIP code, assign demand to substations,
```
BainanXia, line 11: Maybe `demand` -> `demand parameter` to be specific?

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: For now, this would be equivalent, but I figured that the location of the data directory would be more stable if it's defined as relative to the `prereise.gather.griddata.hifld` directory, rather than relative to the location of this chunk of code (which might be moved in a refactor).

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: In the function name?

ATTRIBUTION.md
danielolsen: Good call, done.

prereise/gather/griddata/hifld/data_process/demand.py
BainanXia: Nope, in the doc strings.

prereise/gather/griddata/hifld/data_process/demand.py
BainanXia: Agree. However, we will need to import the `hifld` module just for the `inspect` call here. Maybe even we move this chunk of code around, it might still end up with two hops from the root? Anyway, will defer to your decision here.

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: I made a tweak, let me know what you think.

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: I bet you're right that this code will probably stay in a compatible place, but I prefer the more stable version, even if it's a little more verbose.

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    data_dir = os.path.join(os.path.dirname(inspect.getfile(hifld)), "data")
+    zip_path = os.path.join(data_dir, "uszips.csv")
+    zip_data = pd.read_csv(zip_path, dtype={"zip": "string"}).set_index("zip")
+    county_data = pd.read_csv(os.path.join(data_dir, "uscounties.csv"))
+
+    # Determine each substation's transmission capacity, then sort for selection
+    filtered_branch = branch.query("SUB_1_ID != SUB_2_ID")
+    from_cap = filtered_branch.groupby("SUB_1_ID").sum()["rateA"]
+    to_cap = filtered_branch.groupby("SUB_2_ID").sum()["rateA"]
+    sub_cap = from_cap.combine(to_cap, lambda x, y: x + y, fill_value=0)
```
BainanXia, line 30: I was thinking about refactoring `calculate_substation_capacity` function in `powersimdata/design/transmission/substations.py`, which currently takes a grid object as input, but failed to come up with a compatible idea given the column names are different as well...Let's keep these lines here then.

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    # Sort substations by their capacities for later ordered selection
+    sorted_subs = substations.loc[sub_cap.sort_values(ascending=False).index].copy()
+
+    # Determine for each ZIP, how much demand to assign to each load substation
+    # Assume here that generator substations don't have load attached to them
+    filtered_subs = sorted_subs.loc[~sorted_subs.index.isin(plant["sub_id"])]
+    subs_per_zip = filtered_subs.value_counts("ZIP")
+    zip_load_substations = subs_per_zip * const.substation_load_share
+    zip_load_substations = zip_load_substations.round().clip(lower=1)
+    zip_assigned_population = (zip_data["population"] / zip_load_substations).dropna()
```
BainanXia, line 40: When I was reading through the reference code from collaborators, I was thinking whether we should distribute load (population) proportional to the substation capacities instead of uniformly. Do you think that will make a difference?

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    # Select the N substations per ZIP with greatest transmission capacity
+    load_substations = pd.concat(
+        df.head(int(zip_load_substations[name]))
+        for name, df in filtered_subs.groupby("ZIP")
```
BainanXia, line 44: Aha, nice.

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    substations["pop_ZIP"] = load_substations["ZIP"].map(zip_assigned_population)
+
+    # Assign remaining county population to substations with load already,
+    #   plus the most connected substation in any county without a load substation
+    load_subs_from_zips = substations.query("pop_ZIP > 0")
+    load_subs_per_county = load_subs_from_zips.value_counts("COUNTYFIPS")
+    county_pop = county_data.set_index("county_fips")["population"]
```
BainanXia, line 52: To be consistent, let's do `set_index` for `county_data` right after `read_csv`, as we do for `zip_data`.

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    zip_assigned_population = (zip_data["population"] / zip_load_substations).dropna()
+    # Select the N substations per ZIP with greatest transmission capacity
+    load_substations = pd.concat(
+        df.head(int(zip_load_substations[name]))
+        for name, df in filtered_subs.groupby("ZIP")
+    )
+    substations["pop_ZIP"] = load_substations["ZIP"].map(zip_assigned_population)
+
+    # Assign remaining county population to substations with load already,
+    #   plus the most connected substation in any county without a load substation
```
BainanXia, line 49: Unexpected spaces in doc strings here?

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: It will definitely make some kind of difference, but I'm not not sure in which general direction the difference will be. Thinking about a low-medium density area, I would imagine that the population is fairly evenly spread out, and so distributing uniformly probably makes sense, since a high-capacity substation may just be a collector, not in response to a pocket of density. However, in an area that truly has different densities, higher-capacity substations may actually be in reaction to higher density. Without further information, I'm not sure what the best conclusion is.

If we run a simulation with this approach and find issues, then I think we may need to revisit this question. I think it's also potentially related to https://github.com/Breakthrough-Energy/PreREISE/issues/234; we may want to patch the algorithm, or we may alternatively want to patch the data that feeds the algorithm.

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: I did this as a continuation of the sentence of the previous line. Is this against our conventions, and if so, what's the better approach?

prereise/gather/griddata/hifld/data_process/demand.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    load_subs_from_zips = substations.query("pop_ZIP > 0")
+    load_subs_per_county = load_subs_from_zips.value_counts("COUNTYFIPS")
+    county_pop = county_data.set_index("county_fips")["population"]
+
+    # Select the one substation per missing county with greatest transmission capacity
+    counties_without_load_subs = set(county_pop.index) - set(load_subs_per_county.index)
+    subs_in_counties_without_load_subs = sorted_subs.loc[
+        sorted_subs["COUNTYFIPS"].isin(counties_without_load_subs)
+    ]
```
BainanXia, line 59: Any luck it fit in one line via `query`?
```python
subs_in_counties_without_load_subs = sorted_subs.query("COUNTYFIPS.isin(@counties_without_load_subs)")
```

BainanXia: Oops, maybe not. Never mind.

danielolsen: It was like that before, but then the linter complained about not having used the `counties_without_load_subs` variable, and there was no room to add a `# noqa` comment on _that_ line.

prereise/gather/griddata/hifld/data_process/demand.py
```diff
@@ -0,0 +1,81 @@
...
+    added_load_subs = pd.concat(
+        df.head(1)
```
BainanXia, line 61: I was thinking whether it is worth doing a similar N distribution instead of picking the one with largest transmission capacity.

prereise/gather/griddata/hifld/data_process/demand.py
BainanXia: Aha, what a hack!

prereise/gather/griddata/hifld/data_process/demand.py
BainanXia: I've no idea. My impression is we always have same spaces after the hashtag. I could be wrong.

danielolsen: I changed it to be even. It definitely looks less strange that way, and hopefully it's clear enough from the context that the two lines are part of one sentence.

BainanXia (approved):
Very nice. Thanks!

danielolsen:
I chatted with @rouille today, and he raised a suggestion: should these CSVs live in the blob storage, rather than the repo?

BainanXia:
> I chatted with @rouille today, and he raised a suggestion: should these CSVs live in the blob storage, rather than the repo?

Good call. Looking at the size of the two files, <10M in total. I would say it won't hurt to leave it in the repo for now so that it is easier to move around. But it is also clean put them in the blob storage if we have a folder structure in mind (we still have decent amount of files in the repo now). Again, your choice.

danielolsen:
This feature has been refactored to get the county and ZIP data files from blob storage, rather than from the repo itself. The new call signature (including changes to `build_transmission` via #237):
```python
from prereise.gather.griddata.hifld.data_process.demand import assign_demand_to_buses
from prereise.gather.griddata.hifld.data_process.generators import build_plant
branch, bus, sub, dcline = build_transmission()
generators = build_plant(bus, sub)
assign_demand_to_buses(sub, branch, generators, bus)
```

PR #237, opened 2021-11-02
feat: filter DC lines out of AC transmission network processing within HIFLD construction

danielolsen:

### Purpose
Partial fulfilment of #233: this addresses the HVDC transmission lines within the HIFLD dataset, but does not yet address the B2Bs, which will require a different method of splitting substations into sub-substations for each interconnection.

### What the code is doing
In **const.py**, ratings are added for DC lines with known capacities.

In **transmission.py**:
- A small helper function `split_lines_to_ac_and_dc` is added to filter out DC lines via some combination of the `TYPE` attribute of the HIFLD dataset and manually specified indices.
- `build_transmission` uses this function, then adds impedances and AC ratings to AC branches only (using functionality introduced in #218 and #219), and uses the pre-defined ratings from **const.py** to add capacities to the DC lines.

### Testing
Tested manually.

### Usage Example/Visuals
```python
>>> branch, substations, dc_lines = build_transmission()
filter substations based on lines
---------------------------------
assigning substations to lines' endpoints by mapping their rounded (3 digits) coordinates
dropping 3314 lines having same endpoints coordinates after rounding
finding closest substation to unmapped lines' endpoint(s)
100%|███████████████████████████████████| 2624/2624 [00:00<00:00, 19820.94it/s]
dropping 0 lines due to island size filtering
dropping 0 substations due to island size filtering
Reading cached minimum spanning tree
2948 line voltages can't be found via neighbor consensus
No more missing voltages remain after neighbor minimum
>>> branch.columns
Index(['TYPE', 'STATUS', 'NAICS_CODE', 'NAICS_DESC', 'SOURCE', 'SOURCEDATE',
       'VAL_METHOD', 'VAL_DATE', 'OWNER', 'VOLTAGE', 'VOLT_CLASS', 'INFERRED',
       'SUB_1', 'SUB_2', 'COORDINATES', 'SUB_1_ID', 'SUB_2_ID', 'type',
       'length', 'from_bus_id', 'to_bus_id', 'x', 'rateA'],
      dtype='object')
>>> dc_lines.columns
Index(['TYPE', 'STATUS', 'NAICS_CODE', 'NAICS_DESC', 'SOURCE', 'SOURCEDATE',
       'VAL_METHOD', 'VAL_DATE', 'OWNER', 'VOLTAGE', 'VOLT_CLASS', 'INFERRED',
       'SUB_1', 'SUB_2', 'COORDINATES', 'SUB_1_ID', 'SUB_2_ID', 'Pmax',
       'Pmin'],
      dtype='object')
>>> dc_lines["Pmax"]
200823    3100
308464    2400
310053     400
311958       5
157627     330
157629     660
158515    2000
108354     500
113313     660
131914    2000
Name: Pmax, dtype: int64
```

Branch entries get impedance (`x`) and rating, DC lines get `Pmax` and `Pmin`.

### Time estimate
15 minutes.

### DIFF
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -385,3 +385,20 @@ heat_rate_assumptions = {
     "Solar Thermal with Energy Storage": 0,
     "Solar Thermal without Energy Storage": 0,
 }
+
+# These lines were manually identified based on a combination of: their 'TYPE'
+# classification, their substation names, and their geographical paths. The capacities
+# for each line were compiled from a variety of public sources.
+dc_line_ratings = {  # MW
+    108354: 500,  # Square Butte
+    113313: 660,  # Neptune Cable
+    131914: 2000,  # Quebec – New England Transmission (Ayer to Monroe)
+    150123: 1000,  # CU
+    157627: 330,  # Cross-Sound Cable
+    157629: 660,  # Hudson Project
+    158515: 2000,  # Quebec – New England Transmission (Quebec to Monroe)
+    200823: 3100,  # Pacific DC Intertie (Path 65)
+    308464: 2400,  # Intermountain Power Project (Path 27)
+    310053: 400,  # Trans-Bay Cable
+    311958: 5,  # Alamogordo Solar Energy Center
+}
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -604,6 +604,22 @@ def estimate_branch_rating(branch, bus_voltages):
     raise ValueError(f"{branch.loc['type']} not a valid branch type")
 
 
+def split_lines_to_ac_and_dc(lines, dc_override_indices=None):
+    """Given a data frame of mixed AC & DC lines, where some DC lines are not
+    appropriately labeled, split into an AC data frame and a DC data frame.
+
+    :param pandas.DataFrame lines: combined data frame of AC & DC lines.
+    :param iterable dc_override_indices: indices to coerce to DC classification.
+    :return: (*tuple*) -- two data frames, of AC & DC lines, respectively.
+    """
+    if dc_override_indices is None:
+        dc_override_indices = {}
+    dc_types = {"DC; OVERHEAD", "DC; UNDERGROUND"}  # noqa: F841
+    dc_lines = lines.query("TYPE in @dc_types or index in @dc_override_indices")
+    ac_lines = lines.query("index not in @dc_lines.index")
+    return ac_lines.copy(), dc_lines.copy()
+
+
 def build_transmission(method="line2sub", **kwargs):
     """Build transmission network
 
--- a/prereise/gather/griddata/hifld/data_process/transmission.py
+++ b/prereise/gather/griddata/hifld/data_process/transmission.py
@@ -632,6 +632,8 @@ def build_transmission(method="line2sub", **kwargs):
         :py:func:`get_mst_edges`.
     :raises TypeError: if ``method`` is not a str.
     :raises ValueError: if ``method`` is unknown.
+    :return: (*tuple*) -- four data frames:
+        AC branches, buses, substations, and DC lines, respectively.
     """
     if not isinstance(method, str):
         raise TypeError("method must be a str")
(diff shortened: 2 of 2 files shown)


prereise/gather/griddata/hifld/const.py
```diff
@@ -385,3 +385,17 @@
...
+    131914: 2000,  # Quebec – New England Transmission (Ayer to Monroe)
+    150123: 1000,  # CU
+    157627: 330,  # Cross-Sound Cable
+    157629: 660,  # Hudson Project
+    158515: 2000,  # Quebec – New England Transmission (Quebec to Monroe)
+    200823: 3100,  # Pacific DC Intertie (Path 65)
+    308464: 2400,  # Intermountain Power Project (Path 27)
+    310053: 400,  # Trans-Bay Cable
+    311958: 5,  # Alamogordo Solar Energy Center
+}
```
BainanXia, line 401: Let's put down the data source of these references somewhere in the doc string.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -668,12 +686,33 @@ def build_transmission(method="sub2line", **kwargs):
         lines, substations, **island_kwargs
     )
 
+    # Separate DC lines from AC line processing, add ratings
+    ac_lines, dc_lines = split_lines_to_ac_and_dc(lines, const.dc_line_ratings.keys())
+    dc_lines["Pmax"] = dc_lines.index.to_series().map(const.dc_line_ratings)
```
BainanXia, line 691: So the total number of dc lines we have is `len(const.dc_line_ratings)`, and we don't have any line with `TYPE in dc_types` but index is not in `const.dc_line_ratings`?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -668,12 +686,33 @@ def build_transmission(method="sub2line", **kwargs):
...
+    dc_lines["Pmin"] = -1 * dc_lines["Pmax"]
```
BainanXia, line 692: Are we going to have asymmetric dc lines in the future?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -668,12 +686,33 @@ def build_transmission(method="sub2line", **kwargs):
...
+
     # Add voltages to lines with missing data
-    augment_line_voltages(lines, substations)
+    augment_line_voltages(ac_lines, substations)
+
+    # Create buses from lines
+    bus = create_buses(ac_lines)
```
BainanXia, line 698: Not sure whether we should return `bus` data frame as well, otherwise we will need to call `create_buses` in the higher-level orchestration function again using `branch` as inputs, right?

prereise/gather/griddata/hifld/const.py
danielolsen: The lines themselves were identified based on personal judgement (substation names, geographical paths), and the capacities were compiled based on variety of sources. I'm not sure how to really cite that.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: That is correct. The function could identify DC lines based on the type without us knowing their capacity, it's just that in this particular application we do know.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Good catch, done.

prereise/gather/griddata/hifld/const.py
rouille: Just wondering, since personal judgment is involved, should we create a DC line file with the needed information that would be loaded and remove all DC lines from the transmission file?

danielolsen: I think this way seems more transparent: we intake raw un-processed external data, then apply the minimum amount of assumptions, made in the open.

BainanXia: Maybe we can just put down this line for future references:
`These DC lines are identified based on substation names, geographical paths, etc. manually, the capacities of which are compiled based on public accessible sources.`?

danielolsen: Good idea, I just pushed something to that effect.

BainanXia (approved):
Good to go. Thanks!

PR #240, opened 2021-11-16
feat: separate interconnections via network topology

### LINKED ISSUE #233: Improve handling of DC lines when loading HIFLD data

danielolsen:

### Purpose
- Load substations whose status is either `IN SERVICE` or `NOT AVAILABLE`. Although `NOT AVAILABLE` substations are only about 5% of the total, they're about 20% of the substations within Texas and New Mexico, which causes lines to get mapped to faraway substations which are sometimes in other interconnections.
- Add a function which can separate and label the `interconnect` attribute of lines and substations, given input on where to split the substations and how to re-join them.
- Add data that when input into the new function, splits the HIFLD network into three interconnections. This data is imperfect, especially regarding the Eastern/ERCOT seam around Lubbock and the Texas panhandle more broadly, but it demonstrates how the new function works and should be able to be revised later without any code changes.

Closes #233.

### What the code is doing
- **const.py** gets the (imperfect) information which splits the interconnection into different networks.
- **topology.py** gets the new function `add_interconnects_by_connected_components` that:
  - Builds a NetworkX graph of the network once a subset of substations (and all lines connected to them) are removed.
  - Labels the lines of the connected components based on assumptions of size (Eastern > Western > ERCOT).
  - Labels lines based on user-supplied information (assumptions for lines & substations).
  - Labels lines based on the interconnects of their neighbors at non-dropped substations.
  - Using the `interconnect` labels for lines, identify which interconnects are present within substations that get split, and create new pseudo-substations for each interconnect within a substation, and re-connect lines as necessary. New pseudo-substations get added, and substations which got split are removed.
  - Labels the substations based on the new topology.
 - **transmission.py** uses this new function to assign interconnects to lines & substations, rather than the assumptions by county as previously.

### Testing
Tested manually:
```python
branch, bus, substations, dclines = build_transmission()
```

Validation:
```python
>>> import networkx as nx
>>> g = nx.convert_matrix.from_pandas_edgelist(
...     branch.query("not SUB_1_ID.isnull()"), "SUB_1_ID", "SUB_2_ID",
... )
>>> interconnects = list(nx.connected_components(g))
>>> sorted_interconnects = sorted(interconnects, key=len)[::-1]
>>> for s in sorted_interconnects:
...     print(f"connected component with {len(s)} substations")
...     print("substations:", substations.loc[s].value_counts("interconnect"))
...     print("lines:", branch.query("SUB_1_ID in @s or SUB_2_ID in @s").value_counts("interconnect"))
...
connected component with 45141 substations
substations: interconnect
Eastern    45141
dtype: int64
lines: interconnect
Eastern    62046
dtype: int64
connected component with 12983 substations
substations: interconnect
Western    12983
dtype: int64
lines: interconnect
Western    17910
dtype: int64
connected component with 3352 substations
substations: interconnect
ERCOT    3352
dtype: int64
lines: interconnect
ERCOT    4980
dtype: int64
```

### Usage Example/Visuals

Here are the transmission lines for the states required to define the interconnections, colored by interconnect, with tiny yellow stars at the location of the dropped substations:
map_all_substations_reconnected_reduced

### Time estimate
30-60 minutes for the current code/logic. I'm also going to add some tests and try to factor out some of the for loops that were added for convenience but could probably be equivalently done in a cleaner way, although the elapsed time for these steps is quite short so it doesn't need too much optimization.

### DIFF
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -534,3 +534,5 @@ b2b_ratings = {  # MW
     "UNKNOWN304994": 600,  # Welsh (Eastern/ERCOT)
     "VIRGINIA SMITH CONVERTER STATION": 200,  # a.k.a. 'Sidney' (Eastern/Western)
 }
+
+interconnect_size_rank = ["Eastern", "Western", "ERCOT"]
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -415,3 +415,56 @@ contiguous_us_bounds = {
     "south": 25,
     "west": -125,
 }
+
+seams_substations = {
+    "east_west": {
+        202364,  # 'LAMAR HVDC TIE'
+        202159,  # 'UNKNOWN202159', NE (effecitvely Sidney/Virginia Smith)
+        202160,  # 'VIRGINIA SMITH CONVERTER STATION'
+        202177,  # Sidney-adjacent
+        202178,  # 'STEGAL', NE
+        131797,  # Stegal-adjacent
+        203572,  # 'MILES CITY', MT (the substation that appears more likely to be the real one)
+        203590,  # 'RICHARDSON COULEE' (near MALTA, shouldn't be necessary?)
+        303738,  # 'BLACKWATER TIE', NM
+        304165,  # 'EDDY AC-DC-AC TIE', NM
+        # Rapid City Disconnections
+        131171,  # North of Rapid City
+        131176,  # North of Rapid City
+        202567,  # East of Rapid City
+        # Highline NE/CO border
+        205884,  # Julesburg, CO
+        205888,  # Holyoke, CO
+        203719,  # 'ALVIN' substation
+    },
+    "east_ercot": {
+        161924,  # Logansport, TX
+        300490,  # Vernon, TX
+        301314,  # Valley Lake, TX connection to OK
+        301729,  # Hawkins, TX
+        302012,  # Vernon, TX
+        302274,  # 'COTTONWOOD', Glenn, TX
+        303004,  # Crowell, TX
+        303646,  # San Augustine, TX
+        303719,  # Big Sandy, TX
+        304100,  # Matador, TX
+        304328,  # Midland, TX
+        304477,  # Oklaunion substation (B2B)
+        304825,  # Dennison, TX connection to OK
+        304391,  # Long Branch, TX
+        304994,  # Welsh substation (B2B)
+        306058,  # Munday, TX
+        306638,  # Pittsburg, TX
+        306738,  # Henderson, TX
+        307121,  # Kirkland, TX
+        307363,  # Navasota, TX
+        307539,  # Mt. Pleasant, TX
+        307735,  # Shiro, TX
+        308062,  # Lufkin, TX
+        308951,  # Beckville, TX
+        308976,  # Dayton, TX
+        309403,  # Kilgore, TX
+        310861,  # Overton, TX
+        310879,  # Huntsville, TX
+    },
+}
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -468,3 +468,57 @@ seams_substations = {
         310879,  # Huntsville, TX
     },
 }
+
+substation_interconnect_assumptions = {
+    "Eastern": {
+        131171,
+        131172,
+        131853,
+        161925,
+        167678,
+        167679,
+        167681,
+        167682,
+        167684,
+        307364,
+    },
+    "Western": {
+        201396,
+        202172,
+        205667,
+        205889,
+        205890,
+    },
+    "ERCOT": {
+        301181,
+        301291,
+        302826,
+        303024,  # Substations between East/ERCOT AC connector and Oklaunion B2B station
+        303394,
+        303406,
+        303433,
+        304994,  # Welsh B2B
+        309433,
+        309658,
+    },
+}
+
+line_interconnect_assumptions = {
+    "Eastern": {
+        128641,
+        132264,
+        135527,
+        141367,
+        300170,
+        301858,
+        303906,
+        305887,
+        306332,
+        306885,
+        310668,
+        311279,
+        311520,
+    },
+    "Western": {123525, 141873},
+    "ERCOT": {305330, 309428, 310121},
+}
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -522,3 +522,15 @@ line_interconnect_assumptions = {
     "Western": {123525, 141873},
     "ERCOT": {305330, 309428, 310121},
 }
+
+b2b_ratings = {  # MW
+    "BLACKWATER TIE": 200,  # a.k.a. 'Clovis'/'Roosevelt County' (Eastern/Western)
+    "EDDY AC-DC-AC TIE": 200,  # a.k.a. 'Artesia' (Eastern/Western)
+    "LAMAR HVDC TIE": 210,  # (Eastern/Western)
+    "MILES CITY": 200,  # (Eastern/Western)
+    "NEW UNDERWOOD": 200,  # representative of the Rapid City DC Tie (Eastern/Western)
+    "STEGALL": 110,  # (Eastern/Western)
+    "UNKNOWN304477": 220,  # Oklaunion (Eastern/ERCOT)
+    "UNKNOWN304994": 600,  # Welsh (Eastern/ERCOT)
+    "VIRGINIA SMITH CONVERTER STATION": 200,  # a.k.a. 'Sidney' (Eastern/Western)
+}
--- a/prereise/gather/griddata/hifld/data_process/topology.py
+++ b/prereise/gather/griddata/hifld/data_process/topology.py
@@ -4,6 +4,7 @@ import pickle
 from itertools import combinations, product
 
 import networkx as nx
+import pandas as pd
 from powersimdata.utility.distance import haversine, ll2uv
 from scipy.spatial import KDTree
 from tqdm import tqdm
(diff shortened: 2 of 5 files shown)


prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -110,7 +110,7 @@ def map_lines_to_substations_using_coords(
     }
     for a in all2one:
         type2id = substations.loc[list(a)].reset_index().groupby("TYPE").first()["ID"]
-        for t in ["SUBSTATION", "TAP", "RISER", "DEAD END"]:
+        for t in ["SUBSTATION", "TAP", "RISER", "DEAD END", "NOT AVAILABLE"]:
```
BainanXia, line 113: So "NOT AVAILABLE" is not only one of the status of substations but also one of substation types that we didn't consider in picking those as a primary substation before?

danielolsen: Exactly. Some of the `NOT AVAILABLE` status substations also have type `NOT AVAILABLE`, so we need to add to this list here, otherwise we end up with NaN/None values downstream when the only substation at a given location is not one of `["SUBSTATION", "TAP", "RISER", "DEAD END"]`.

prereise/gather/griddata/hifld/const.py
```diff
@@ -415,3 +415,112 @@
...
+        307539,  # Mt. Pleasant, TX
+        307735,  # Shiro, TX
+        308062,  # Lufkin, TX
+        308951,  # Beckville, TX
+        308976,  # Dayton, TX
+        309403,  # Kilgore, TX
+        310861,  # Overton, TX
+        310879,  # Huntsville, TX
+    },
+}
```
BainanXia, line 470: We don't need to split between `west` and `ercot`? I could tell from the plot that `ercot` is surrounded by `east` only, however I'm wondering whether that is the case in practice. Also, the projection of the plot is a little weird which seems to be over shrunk along the east-west axis.

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -217,3 +218,100 @@ def get_mst_edges(lines, substations, **kwargs):
...
+    interconnect_size_rank,
+):
+    """Disconnect a large connected component using a set of connecting substations,
+    label the resulting connected components, then reconnect dropped lines and
+    substations, using explicit assumptions plus inference from neighboring lines. The
+    ``lines`` and ``substations`` data frames are modified inplace with a new
+    'interconnect' column.
+
+    :param pandas.DataFrame lines: data frame of line information.
+    :param pandas.DataFrame lines: data frame of substations information.
```
BainanXia, line 238: `lines` -> `substations`

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -217,3 +218,100 @@ def get_mst_edges(lines, substations, **kwargs):
...
+            lines["SUB_1_ID"].isin(sub_ids) | lines["SUB_2_ID"].isin(sub_ids),
+        ] = interconnect
+    for interconnect, line_ids in line_assumptions.items():
+        labels.loc[line_ids] = interconnect
+    # Use neighboring lines at non-dropped substations to infer line interconnections
+    dropped_lines = lines.loc[labels == "dropped"]
+    non_dropped_lines = lines.loc[labels != "dropped"]
+    for id, line in dropped_lines.iterrows():
+        non_dropped_sub = (  # noqa: F841
+            line.SUB_1_ID if line.SUB_2_ID in seams_substations else line.SUB_2_ID
```
BainanXia, line 272: So it is guaranteed there are no lines with both ends in seams substations, right?

prereise/gather/griddata/hifld/const.py
danielolsen: As far as I can tell, there are no substations that need to be disconnected that would otherwise bridge West and ERCOT. The plot projection is very crude, I was working within `matplotlib` for convenience and not enforcing a proper aspect ratio, but the coordinates themselves aren't being touched within this code so they should project properly when used the right way.

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -217,3 +218,100 @@ def get_mst_edges(lines, substations, **kwargs):
...
+        first_new_sub_id = substations.index.max() + 1
+        new_substations = pd.concat(
+            [substations.loc[sub_id]] * len(new_sub_interconnects), axis=1
+        ).T
+        new_substations.index = pd.RangeIndex(
+            first_new_sub_id, first_new_sub_id + len(new_sub_interconnects)
+        )
+        new_substations["NAME"] = [
+            substations.loc[sub_id, "NAME"] + f"_{i}" for i in new_sub_interconnects
+        ]
```
BainanXia, line 299: Keeping track of this using `NAME` field is nice, not sure whether it is more straight forward to inferring this using `id` directly, i.e. `xxxx_1` and `xxxx_2` where `xxxx` is the old sub id and the numbers are interconnect rankings. It is not necessarily to have consecutive numbers sub_ids right? We might need them to be integers though.

prereise/gather/griddata/hifld/data_process/topology.py
```diff
@@ -217,3 +218,100 @@ def get_mst_edges(lines, substations, **kwargs):
...
+        for new_id, new_substation in new_substations.iterrows():
+            substations.loc[new_id] = new_substation
+        # Re-map the labelled lines to the new substations
+        for line_id, line in sub_dropped_lines.iterrows():
+            for new_sub_id, sub in new_substations.iterrows():
+                if line["interconnect"] == sub["NAME"].split("_")[1]:
+                    if line["SUB_1_ID"] == sub_id:
+                        lines.loc[line_id, "SUB_1_ID"] = new_sub_id
+                    if line["SUB_2_ID"] == sub_id:
+                        lines.loc[line_id, "SUB_2_ID"] = new_sub_id
```
BainanXia, line 310: Not sure whether this could be simplified if we give an interconnect tag to substations first.

BainanXia:
Thanks for making all the wild ideas happen. Although I'm aware of the full story line via offline discussions, I still found the logic brain burning when reading through the code. Here is my two cents popping up upon review, not sure whether that would be clearer:
- one function for splitting into certain number of islands given a list of chosen nodes dropped
- one function for adding interconnect tag (column) to the nodes in each island based on the given interconnect size ranking
- one function for making copies of dropped nodes based on the other ends of the lines connecting to it (assuming we don't have lines with both ends dropped), then reconnect the corresponding lines to the new substations, adding those nodes and edges back.
In this way, we don't need to label lines with interconnects but substations only. In the end, we could have a sanity check to see whether all lines with both ends locating into the same interconnection.

danielolsen:
@BainanXia I like your plan of splitting `add_interconnects_by_connected_components` into smaller sub-functions. That would make it easier to follow the logic and easier to test as well. For your suggestion of refactoring the third part to not assign interconnects to lines, would that impact the results at all, or would it just be a (potentially simpler) way to accomplish the same thing in a different way?

BainanXia:
> @BainanXia I like your plan of splitting `add_interconnects_by_connected_components` into smaller sub-functions. That would make it easier to follow the logic and easier to test as well. For your suggestion of refactoring the third part to not assign interconnects to lines, would that impact the results at all, or would it just be a (potentially simpler) way to accomplish the same thing in a different way?

It's just an alternative way to achieve the same thing. I was thinking all we need is to tag every node with interconnections rather than edges via assumptions + neighbors. Specifically, if one node has neighbors with multiple interconnection tags (dropped nodes), we make copies and tag each copy with one interconnect among the set.

prereise/gather/griddata/hifld/data_process/topology.py
danielolsen: It is not guaranteed, in fact it does happen in our data set! The path of connection from the Eastern interconnect to the Oklaunion B2B station appears to go from a full-Eastern substation through another intermediate mixed-interconnect substation first before getting to the B2B substation, so both the B2B substation and the 'intermediate' are considered to be 'part of the seam', since both have buses in both interconnects, and the line that connects the two has both ends 'on the seam'. With the input data we have right now, the interconnects still do end up getting separated properly, but there's definitely potential for this sort of edge case to bite us with different input data. This would also preclude your other refactor idea of always looking to the substation at the 'other end' to determine which interconnect a line should connect to within a dropped substation.

prereise/gather/griddata/hifld/data_process/topology.py
danielolsen: `We might need them to be integers though`: I think you might be right, I'm not sure how much downstream code is implicitly expecting integers vs. just some sort of unique index.

prereise/gather/griddata/hifld/data_process/topology.py
danielolsen: Done.

danielolsen:
B2B facility information has been added, as well as a function to add these B2B lines to the DC line table by looking at the now-separated substations. Therefore, this should now fulfill all remaining requirements of #233.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -622,6 +623,33 @@ def split_lines_to_ac_and_dc(lines, dc_override_indices=None):
...
 
 
+def add_b2bs_to_dc_lines(dc_lines, substations, b2b_ratings):
+    """Given back-to-back converter station ratings, add entries to the DC lines table
+    (modified inplace) representing the HVDC links between interconnections.
+
+    :param pandas.DataFrame dc_lines: table of HVDC line information.
+    :param pandas.DataFrame substations: table of substation information.
+    :param dict/pandas.Series b2b_capacities: capacities of B2B HVDC facilties.
+    :raises ValueError: if
```
kasparm, line 633: Did something got cut off here?

kasparm (approved):
I can reproduce your test and see that the B2B where added to the dclines.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Fixed.

PR #243, opened 2021-12-03
feat: adding data processing script

isond:

### Purpose
To add the data processing script that will organize the NHTS data.

### What the code is doing
How is the purpose executed?
This code filters through the NHTS data based on the inputted census division and reorganizes the information. It also calculates additional information based on the given data.

### Testing
This was tested through manual testing using data from "trippub.csv".

### Where to look
The new code is located at "PreREISE/prereise/gather/demanddata/transportation_electrification/data_process.py"

data_process.py which is currently the only file

### Usage Example/Visuals

### Time estimate
It should take a few days for reviewers to understand.

### DIFF
--- a/prereise/gather/demanddata/transportation_electrification/data_process.py
+++ b/prereise/gather/demanddata/transportation_electrification/data_process.py
@@ -21,6 +21,25 @@ def load_data(census_division: int, filepath: str = "trippub.csv"):
     return df
 
 
+def calculate_dwell_time(data: pd.DataFrame):
+    """Calculates the dwell time, how long a vehicle has been charging
+
+    :param pd.DataFrame data: the data to calculate the dwell time from
+    :return: (*pd.Series*) -- list of dwell times
+    """
+    dwells = (
+        data["Start time (hour decimal)"].iloc[1:].values
+        - data["End time (hour decimal)"].iloc[:-1]
+    )
+    dwells.loc[data.index[-1]] = (
+        24
+        - data["End time (hour decimal)"].iloc[-1]
+        + data["Start time (hour decimal)"].iloc[0]
+    )
+
+    return dwells
+
+
 def data_filtering(census_division):
     """Filter raw NHTS data to be used in mileage.py
 
@@ -97,176 +116,18 @@ def data_filtering(census_division):
         / sorted_data["Travel time (hour decimal)"]
     )
 
-    i = 0
-    l = len(sorted_data)
-    curr_household = sorted_data.iloc[0, sorted_data.columns.get_loc("Household")]
-    curr_vehicle_id = sorted_data.iloc[0, sorted_data.columns.get_loc("Vehicle ID")]
-    sample_veh_num = 1
-    total_trips = 0
-    total_miles = 0
+    grouping = sorted_data.groupby(["Household", "Vehicle ID"])
+    sorted_data["sample vehicle number"] = grouping.ngroup() + 1
+    sorted_data["total vehicle trips"] = grouping["Vehicle miles traveled"].transform(
+        len
+    )
+    sorted_data["total vehicle miles traveled"] = grouping[
+        "Vehicle miles traveled"
+    ].transform(sum)
 
-    while i < l:
-        if (
-            curr_vehicle_id
-            == sorted_data.iloc[i, sorted_data.columns.get_loc("Vehicle ID")]
-            and curr_household
-            == sorted_data.iloc[i, sorted_data.columns.get_loc("Household")]
-        ):
-
-            total_trips += 1
-            total_miles += sorted_data.iloc[
-                i, sorted_data.columns.get_loc("Vehicle miles traveled")
-            ]
-
-            if total_trips > 1:
-                sorted_data.iloc[
-                    i - 1, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                ] = (
-                    sorted_data.iloc[
-                        i, sorted_data.columns.get_loc("Start time (hour decimal)")
-                    ]
-                    - sorted_data.iloc[
-                        i - 1, sorted_data.columns.get_loc("End time (hour decimal)")
-                    ]
-                )
-
-            # if the last entry is == the prev few entries, this makes sure they get added in;
-            # otherwise, it would break out of loop w/o adding
-            if i == (l - 1):
-                sorted_data.iloc[
-                    i - total_trips + 1 :,
-                    sorted_data.columns.get_loc("sample vehicle number"),
-                ] = sample_veh_num
-                sorted_data.iloc[
-                    i - total_trips + 1 :,
-                    sorted_data.columns.get_loc("total vehicle trips"),
-                ] = total_trips
-                sorted_data.iloc[
-                    i - total_trips + 1 :,
-                    sorted_data.columns.get_loc("total vehicle miles traveled"),
-                ] = total_miles
-
-                if total_trips == 1:
-                    sorted_data.iloc[
-                        i, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                    ] = 24 - (
-                        sorted_data.iloc[
-                            i - 1,
-                            sorted_data.columns.get_loc("Start time (hour decimal)"),
-                        ]
-                        - sorted_data.iloc[
-                            i - 1,
-                            sorted_data.columns.get_loc("End time (hour decimal)"),
-                        ]
-                    )
-                else:
-                    sorted_data.iloc[
-                        i, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                    ] = (
-                        24
-                        - sorted_data.iloc[
-                            i - 1,
-                            sorted_data.columns.get_loc("End time (hour decimal)"),
-                        ]
-                        + sorted_data.iloc[
-                            i - total_trips - 1,
-                            sorted_data.columns.get_loc("Start time (hour decimal)"),
-                        ]
-                    )
-
-        else:
-            sorted_data.iloc[
-                i - total_trips : i,
-                sorted_data.columns.get_loc("sample vehicle number"),
-            ] = sample_veh_num
-            sorted_data.iloc[
-                i - total_trips : i, sorted_data.columns.get_loc("total vehicle trips")
-            ] = total_trips
-            sorted_data.iloc[
-                i - total_trips : i,
-                sorted_data.columns.get_loc("total vehicle miles traveled"),
-            ] = total_miles
-
-            sample_veh_num += 1
-
-            if total_trips == 1:
-                sorted_data.iloc[
-                    i - 1, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                ] = 24 - (
-                    sorted_data.iloc[
-                        i - 1, sorted_data.columns.get_loc("Start time (hour decimal)")
-                    ]
-                    - sorted_data.iloc[
-                        i - 1, sorted_data.columns.get_loc("End time (hour decimal)")
-                    ]
-                )
-            else:
-                sorted_data.iloc[
-                    i - 1, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                ] = (
-                    24
-                    - sorted_data.iloc[
-                        i - 1, sorted_data.columns.get_loc("End time (hour decimal)")
-                    ]
-                    + sorted_data.iloc[
-                        i - total_trips - 1,
-                        sorted_data.columns.get_loc("Start time (hour decimal)"),
-                    ]
-                )
-
-            total_trips = 1
-            total_miles = sorted_data.iloc[
-                i, sorted_data.columns.get_loc("Vehicle miles traveled")
-            ]
-
-            curr_vehicle_id = sorted_data.iloc[
-                i, sorted_data.columns.get_loc("Vehicle ID")
-            ]
-            curr_household = sorted_data.iloc[
-                i, sorted_data.columns.get_loc("Household")
-            ]
-
-            # this makes sure the last entry gets included;
-            # otherwise, it would break out of loop w/o adding
-            if i == (l - 1):
-                sorted_data.iloc[
-                    i, sorted_data.columns.get_loc("total vehicle trips")
-                ] = total_trips
-                sorted_data.iloc[
-                    i, sorted_data.columns.get_loc("total vehicle miles traveled")
-                ] = total_miles
-                sorted_data.iloc[
-                    i, sorted_data.columns.get_loc("sample vehicle number")
-                ] = sample_veh_num
-
-                if total_trips == 1:
-                    sorted_data.iloc[
-                        i, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                    ] = 24 - (
-                        sorted_data.iloc[
-                            i - 1,
-                            sorted_data.columns.get_loc("Start time (hour decimal)"),
-                        ]
-                        - sorted_data.iloc[
-                            i - 1,
-                            sorted_data.columns.get_loc("End time (hour decimal)"),
-                        ]
-                    )
-                else:
-                    sorted_data.iloc[
-                        i, sorted_data.columns.get_loc("Dwell time (hour decimal)")
-                    ] = (
-                        24
-                        - sorted_data.iloc[
-                            i - 1,
-                            sorted_data.columns.get_loc("End time (hour decimal)"),
-                        ]
-                        + sorted_data.iloc[
-                            i - total_trips - 1,
-                            sorted_data.columns.get_loc("Start time (hour decimal)"),
-                        ]
-                    )
-
-        i += 1
+    # drop the 'household' and 'vehicle id' index
+    sorted_data["Dwell time (hour decimal)"] = grouping.apply(
+        calculate_dwell_time
+    ).droplevel([0, 1])
 
     return sorted_data
--- a/prereise/gather/demanddata/transportation_electrification/data_process.py
+++ b/prereise/gather/demanddata/transportation_electrification/data_process.py
@@ -51,13 +51,13 @@ def data_filtering(census_division):
     nhts_raw = load_data(census_division)
 
     # filter to be only vehicle trips (TRPTRANS values 1-6)
-    nhts = nhts_raw[nhts_raw.TRPTRANS.isin(range(1, 7))]
-
     # filter out repeated trips (VMT = -1)
-    nhts = nhts[nhts.VMT_MILE != -1]
-
     # get correct census division
-    nhts = nhts[nhts.CENSUS_D == census_division]
+    nhts = nhts_raw.loc[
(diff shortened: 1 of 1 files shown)


prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
...
+                   "Travel Minutes": nhts["TRVLCMIN"],
+                   "Dwell time": nhts["DWELTIME"],
+                   "Miles traveled": nhts["VMT_MILE"],          
+                   "Vehicle miles travelled": nhts["TRPMILES"], 
+                   "why from": nhts["WHYFROM"],
+                   "why to": nhts["WHYTO"],
+                   "Vehicle type": nhts["TRPTRANS"],            
+                   "Household vehicle count": nhts["HHVEHCNT"],
+                   "Household size": nhts["HHSIZE"]
+                   })
```
danielolsen, line 44: Since `nhts` is already a DataFrame, I think we can simplify this using the `rename` method. So instead of:
```python
sorted_data = pd.DataFrame({"oldname1": nhts["newname1"], "oldname2": nhts["newname2"], ...})
```
we could do something like:
```python
column_translations = {"oldname1": "newname1", "oldname2": "newname2", ...}
sorted_data = nhts[column_translations.keys()].rename(column_translations, axis="columns")
```

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
+import pandas as pd
+
+#Download CSV of NHTS 2017 from https://nhts.ornl.gov/
+
+def data_filtering(census_division, filepath: str = "trippub.csv"):
+    """Filter raw NHTS data to be used in mileage.py
+    
+    :param str filepath: filepath for the "trippub.csv" file in downloaded NHTS data
```
danielolsen, line 8: Missing docstring parameter for the `census_division` input.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
...
+                   "why to": nhts["WHYTO"],
+                   "Vehicle type": nhts["TRPTRANS"],            
+                   "Household vehicle count": nhts["HHVEHCNT"],
+                   "Household size": nhts["HHSIZE"]
+                   })
+    
+    #columns that require calculations: 21-28
+    sorted_data["Start time (hour decimal)"] = [row//100 + row % 100 / 60 for row in sorted_data["Trip start time (HHMM)"]]
+    sorted_data["End time (hour decimal)"] = [row//100 + row % 100 / 60 for row in sorted_data["Trip end time (HHMM)"]]
+    sorted_data['Dwell time (hour decimal)'] = ''
```
danielolsen, line 49: If all we're doing here is expanding the set of columns without filling meaningful data, then instead of:
```python
df["a"] = '',
df["b"] = '',
...
```
we can add multiple columns at once without them being set to string/object type with something like:
```python
df = df.reindex([df.columns + ["a", "b", ...], axis="columns")
```

danielolsen:
As a note for BE folks: the target for this PR is currently the `develop` branch, so that we can enable the CI linting, but we will want to change the target to the `transportation_electrification` branch before merging.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
...
+#Download CSV of NHTS 2017 from https://nhts.ornl.gov/
+
+def data_filtering(census_division, filepath: str = "trippub.csv"):
+    """Filter raw NHTS data to be used in mileage.py
+    
+    :param str filepath: filepath for the "trippub.csv" file in downloaded NHTS data
+    :return: (*pd.DataFrame*) -- filtered and sorted trip data
+    """
+
+    nhts_raw = pd.read_csv(filepath)
```
rouille, line 12: @BainanXia and @danielolsen, I believe we want this file on our blob storage, read from there and refer to it in the **ATTRIBUTION** file. I believe, we could have a year parameter that would get the file for a given year and raise an error if it is not available on the blob storage.

prereise/gather/demanddata/transportation_electrification/data_process.py
rouille: Or we can name overwrite column names using `header=0` and the `names` parameter in `read_csv`.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
...
+                if total_trips == 1:
+                    sorted_data.iloc[i, sorted_data.columns.get_loc('Dwell time (hour decimal)')] = 24 - \
+                        (sorted_data.iloc[i-1, sorted_data.columns.get_loc('Start time (hour decimal)')] - sorted_data.iloc[i-1, sorted_data.columns.get_loc('End time (hour decimal)')])
+                else:
+                    sorted_data.iloc[i, sorted_data.columns.get_loc('Dwell time (hour decimal)')] = 24 - \
+                        sorted_data.iloc[i-1, sorted_data.columns.get_loc('End time (hour decimal)')] + sorted_data.iloc[i-total_trips-1, sorted_data.columns.get_loc('Start time (hour decimal)')]
+
+        i += 1
+
+    sorted_data.to_excel("output.xlsx")
```
danielolsen, line 128: We probably want to make this step optional, since I don't think we use Excel files downstream.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
...
+
+    return sorted_data
+
+# data_filtering(1)
```
danielolsen, line 132: We'll want to remove this comment.

prereise/gather/demanddata/transportation_electrification/data_process.py
danielolsen: I think if we use the `names` parameter, we'll need to pass a list of names that cover all of the columns, even though we're only interested in a subset of them.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
+import pandas as pd
+
+#Download CSV of NHTS 2017 from https://nhts.ornl.gov/
+
+def data_filtering(census_division, filepath: str = "trippub.csv"):
```
danielolsen, line 5: If we refactor this function to take a dataframe instead of a filepath, it will be easier to write automated tests which operate only on objects in-memory.

prereise/gather/demanddata/transportation_electrification/data_process.py
rouille: And using `usecols` together with `names`? The nice thing with `usecols` is that it can take à `range` or integers.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,132 @@
...
+
+    i = 0
+    l = len(sorted_data)
+    curr_household = sorted_data.iloc[0, sorted_data.columns.get_loc('Household')]
+    curr_vehicle_id = sorted_data.iloc[0, sorted_data.columns.get_loc('Vehicle ID')]
+    sample_veh_num = 1
+    total_trips = 0
+    total_miles = 0
+
+    while i < l:
```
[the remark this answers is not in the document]
danielolsen: As @limingzhou2004 points out, we can also combine the calculation of group totals with the assignment back to individual rows:
```python
total_trips = grouping.size()
df["total_trips"] = df["class"].map(total_trips)
total_miles = grouping["miles"].sum()
df["total_miles"] = df["class"].map(total_miles)
```
can become:
```python
df["total_trips"] = grouping["miles"].transform(len)
df["total_miles"] = grouping["miles"].transform(sum)
```
This will avoid us having to define a series of `keys` values.

prereise/gather/demanddata/transportation_electrification/data_process.py
isond: This was actually more for me when testing things. I can go ahead and remove it

prereise/gather/demanddata/transportation_electrification/data_process.py
isond: I decided to go with Daniel's first suggestion since not all the columns in the original data set have an alternative name they could map to if we use `names`. But if I overlooked how to use it, let me know and I'll try to implement it!

danielolsen:
After the refactor, the code runs much faster: less than 4 minutes now to go through all nine regions on my laptop, rather than the hour and a half before. Bravo!

Comparing the results from before the refactor to the results afterward, I see a few differences:
- Some column renaming for spelling (`travelled` -> `traveled`)
- Different lengths of the dataframes for each region

| Region | Length before refactor | Length after refactor |
| ------- | ------------------------- | ----------------------- |
| 1 | 13,323 | 9,030 |
| 2 | 123,616 | 82,355 |
| 3 | 104,046 | 73,197 |
| 4 | 35,424 | 24,964 |
| 5 | 192,380 | 136,168 |
| 6 | 8,424 | 5,999 |
| 7 | 183,089 | 130,259 |
| 8 | 34,968 | 23,520 |
| 9 | 189,215 | 122,217 |

If I look at the indices from the newly-returned dataframes, and look at the corresponding values in the previously-returned dataframes, I see that most columns are the same, some columns have values that are different but nearly-identical: `Scaling Factor Applied`, `Miles traveled`, `Vehicle miles traveled`, `Start time (hour decimal)`, `End time (hour decimal)`, `Travel time (hour decimal)`, `Vehicle speed (mi/hour)` (probably differences due to floating-point rounding across the two different methods of summing), and some columns have values that are noticeably different: `Dwell time (hour decimal)`, `sample vehicle number`, `total vehicle trips`, and `total vehicle miles traveled`.

@isond can you confirm that the behavior change is intentional, and that the new behavior is the one we want?

One way to validate the new behavior could be to refactor the `data_filtering` function to take the raw NHTS dataframe as an explicit input, rather than to load it from disk. So instead of the user getting data as:
```python
df1 = data_filtering(census_division=1)
```
we could:
```python
raw_nhts = pd.read_csv(FILENAME)
df1 = data_filtering(raw_nhts, census_division=1)
```
The value of an approach like this is that we can design a test that passes a very small amount of data to `data_filtering`, and compares the output to a known expectation (calculated manually). Another benefit is that we can run the `data_filtering` function from anywhere, and we're less tied to specific filenames; we won't need the current working directory of the script to contain a file named `trippub.csv`.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,133 @@
...
+    :param str filepath: filepath for the "trippub.csv" file in downloaded NHTS data
+    :raises ValueError: if the census division is not between 1 and 9, inclusive.
+    :return: (*pd.DataFrame*) -- the data loaded from trippub.csv.
+    """
+    if not (1 <= census_division <= 9):
+        raise ValueError("census_region must be between 1 and 9 (inclusive).")
+
+    nhts_census = pd.read_csv(filepath)
+
+    df = pd.DataFrame(nhts_census)
```
danielolsen, line 19: `pd.read_csv` already returns a `DataFrame`, so we don't need the `df = pd.DataFrame(nhts_census)` line, we could just do:
```python
nhts_census = pd.read_csv(filepath)
return nhts_census
```
or even:
```python
return pd.read_csv(filepath)
```

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,133 @@
...
+# Download CSV of NHTS 2017 from https://nhts.ornl.gov/
+
+
+def load_data(census_division: int, filepath: str = "trippub.csv"):
+    """Load the data at trippub.csv.
+
+    :param int census_division: any of the 9 census regions defined by the US census bureau
+    :param str filepath: filepath for the "trippub.csv" file in downloaded NHTS data
+    :raises ValueError: if the census division is not between 1 and 9, inclusive.
+    :return: (*pd.DataFrame*) -- the data loaded from trippub.csv.
```
danielolsen, line 12: In docstrings and type annotations, we still want to use `pandas.DataFrame` rather than `pd.DataFrame`.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,133 @@
...
+    return dwells
+
+
+def data_filtering(census_division):
+    """Filter raw NHTS data to be used in mileage.py
+
+    :param int census_division: any of the 9 census regions defined by the US census bureau
+    :param str filepath: filepath for the "trippub.csv" file in downloaded NHTS data
+    :return: (*pd.DataFrame*) -- filtered and sorted trip data
+    """
```
danielolsen, line 49: Since this is the main user-facing function, we want to make sure that the docstring here has the right input parameters listed.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,133 @@
...
+    nhts_raw = load_data(census_division)
+
+    # filter to be only vehicle trips (TRPTRANS values 1-6)
+    nhts = nhts_raw[nhts_raw.TRPTRANS.isin(range(1, 7))]
+
+    # filter out repeated trips (VMT = -1)
+    nhts = nhts[nhts.VMT_MILE != -1]
+
+    # get correct census division
+    nhts = nhts[nhts.CENSUS_D == census_division]
```
danielolsen, line 60: Since each of these comparisons produces a binary array, they could be chained together as:
```python
nhts = nhts_raw.loc[
    nhts_raw.TRPTRANS.isin(range(1, 7))
    & (nhts.VMT_MILE != -1)
    & (nhts.CENSUS_D == census_division)
]
```
Both are equivalent though, so it's up to you which version you prefer.

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,133 @@
...
+        "WHYFROM": "why from",
+        "WHYTO": "why to",
+        "TRPTRANS": "Vehicle type",
+        "HHVEHCNT": "Household vehicle count",
+        "HHSIZE": "Household size",
+    }
+
+    sorted_data = nhts[column_translations.keys()].rename(
+        column_translations, axis="columns"
+    )
```
danielolsen, line 86: I'm not sure I understand this naming: are we actually sorting the data?

isond:
> After the refactor, the code runs much faster: less than 4 minutes now to go through all nine regions on my laptop, rather than the hour and a half before. Bravo!
> 
> Comparing the results from before the refactor to the results afterward, I see a few differences:
> 
> * Some column renaming for spelling (`travelled` -> `traveled`)
> * Different lengths of the dataframes for each region
> 
> Region	Length before refactor	Length after refactor
> 1	13,323	9,030
> 2	123,616	82,355
> 3	104,046	73,197
> 4	35,424	24,964
> 5	192,380	136,168
> 6	8,424	5,999
> 7	183,089	130,259
> 8	34,968	23,520
> 9	189,215	122,217
> If I look at the indices from the newly-returned dataframes, and look at the corresponding values in the previously-returned dataframes, I see that most columns are the same, some columns have values that are different but nearly-identical: `Scaling Factor Applied`, `Miles traveled`, `Vehicle miles traveled`, `Start time (hour decimal)`, `End time (hour decimal)`, `Travel time (hour decimal)`, `Vehicle speed (mi/hour)` (probably differences due to floating-point rounding across the two different methods of summing), and some columns have values that are noticeably different: `Dwell time (hour decimal)`, `sample vehicle number`, `total vehicle trips`, and `total vehicle miles traveled`.
> 
> @isond can you confirm that the behavior change is intentional, and that the new behavior is the one we want?
> 
> One way to validate the new behavior could be to refactor the `data_filtering` function to take the raw NHTS dataframe as an explicit input, rather than to load it from disk. So instead of the user getting data as:
> 
> ```python
> df1 = data_filtering(census_division=1)
> ```
> 
> we could:
> 
> ```python
> raw_nhts = pd.read_csv(FILENAME)
> df1 = data_filtering(raw_nhts, census_division=1)
> ```
> 
> The value of an approach like this is that we can design a test that passes a very small amount of data to `data_filtering`, and compares the output to a known expectation (calculated manually). Another benefit is that we can run the `data_filtering` function from anywhere, and we're less tied to specific filenames; we won't need the current working directory of the script to contain a file named `trippub.csv`.

Regarding the differences in lengths of the dataframes, I noticed that the condition to filter out the repeated trips was using the incorrect column prior to the refactor. 

`nhts = nhts[nhts.TRPMILES != -1]` should be `nhts = nhts[nhts.VMT_MILE != -1]`

Basically in the original code, the two columns were switched around, and so I've updated the code to reflect that change. 
I manually filtered out the columns in the `trippub.csv` because I wanted to double-check the length of the results from the dataframes after the refactor, and they all match up for each region. So the lengths that you are seeing now should be good :)

The significant change in values for `Dwell time (hour decimal)`, `sample vehicle number`, `total vehicle trips`, and `total vehicle miles traveled` is due to the fact that they used `vehicle miles traveled` (originally using values from `TRPMILES` but is now using `VMT_MILE`) which was affected by the change I previously described.

As for the spelling change, I updated the "traveled" variable to be consistent with the variables listed in Kate's data conversion document.

prereise/gather/demanddata/transportation_electrification/data_process.py
isond: I guess it's more like filtering the data; hence the function name haha

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -0,0 +1,113 @@
...
+        + data["Start time (hour decimal)"].iloc[0]
+    )
+
+    return dwells
+
+
+def data_filtering(raw_nhts, census_division):
+    """Filter raw NHTS data to be used in mileage.py
+
+    :param: (*pandas.DataFrame*) raw_nhts: raw NHTS dataframe
```
danielolsen: `:param: (*pandas.DataFrame*) raw_nhts:` -> `:param pandas.DataFrame raw_nhts:`. We only add the parentheses for the return variable.

danielolsen:
All checks pass, and  the code runs successfully and produces dataframes of the appropriate length, in 2 minutes:
```python
import pandas as pd
from prereise.gather.demanddata.transportation_electrification.data_process import data_filtering
raw_nhts = pd.read_csv("trippub.csv")
data = {i: data_filtering(raw_nhts, i) for i in range(1, 10)}
```

There's one small tweak that needs to be made to a docstring, and the branch needs to be rebased onto the latest `transportation_electrification` commit, and then we're ready to merge!

danielolsen (approved):
Looks good to me!

Issue #227, opened 2021-09-21
Add scripts to generate profiles for HIFLD grid


danielolsen:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
Similar to what was done for the TAMU Grid, we want scripts that can ingest data from external sources and produce profiles formatted for intake by PowerSimData/REISE.jl.

danielolsen:
Using some work-in-progress branches, I've got a workflow that appears to be working at least for the generation of wind profiles:
1. Begin with a pipenv virtual environment that contains both the `daniel/read_hifld_grid` branch of PowerSimData and the `daniel/decouple_profiles_from_grid` branch of PreREISE, and with a folder of grid CSVs created by the `hifld` branch of PreREISE (we'll call this folder `hifld_csvs` for this example).
2. Within bash, copy the HIFLD CSVs into the appropriate location within the installed version of PowerSimData so that they can be read to create an HIFLD Grid object:
```bash
mkdir $(pipenv --venv)/src/powersimdata/powersimdata/network/hifld/data
cp -r test_csvs/. $(pipenv --venv)/src/powersimdata/powersimdata/network/hifld/data/
```
3. Start a python session and follow the [HRRR demo](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/gather/winddata/hrrr/demo/hrrr_demo.ipynb) to download wind speed data and convert it to power profiles:
```python
from datetime import datetime
from powersimdata import Grid
from prereise.gather.winddata.hrrr.hrrr import retrieve_data
from prereise.gather.winddata.hrrr.calculations import calculate_pout

start_dt = datetime.fromisoformat("2020-01-01")
end_dt = datetime.fromisoformat("2021-01-01")
directory = "./"

grid = Grid("USA", "hifld")
wind_farms = grid.plant.query("type == 'wind' or type == 'wind_offshore'")
retrieve_data(start_dt=start_dt, end_dt=end_dt, directory=directory)
df = calculate_pout(wind_farms=wind_farms, start_dt=start_dt, end_dt=end_dt, directory=directory)
df.to_csv("wind.csv")
```

Downloading the wind data will take several hours, I'll update this post once it's done and I can confirm whether the method worked all the way through.

danielolsen:
Getting solar data can follow a similar pattern. Repeat steps 1 and 2 from wind data procedure above, then:
```python
from powersimdata import Grid
[...]

PR #245, opened 2021-12-06
fix: handle empty HRRR data files via linear imputation

### LINKED ISSUE #244: Any missing HRRR data causes calculation of power output to fail
# :beetle:

- [x] I have checked that this issue has not already been reported.

### Bug summary
If any of the hourly GRIB files downloaded by `prereise.gather.winddata.hrrr.hrrr.retrieve_data` are empty, calling `prereise.gather.winddata.hrrr.calculations.calculate_pout` will result in an error.

### Code for reproduction
The process followed what's shown in https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/gather/winddata/hrrr/demo/hrrr_demo.ipynb. For the 2020 year, I got four hours with no data: July 22 at 17:00, August 04 at 16:00, September 29 at 0:300, and Nov 12 at 09:00.

### Actual outcome
The output produced by the above code, which may be a screenshot, console output, etc.
```python
  File "/PreREISE/prereise/gather/winddata/hrrr/calculations.py", line 102, in calculate_pout
    u_component = gribs.select(name=U_COMPONENT_SELECTOR)[0].values.flatten()
  File "src/pygrib/_pygrib.pyx", line 538, in pygrib._pygrib.open.select
ValueError: no matches found
```

### Expected outcome
Some method of imputation, like what's done when generating wind profiles using the RAP method.

### Environment
Please specify your platform and versions of the relevant libraries you are using:
* Operating system: Linux
* PreREISE revision: the current `daniel/decouple_profiles_from_grid` branch
* Python version: the current `daniel/read_hifld_grid` branch

danielolsen:

### Purpose
When generating wind power profiles from HRRR data, gracefully handle any missing data via linear interpolation. Closes #244.

### What the code is doing
The `impute` module is moved from `prereise.gather.winddata.rap.impute` to `prereise.gather.winddata.impute`, and a new linear interpolation method is added which should perform well on small data gaps.

Within `prereise.gather.winddata.hrrr.calculations`, `calculate_pout` is refactored to first build an array of all wind speed magnitudes obtained from the NOAA grib files (filling in NA when files are empty), then impute missing values as necessary, and finally convert wind speeds to wind powers.

### Testing
Unit tests still pass, and this has been tested end-to-end when generating 2020 wind power profiles for the HIFLD grid (see https://github.com/Breakthrough-Energy/PreREISE/issues/227#issuecomment-985745255). When downloading the 2020 data, there were four files which downloaded empty, even after several attempts, suggesting that the data are missing from the NOAA server.

### Usage Example/Visuals
```python
from datetime import datetime
from powersimdata import Grid
from prereise.gather.winddata.hrrr.hrrr import retrieve_data
from prereise.gather.winddata.hrrr.calculations import calculate_pout

start_dt = datetime.fromisoformat("2020-01-01")
end_dt = datetime.fromisoformat("2021-01-01")
directory = "./"

grid = Grid("USA", "hifld")
wind_farms = grid.plant.query("type == 'wind' or type == 'wind_offshore'").copy()
wind_farms["state_abv"] = wind_farms.zone_id.map(grid.model_immutables.zones["id2abv"])
retrieve_data(start_dt=start_dt, end_dt=end_dt, directory=directory)
df = calculate_pout(wind_farms=wind_farms, start_dt=start_dt, end_dt=end_dt, directory=directory)
```

### Time estimate
15-30 minutes.


jon-hagg (approved):
Makes sense

prereise/gather/winddata/impute.py
```diff
@@ -137,3 +137,26 @@ def gaussian(data, wind_farm, inplace=True, curve="state"):
...
+        a_nan = np.isnan(a)
+        if a_nan.sum() == 0:
+            continue
+        data_impute[a_nan, i] = np.interp(
+            a_nan.nonzero()[0],  # Indices of bad data
+            (~a_nan).nonzero()[0],  # Indices of good data
+            a[~a_nan],  # Values of good data
+        )
+    if not inplace:
+        return data_impute
```
BainanXia, line 162: I'm wondering whether we could use `pandas.DataFrame.interpolate(method="linear")` which can be done directly over the resultant data frame with `np.nan`?

danielolsen: I think you are right!

prereise/gather/winddata/hrrr/calculations.py
```diff
@@ -98,34 +99,42 @@ def calculate_pout(wind_farms, start_dt, end_dt, directory):
...
+        [
+            get_power(
+                turbine_power_curves,
+                state_power_curves,
+                wind_speed_data[i, j],
+                turbine_types[j],
+            )
+            for j in range(wind_farm_ct)
+        ]
+        for i, _ in tqdm(enumerate(dts))
```
rouille, line 136: Do we need the enumerate here? It seems we just need the length of the `dts`

prereise/gather/winddata/impute.py
rouille: That said it would be nice to have a `linear` function in the `impute` module so perhaps we can have a wrapper around `pandas.DataFrame.interpolate(method="linear")`

prereise/gather/winddata/hrrr/calculations.py
danielolsen: `enumerate()` vs. `range(len())`, which is more pythonic?

jon-hagg: `for i in tqdm(dts)`?

BainanXia: > 

I think we will need the index rather than the element of `dts` here.

jon-hagg: `tqdm(dts, total=len(dts))` would work, but not sure if that's better than `enumerate`/`range`. It's unclear to me if any is the most pythonic, but I think `enumerate` is fine

prereise/gather/winddata/impute.py
danielolsen: Done.

prereise/gather/winddata/impute.py
```diff
@@ -137,3 +137,17 @@ def gaussian(data, wind_farm, inplace=True, curve="state"):
...
+
+    :param numpy.array/pandas.DataFrame data: data to interpolate.
+    :param bool inplace: whether to modify the data inplace or return a modified copy.
+    :return: (*None/pandas.DataFrame*) -- if ``inplace`` is False, data frame with
+        missing entries imputed.
+    """
+    data_impute = data if inplace else data.copy()
+    data_impute[:] = pd.DataFrame(data_impute).interpolate()
+    if not inplace:
+        return data_impute
```
BainanXia, line 153: Not sure whether we can simply `return pd.DataFrame(data).interpolate(inplace=inplace)` here?

danielolsen: I don't think we can, since `pd.DataFrame(data)` creates a new object. That's why I have the `data_impute[:] = ...` line: so we can modify the data contained in this object without creating a new one.

danielolsen:
```python
import numpy as np
import pandas as pd
>>> a = np.array([[1, float("nan"), 4, 5], [0, float("nan"), float("nan"), 4]]).T
>>> a
array([[ 1.,  0.],
       [nan, nan],
       [ 4., nan],
       [ 5.,  4.]])
>>> pd.DataFrame(a).interpolate(inplace=True)
>>> a
array([[ 1.,  0.],
       [nan, nan],
       [ 4., nan],
       [ 5.,  4.]])
```

prereise/gather/winddata/hrrr/calculations.py
danielolsen: We could also refactor how we build `wind_speed_data`--as a dataframe instead of a numpy array--and then we could build `wind_power_data` using `apply` calls instead of list comprehensions.

BainanXia (approved):
Good catch. Thanks!

prereise/gather/winddata/hrrr/calculations.py
rouille: See https://github.com/Breakthrough-Energy/PreREISE/pull/221#discussion_r708509549.

danielolsen: Aha. Maybe I'll leave this alone for now then.

danielolsen: Done. I tried to strike a good balance between compactness and readability.

Issue #242, opened 2021-11-30
Add extra data columns to HIFLD grid CSVs beyond those required by PowerSimData


danielolsen:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
Add extra data columns to HIFLD grid CSVs beyond those required by PowerSimData.

### Describe your proposed implementation
The change can be as simple as expanding the list of columns that are written via the top-level orchestration function, or not filtering the columns at all.

### Additional context
The corresponding PowerSimData issue to be able to properly interpret CSVs with extra columns is https://github.com/Breakthrough-Energy/PowerSimData/issues/567.

danielolsen:
Adding extra data columns also helps with issues like https://github.com/Breakthrough-Energy/PreREISE/issues/227, since:
1. We can generate better solar and wind profiles if we can associate our plant table entries with the richer EIA unit information that's available for solar and wind plants within Form 860. Table 3.1 is generator data for all generators, Table 3.2 is additional wind-specific data, and Table 3.3 is additional solar-specific data.
2. Without adding extra columns, the only way (?) to be able to look up the corresponding entries after the grid is made is to encode both the Plant Code (int) and the Generator ID (str) as the plant table index, but
3. PowerSimData expects that the plant table index is integers, to be able to create new indices when adding generators, storage units, etc.

danielolsen:
I think this is effectively achieved via #256, since the `create_grid` function now returns tables with all columns, and optionally writes just the ones required for PowerSimData, so I think we can close this.

Issue #213, opened 2021-09-02
Decouple profile creation from PowerSimData Grid object


danielolsen:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish I could generate hydro/solar/wind profiles using an arbitrary collection of lat/lons (and maybe power levels for hydro), without requiring a Grid object and all of the other information it contains/requires.

### Describe your proposed implementation
Anything that currently takes a Grid object will only take the relevant information directly, rather than looking into the Grid object to obtain this information.

BainanXia:
Thinking about it carefully, I found this could be tricky. For the profiles that are generated using weather data + location, this can be easily decoupled. However, for profiles generated using other information, it could be tough. For example, generating hydro profiles, we reply on net demand, generating demand profiles, we reply on Pd, etc.

danielolsen:
Could we de-couple in stages? It seems like the wind and solar profiles would be fairly straightforward at least.

danielolsen:
Even for wind and solar, the current wind and solar profile generation scripts use some imports from `powersimdata.network.usa_tamu.constants` (e.g. to do aggregation of individual plant attributes to state sums, or to look up the appropriate state average for a given plant's state). This presents problems for generating HIFLD profiles which have different zone IDs and other elements within the model immutables. I think we'll want to refactor these scripts so that they look only at the dataframe(s) that are passed directly to the functions, and don't assume that all queries relate to any particular grid model.

rouille:
> Even for wind and solar, the current wind and solar profile generation scripts use some imports from `powersimdata.network.usa_tamu.constants` (e.g. to do aggregation of individual plant attributes to state sums, or to look up the appropriate state average for a given plant's state). This presents problems for generating HIFLD profiles which have different zone IDs and other elements within the model immutables. I think we'll want to refactor these scripts so that they look only at the dataframe(s) that are passed directly to the functions, and don't assume that all queries relate to any particular grid model.
> 
> EDIT: branch-in-progress is the `daniel/decouple_profiles_from_grid` branch.

It would be nice if we could find a way to throw an error if the meteorological data don't cover the passed plant data frame. The NSRDB and NOAA database won't work with the European grid and we want to let the user know that he cannot use this script for plants loacted in Europe. We can work with a bounding box defined in a **const.py** in `prereise/gather`

danielolsen:
[...]

PR #247, opened 2021-12-14
refactor/feat: decouple solar profile generation from Grid object, add individual calculation

danielolsen:

### Purpose
This partially addresses #213 by decoupling the solar profile creation from the Grid object and associated network-specific mappings. The user can now pass to the existing blended-state-attributes function: a `Grid` object (for convenience), or a dataframe of solar plants and a mapping of interconnections to state abbreviations (for maximum decoupling). A few miscellaneous unrelated changes also decouple other modules from the `usa_tamu` grid model.

This also introduces a new solar profile generation function which uses the individual attributes of each solar plant, since we have this information available if we start from the raw EIA data. The pre-existing function has been renamed from `retrieve_data` to `retrieve_data_blended`, and the new function is named `retrieve_data_individual`. I'm not married to any of these names, in case anyone else has better ideas.

A few more helper functions have been added to the `prereise.gather.solardata.nsrdb.sam` module to serve the shared logic of these two functions.

The underlying NREL data downloading module has been updated to include local caching of the downloaded NREL data files.

### What the code is doing
Within **nrel_api.py**: we add an optional `cache_dir` input, which if it is provided acts as a location to check for pre-downloaded files, and store newly-downloaded files. There's a new helper function `_build_filename` to translate queries to filenames, similar to the existing function to translate queries to URLs.

Within **sam.py**:
- There are new generalized helper functions `generate_timestamps_without_leap_day` and `calculate_power` to support the two new solar profile calculation functions.
- The existing `retrieve_data` function has been renamed to `retrieve_data_blended` and refactored to be more flexible with its inputs. If the user passes a Grid object, then all relevant information (including zone mappings) comes from this object. Otherwise, the user must pass a `solar_plant` dataframe with enough columns to derive relationships between zone IDs and states/interconnects, and a mapping of interconnect names to state abbreviation sets (`interconnect_to_state_abvs`). This mapping can't be reliably derived from the `solar_plant` dataframe, since there may be states without solar plants in the given dataframe, but these states should still be considered in interconnect averages when looking at the EIA data table.
- `retrieve_data` has also been refactored to no longer need the `to_reise` data reshaping step, by constructing a dictionary of arrays as we go, and converting these to a dataframe as the final step of the function.
- A new function `retrieve_data_individual` has been added, which expects extra data columns to be present in `solar_plant`, and uses these extra data columns to produce more precise profiles for each plant, considering its specific tracking type, inverter loading ratio (ILR), and tilt angle (for fixed-tilt systems).

Within all other files: miscellaneous updates to constants or docstrings, no changes to functionality.

### Testing
Unit tests seem to be failing sporadically because of a 502 Bad Gateway when trying to map buses within `prereise.gather.demanddata.eia.map_ba`, which I didn't touch. Not sure what's going on there.

Both functions have been tested manually on the new HIFLD grid, with both invocation methods of the blended profile generator.

### Usage Example/Visuals
The second function needs to be able to associate extra information from EIA Form 860 Table 3.3 (see https://github.com/Breakthrough-Energy/PreREISE/issues/242#issuecomment-993078516), so uses the CSVs created by the latest HIFLD-grid-generating code in the `daniel/solar_pv_only_plus_index` branch (message me and I'll point you to the files for testing), plus the PowerSimData branch which enables the HIFLD grid (https://github.com/Breakthrough-Energy/PowerSimData/pull/566). For clarity, all usage examples will use this same HIFLD grid model.

Generating blended profiles using just a `Grid` object:
```python
from powersimdata import Grid
from prereise.gather.solardata.nsrdb import sam
grid = Grid("USA", "hifld")
data = sam.retrieve_data_blended(
    YOUR_EMAIL,
    YOUR_NREL_API_KEY,
    grid=grid,
    year=2020,
    cache_dir=YOUR_CACHE_DIR,
)
```

Generating blended profiles using a dataframe and dictionary (athough both are derived from a Grid object in this example):
```python
from powersimdata import Grid
from prereise.gather.solardata.nsrdb import sam
grid = Grid("USA", "hifld")
mi = grid.model_immutables
solar_plant = grid.plant.query("type == 'solar'").copy()
solar_plant["state_abv"] = solar_plant["zone_id"].map(mi.zones["id2abv"])
interconnect_to_state_abvs = mi.zones["interconnect2abv"]
data = sam.retrieve_data_blended(
    YOUR_EMAIL,
    YOUR_NREL_API_KEY,
    solar_plant=solar_plant,
    interconnect_to_state_abvs=interconnect_to_state_abvs,
    year=2020,
    cache_dir=YOUR_CACHE_DIR,
)
```

Generating individual profiles using additional information within EIA's Form 860 Table 3.3
```python
import pandas as pd
from powersimdata import Grid
from prereise.gather.solardata.nsrdb import sam
grid = Grid("USA", "hifld")
solar_plant = grid.plant.query("type == 'solar'")

# Add a helper function to coerce column types
def floatify(x):
    try:
        return float(x)
    except ValueError:
        return float("nan")

# Read additional data table, and pre-process data
extra_solar_data = pd.read_csv("3_3_Solar_Y2019_Operable.csv")
boolean_columns = ["Single-Axis Tracking?", "Dual-Axis Tracking?", "Fixed Tilt?"]
float_columns = ["DC Net Capacity (MW)", "Nameplate Capacity (MW)", "Tilt Angle"]
for col in float_columns:
    extra_solar_data[col] = extra_solar_data[col].map(floatify)

extra_solar_data.index = extra_solar_data.apply(
    lambda x: f"{x['Plant Code']}_{x['Generator ID']}", axis=1
)
boolean_columns = ["Single-Axis Tracking?", "Dual-Axis Tracking?", "Fixed Tilt?"]
for col in boolean_columns:
    # 'Y' becomes True, anything else ('N', blank, etc) becomes False
    extra_solar_data[col] = extra_solar_data[col] == "Y"

# If more than one column is True, assume Fixed Tilt for the purpose of this example
bad_booleans = extra_solar_data.index[extra_solar_data[boolean_columns].sum(axis=1) != 1]
extra_solar_data.loc[bad_booleans, boolean_columns] = False
extra_solar_data.loc[bad_booleans, "Fixed Tilt?"] = True

# Join frame with additional data to the original dataframe, and process
joined = solar_plant.join(extra_solar_data, rsuffix="_extra")
data = sam.retrieve_data_individual(
    YOUR_EMAIL,
    YOUR_NREL_API_KEY,
    solar_plant=joined,
    year=2020,
    cache_dir=YOUR_CACHE_DIR,
)
```

### Time estimate
1 hour or more.


prereise/gather/helpers.py
```diff
@@ -46,10 +47,9 @@ def get_monthly_net_generation(state, eia_form_923, resource, hps=True):
     if not isinstance(resource, str):
         raise TypeError("resource must be a str")
 
-    if state not in abv2loadzone.keys():
+    if state not in abv2state.keys():
```
jon-hagg, line 50: You could optionally remove the `.keys()` here and in the `ValueError` (the `in` operator checks keys, and calling `set` on a dict gives back the keys). Not sure if that's less readable though.

prereise/gather/solardata/nsrdb/nrel_api.py
```diff
@@ -142,17 +159,37 @@ def download(url):
...
+                )
+            return Psm3Data(
+                float(lat), float(lon), float(tz), float(elevation), data_resource
             )
-        return Psm3Data(
-            float(lat), float(lon), float(tz), float(elevation), data_resource
-        )
+
+        Psm3Data.check_attrs(attributes)
+        if cache_dir is not None:
```
jon-hagg, line 177: I think this would be equivalent but with a few fewer LOC:
```
    # same as before
    except FileNotFoundError:
        pass
url = self._build_url(lat, lon, attributes, year, leap_day)
resp = download(url)
psm3_data = format_to_psm3data(resp)
if cache_dir is not None:
    with open(filepath, "wb") as f:
        pickle.dump(psm3_data, f)
return psm3_data

```

prereise/gather/helpers.py
danielolsen: Good catch, done.

prereise/gather/solardata/nsrdb/nrel_api.py
danielolsen: Nice! Done.

jon-hagg:
Since `retrieve_data` is renamed, we should update the usage [here](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/cli/data_sources/solar_data.py#L203). There is also some mocking in [test_solar_data.py](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/cli/data_sources/tests/test_solar_data.py), and a notebook which call this. Alternatively, we could keep the name and just add the `retrieve_data_individual` function as you have now. I'm good with either way.

danielolsen:
> Since `retrieve_data` is renamed, we should update the usage [here](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/cli/data_sources/solar_data.py#L203). There is also some mocking in [test_solar_data.py](https://github.com/Breakthrough-Energy/PreREISE/blob/develop/prereise/cli/data_sources/tests/test_solar_data.py), and a notebook which call this. Alternatively, we could keep the name and just add the `retrieve_data_individual` function as you have now. I'm good with either way.

I've updated the function name and behavior in the two places you mentioned, plus the demo notebook.

jon-hagg (approved):
This looks reasonable and safe to merge. I think it could be nice to have a demo notebook or some docs regarding the required inputs for the individual calculation (an issue or future PR or whatever is fine) since there is a bit of setup needed before being able to call it.

PR #249, opened 2021-12-15
refactor/feat: decouple HRRR wind profile generation from Grid object, add individual calculation

danielolsen:

### Purpose
This partially addresses #213 by decoupling the HRRR wind profile creation from the Grid object and associated network-specific mappings. Instead of assuming that the dataframe passed is from the `usa_tamu` grid, and using the zone-id-to-state-abbreviations from that grid model, we expect a dataframe with a `state_abv` column. This is simpler than what we need for #247, since we don't do interconnection-level blending for states without entries in the EIA wind data table (and therefore don't need grid-level data); we just need the state abbreviation for each wind farm.

This also introduces a new wind profile generation function which uses the individual attributes of each wind farm, since we have this information available if we start from the raw EIA data. The pre-existing function has been renamed from `calculate_pout` to `calculate_pout_blended`, and the new function is named `calculate_pout_individual`. I'm not married to any of these names, in case anyone else has better ideas.

In the process, I also made a couple of other minor changes to the `winddata` sub-module organization.

### What the code is doing
- the `power_curves` sub-module is moved from `prereise.gather.winddata.rap.power_curves` to `prereise.gather.winddata.power_curves`, since it's used within both `prereise.gather.winddata.rap` and `prereise.gather.winddata.hrrr`. Similarly, we add `prereise.gather.winddata.const` for constants that get used throughout the `prereise.gather.winddata` module.
- Within `prereise.gather.winddata.hrrr.calculations`, we:
  - move reading the downloaded grib files and extracting wind speed magnitude to its own function (`extract_wind_speed`)
  - refactor `calculate_pout_blended` to use the new `extract_wind_speed` function, and to get wind farm states directly from the input CSV, rather than assuming a `usa_tamu` grid and getting the mapping from there.
  - add a new function `calculate_pout_individual` to look at the wind-farm-specific attributes passed, build a shifted power curve for each wind farm (shifting based on the real hub height, so that we can use the power curve with NOAA's 80m wind speed data), cache this curve for any subsequent identical queries, and then use the curve to translate wind speed to power.

### Testing
Both functions have been tested manually on the new HIFLD grid. Existing unit tests pass, besides the unrelated 502 Bad Gateway error we've been seeing lately.

### Usage Example/Visuals
Existing blended version:
```python
from datetime import datetime
from powersimdata.input.grid import Grid
from prereise.gather.winddata.hrrr.hrrr import retrieve_data
from prereise.gather.winddata.hrrr.calculations import calculate_pout_blended
start_dt = datetime.fromisoformat("2020-01-01")
end_dt = datetime.fromisoformat("2021-01-01")
directory = "YOUR_DESIRED_DOWNLOAD_DIRECTORY"
grid = Grid("USA", "hifld")
wind_farms = grid.plant.query("type == 'wind' or type == 'wind_offshore'").copy()
wind_farms["state_abv"] = wind_farms.zone_id.map(grid.model_immutables.zones["id2abv"])
retrieve_data(start_dt=start_dt, end_dt=end_dt, directory=directory)
df = calculate_pout_blended(
    wind_farms=wind_farms,
    start_dt=start_dt,
    end_dt=end_dt,
    directory=directory,
)
```

New individual version:
```python
from datetime import datetime
import pandas as pd
from powersimdata.input.grid import Grid
from prereise.gather.winddata.hrrr.hrrr import retrieve_data
from prereise.gather.winddata.hrrr.calculations import calculate_pout_individual

start_dt = datetime.fromisoformat("2020-01-01")
end_dt = datetime.fromisoformat("2021-01-01")
directory = "YOUR_DESIRED_DOWNLOAD_DIRECTORY"
grid = Grid("USA", "hifld")
wind_farms = grid.plant.query("type == 'wind' or type == 'wind_offshore'")

# Read file of wind-farm-specific data
extra_wind_data = pd.read_csv("3_2_Wind_Y2019_Operable.csv")
extra_wind_data.index = extra_wind_data.apply(
    lambda x: f"{x['Plant Code']}_{x['Generator ID']}", axis=1
)

def floatify(x):
    try:
        return float(x)
    except ValueError:
        return float("nan")

# Pre-process wind-farm-specific-data, filling in defaults (262.467 ft = 80m)
extra_wind_data["Turbine Hub Height (Feet)"] = extra_wind_data["Turbine Hub Height (Feet)"].map(floatify).fillna(262.467)
extra_wind_data["Predominant Turbine Manufacturer"] = extra_wind_data["Predominant Turbine Manufacturer"].astype("string").fillna("")
extra_wind_data["Predominant Turbine Model Number"] = extra_wind_data["Predominant Turbine Model Number"].astype("string").fillna("")

joined = wind_farms.join(extra_wind_data, rsuffix="_extra")
retrieve_data(start_dt=start_dt, end_dt=end_dt, directory=directory)
df = calculate_pout_individual(
    wind_farms=joined,
    start_dt=start_dt,
    end_dt=end_dt,
    directory=directory,
)
```

### Time estimate
1 hour. This follows the pattern of #247, but with fewer wrinkles.

### DIFF
--- a/prereise/gather/winddata/hrrr/calculations.py
+++ b/prereise/gather/winddata/hrrr/calculations.py
@@ -7,7 +7,7 @@ from tqdm import tqdm
 
 from prereise.gather.winddata.hrrr.helpers import formatted_filename
 from prereise.gather.winddata.impute import linear
-from prereise.gather.winddata.rap.power_curves import (
+from prereise.gather.winddata.power_curves import (
     get_power,
     get_state_power_curves,
     get_turbine_power_curves,
--- a/prereise/gather/winddata/hrrr/calculations.py
+++ b/prereise/gather/winddata/hrrr/calculations.py
@@ -62,33 +62,20 @@ def find_closest_wind_grids(wind_farms, wind_data_lat_long):
     return indices
 
 
-def calculate_pout(wind_farms, start_dt, end_dt, directory):
-    """Calculate power output for wind farms based on hrrr data.
-    Function assumes that user has already called
-    :meth:`prereise.gather.winddata.hrrr.hrrr.retrieve_data` with the same
-    start_dt, end_dt, and directory.
+def extract_wind_speed(wind_farms, start_dt, end_dt, directory):
+    """Read wind speed from previously-downloaded files, and interpolate any gaps.
 
-    :param pandas.DataFrame wind_farms: plant data frame, plus 'state_abv' column.
+    :param pandas.DataFrame wind_farms: plant data frame.
     :param str start_dt: start date.
     :param str end_dt: end date (inclusive).
     :param str directory: directory where hrrr data is contained.
-    :return: (*pandas.Dataframe*) -- Pandas containing power out per wind farm
+    :return: (*pandas.Dataframe*) -- data frame containing wind speed per wind farm
         on a per hourly basis between start_dt and end_dt inclusive. Structure of
         dataframe is:
             wind_farm1  wind_farm2
-        dt1    POUT        POUT
-        dt2    POUT        POUT
-    :raises ValueError: if ``wind_farms`` is missing the 'state_abv' column.
+        dt1   speed       speed
+        dt2   speed       speed
     """
-
-    if "state_abv" not in wind_farms.columns:
-        raise ValueError("The wind_farms data frame must have a 'state_abv' column")
-    turbine_types = wind_farms.apply(
-        lambda x: "Offshore" if x["type"] == "wind_offshore" else x["state_abv"], axis=1
-    )
-
-    turbine_power_curves = get_turbine_power_curves()
-    state_power_curves = get_state_power_curves()
     wind_data_lat_long = get_wind_data_lat_long(start_dt, directory)
     wind_farm_to_closest_wind_grid_indices = find_closest_wind_grids(
         wind_farms, wind_data_lat_long
@@ -117,6 +104,42 @@ def calculate_pout(wind_farms, start_dt, end_dt, directory):
 
     # For each column, linearly interpolate any NaN values
     linear(wind_speed_data)
+
+    return wind_speed_data
+
+
+def calculate_pout(wind_farms, start_dt, end_dt, directory):
+    """Calculate power output for wind farms based on hrrr data.
+    Function assumes that user has already called
+    :meth:`prereise.gather.winddata.hrrr.hrrr.retrieve_data` with the same
+    start_dt, end_dt, and directory.
+
+    :param pandas.DataFrame wind_farms: plant data frame, plus 'state_abv' column.
+    :param str start_dt: start date.
+    :param str end_dt: end date (inclusive).
+    :param str directory: directory where hrrr data is contained.
+    :return: (*pandas.Dataframe*) -- Pandas containing power out per wind farm
+        on a per hourly basis between start_dt and end_dt inclusive. Structure of
+        dataframe is:
+            wind_farm1  wind_farm2
+        dt1    POUT        POUT
+        dt2    POUT        POUT
+    :raises ValueError: if ``wind_farms`` is missing the 'state_abv' column.
+    """
+
+    if "state_abv" not in wind_farms.columns:
+        raise ValueError("The wind_farms data frame must have a 'state_abv' column")
+    turbine_types = wind_farms.apply(
+        lambda x: "Offshore" if x["type"] == "wind_offshore" else x["state_abv"], axis=1
+    )
+
+    turbine_power_curves = get_turbine_power_curves()
+    state_power_curves = get_state_power_curves()
+
+    # Read wind speed from previously-downloaded files, and interpolate
+    wind_speed_data = extract_wind_speed(wind_farms, start_dt, end_dt, directory)
+    dts = wind_speed_data.index
+
     # Then calculate wind power based on wind speed
     wind_power_data = [
         [
--- a/prereise/gather/winddata/hrrr/calculations.py
+++ b/prereise/gather/winddata/hrrr/calculations.py
@@ -1,3 +1,5 @@
+import os
+
 import numpy as np
 import pandas as pd
 import pygrib
@@ -27,7 +29,7 @@ def get_wind_data_lat_long(dt, directory):
     :return: (*tuple*) -- A tuple of 2 same lengthed numpy arrays, first one being
         latitude and second one being longitude.
     """
-    gribs = pygrib.open(directory + formatted_filename(dt))
+    gribs = pygrib.open(os.path.join(directory, formatted_filename(dt)))
     grib = next(gribs)
     return grib.latlons()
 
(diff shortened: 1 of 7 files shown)


prereise/gather/winddata/hrrr/calculations.py
```diff
@@ -116,6 +108,44 @@ def calculate_pout(wind_farms, start_dt, end_dt, directory):
...
+    return wind_speed_data
+
+
+def calculate_pout_blended(wind_farms, start_dt, end_dt, directory):
+    """Calculate power output for wind farms based on hrrr data. Each wind farm's power
+    curve is based on the average power curve for that state, based on EIA data on the
+    state's turbines.
+    Function assumes that user has already called
+    :meth:`prereise.gather.winddata.hrrr.hrrr.retrieve_data` with the same
+    start_dt, end_dt, and directory.
```
BainanXia: start_dt -> ``start_dt``
end_dt -> ``end_dt``
directory -> ``directory``

prereise/gather/winddata/hrrr/calculations.py
```diff
@@ -116,6 +108,44 @@ def calculate_pout(wind_farms, start_dt, end_dt, directory):
...
+
+    :param pandas.DataFrame wind_farms: plant data frame, plus 'state_abv' column.
+    :param str start_dt: start date.
+    :param str end_dt: end date (inclusive).
+    :param str directory: directory where hrrr data is contained.
+    :return: (*pandas.Dataframe*) -- Pandas containing power out per wind farm
```
BainanXia: Pandas -> data frame

prereise/gather/winddata/hrrr/calculations.py
```diff
@@ -132,3 +162,81 @@ def calculate_pout(wind_farms, start_dt, end_dt, directory):
...
+        dt1    POUT        POUT
+        dt2    POUT        POUT
+    :raises ValueError: if ``wind_farms`` is missing the 'state_abv' column.
+    """
+
+    def get_shifted_curve(wind_farm_series, reference_curves, curve_cache):
+        """Get the power curve for the given turbine type, shifted to hub height. If the
+        turbine curve isn't found in ``reference_curves``, default to IEC class 2. Cache
+        results from identical queries.
+
```
BainanXia, line 193: Missing parameter doc strings here.

prereise/gather/winddata/hrrr/calculations.py
```diff
@@ -132,3 +162,81 @@ def calculate_pout(wind_farms, start_dt, end_dt, directory):
...
+        except KeyError:
+            try:
+                starting_curve = reference_curves[full_name]
+            except KeyError:
+                starting_curve = reference_curves["IEC class 2"]
+            shifted_curve = shift_turbine_curve(
+                starting_curve, hub_height, const.max_wind_speed, const.new_curve_res
+            )
+            curve_cache[lookup_key] = shifted_curve
+            return shifted_curve
```
BainanXia: Not sure whether using the built-in cache decorator gives us same behavior here:
```python
@cache
def get_shifted_curve(wind_farm_series, reference_curves):
        full_name = " ".join(
            [wind_farm_series[const.mfg_col], wind_farm_series[const.model_col]]
        )
        hub_height = wind_farm_series[const.hub_height_col]
        starting_curve = reference_curves[full_name] if full name in reference_curves else reference_curves["IEC class 2"]
        shifted_curve = shift_turbine_curve(
            starting_curve, hub_height, const.max_wind_speed, const.new_curve_res
        )
        return shifted_curve
```
We do need input parameters to be hashable though...

BainanXia (approved):
This is clean. Thanks

prereise/gather/winddata/hrrr/calculations.py
danielolsen: `@functools.cache`? TIL!

Unfortunately, pandas Series and DataFrames aren't hashable. Python doesn't seem to have a convenient hashable mapping class (although they're considering one in PEP 603), so if we want to pass the same inputs, I think we'd need to mock one up using tuples, and that might end up looking just as ugly as what we have currently.

BainanXia: Yeah, I presume `wind_farm_series` and `reference_curves` are either pandas series or data frame. One potential work around would be make `get_shifted_curve` take three parameters instead, i.e. full_name_first_part, full_name_second_part and height, then treat reference_curves as defined variables outside the function, i.e. use `turbine_power_curves` directly. Just random thoughts. I'm okay with the current implementation as well.

danielolsen: Another implementation of the same sort of thing would be using `functools.partial` to curry the unchanging reference curves into a function that only takes hashable inputs, so we don't need to rely on anything outside of local scope, and then calling the cache decorator function directly on the curried function. Something like:

```python
import functools

def calculate_pout_individual(wind_farms, start_dt, end_dt, directory):
    ....
    def get_shifted_curve(full_name, hub_height, reference_curves):
        try:
            starting_curve = reference_curves[full_name]
        except KeyError:
            starting_curve = reference_curves["IEC class 2"]
        return shift_turbine_curve(
            starting_curve, hub_height, const.max_wind_speed, const.new_curve_res
        )

    turbine_power_curves = get_turbine_power_curves()
    partial_func = functools.partial(get_shifted_curve, reference_curves=turbine_power_curves)
    cached_func = functools.cache(partial_func)
    shifted_power_curves = wind_farms.apply(
        lambda x: cached_func(f"{x[const.mfg_col]} {x[const.model_col]}", x[const.hub_height_col]),
        axis=1,
    )
    ...
```

BainanXia: TIL!

danielolsen: Or maybe to be a bit clearer, and to achieve caching of curves with the same hub height and no curve in the reference curves:

```python
import functools

def calculate_pout_individual(wind_farms, start_dt, end_dt, directory):
    ...
    def get_starting_curve_name(mfg_name, model_name, valid_names):
        full_name = " ".join([mfg_name, model_name])
        return full name if full_name in valid_names else "IEC class 2"

    def get_shifted_curve(lookup_name, hub_height, reference_curves):
        return shift_turbine_curve(
            reference_curves[lookup_name],
            hub_height,
            const.max_wind_speed,
            const.new_curve_res,
        )

    # Create cached, curried function for use with apply
    turbine_power_curves = get_turbine_power_curves()
    partial_func = functools.partial(get_shifted_curve, reference_curves=turbine_power_curves)
    cached_func = functools.cache(partial_func)
    # Create dataframe of lookup values
    lookup_names = wind_farms.apply(
        lambda x: get_starting_curve_name(
            x[const.mfg_col], x[const.model_col], turbine_power_curves.columns
        ),
        axis=1,
    )
    lookup_values = pandas.concat([lookup_names, wind_farms[const.hub_height_col]], axis=1)
    # Use lookup values with cached, curried function
    shifted_power_curves = lookup_values.apply(lambda x: cached_func(*x), axis=1)
    ...
```

prereise/gather/winddata/hrrr/calculations.py
danielolsen: Done

prereise/gather/winddata/hrrr/calculations.py
danielolsen: Done

prereise/gather/winddata/hrrr/calculations.py
danielolsen: Done

prereise/gather/winddata/hrrr/calculations.py
danielolsen: One final note: `@functools.cache(myfunc)` isn't available until Python 3.9, so for Python 3.8 compatibility we need the apparently-equivalent `@functools.lru_cache(maxsize=None)(myfunc)` instead.

BainanXia: Yeah, I should have put this note down as well. The `cache` decorator is something I learned when my compiler is complaining that `@lru_cache` needs to specify a size parameter or call as a function, i.e. `@lru_cache()`. Note that if the size of `lru_cache` binds, the behavior deviates from `cache`.

PR #251, opened 2022-01-05
chore: narrow dependencies, bump version number to prepare for v0.4.1 release

danielolsen:

### Purpose
Prepare for a v0.4.1 release, which will be the final release compatible with Python 3.7.

### What the code is doing
- Numpy and scipy dependency versions are made more specific, so that they don't grab versions which don't support Python 3.7.
- The version number for PreREISE is updated to v0.4.1

### Time estimate
30 seconds.


rouille (approved):
Thanks

BainanXia (approved):

PR #255, opened 2022-01-05
Update EFS profile creator with new flexibility headers and user-specified regions

lanesmith:

### Purpose
The purpose of this PR is to update the process for creating demand and demand flexibility profiles from NREL Electrification Futures Study (EFS) data. The process now creates demand flexibility profiles that have the appropriate zone headers, as is specified in the most up-to-date version of `REISE.jl`. The process now also allows users to specify which combination of interconnections and states they would like included in the profiles. The demo notebooks were also updated accordingly to showcase the new functionality.

### What the code is doing
Most of the changes were implemented in the `decompose_demand_profile_by_state_to_loadzone` function in `map_states.py`. To create the proper zone headers for demand flexibility profiles, `decompose_demand_profile_by_state_to_loadzone` now needs to take the profile type (`"demand"` or `"demand_flexibility"`) as input; this is due to the fact that the new zone header convention is only for demand flexibility profiles, but not for all other profiles (including demand). If the profile type is `"demand_flexibility"`, then the column headers of the DataFrame are updated to inlcude `"zone."` as a prefix to all original column headers. To allow users the ability to specify a desired combination of interconnections and states, `decompose_demand_profile_by_state_to_loadzone` now has another input where users can specify a set or list of their desired interconnection names and/or state abbreviations. `None` is the default input (following the immutable default parameter convention), which allows the full USA profile to be returned. Based on the set or list provided by the user, the corresponding load zones are used to pare down the load zones included in the demand or demand flexibility profile. Both new input variables include checks to verify the validity of the user's inputs.

In addition to the above functionality, there was also an issue with how `pandas` now compares an `Index` and a `DatetimeIndex`. Essentially, identical indices would return as not equal due to one being an `Index` and the other being a `DatetimeIndex`. This problem was observed when loading in a profile and trying to verify its timestamps with an artificially created `DatetimeIndex`. This issue was fixed in `aggregate_demand.py`.

### Testing
The new functionality was primarily tested using the demo notebooks and by running local tests that iterate through various scenarios. `test_map_states.py` was also updated in accordance with the need to specify which type of profile is being created (i.e., `"demand"` or `"demand_flexibility"`). All tests pass.

### Where to look
The main code changes live in `PreREISE/prereise/gather/demanddata/nrel_efs/map_states.py`. Small changes were made to `PreREISE/prereise/gather/demanddata/nrel_efs/aggregate_demand.py` and `PreREISE/prereise/gather/demanddata/nrel_efs/tests/test_map_states.py`. The two updated demand notebooks can be found at `PreREISE/prereise/gather/demanddata/nrel_efs/demo/efs_demand_reference_slow_2030_demo.ipynb` and `PreREISE/prereise/gather/demanddata/nrel_efs/demo/efs_base_flexibility_reference_slow_2030_demo.ipynb`.

### Usage Example/Visuals
See the demo notebooks.

### Time estimate
I don't anticipate this taking long to review. Most of the listed differences are due to changes to the two demo notebooks.


prereise/gather/demanddata/nrel_efs/aggregate_demand.py
```diff
@@ -129,12 +129,9 @@ def access_non_efs_demand(dem_paths):
...
     for i in dem_paths:
         # Try loading the locally-stored sectoral demand
-        temp_dem = pd.read_csv(i)
-
-        # Set the index
-        if "Local Time" in temp_dem.columns:
-            temp_dem.set_index("Local Time", inplace=True)
-        else:
+        try:
+            temp_dem = pd.read_csv(i, parse_dates=True, index_col="Local Time")
```
BainanXia, line 133: Nice. New stuff learned:)

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -2,29 +2,56 @@
...
+    df, profile_type, regions=None, save=None
+):
     """Transforms the sectoral demand data so that it is separated by load zone rather
     than by state.
 
     :param pandas.DataFrame df: DataFrame of the sectoral demand data, where the rows
         are time steps (in local time) and the columns are the states. This input is
         intended to be the output of :py:func:`combine_efs_demand` or the components
         that are output from :py:func:`partition_flexibility_by_sector`.
+    :param str profile_type: A str that identifies the type of profile that is provided.
```
BainanXia: A str that ... -> A string that ...

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -2,29 +2,56 @@
...
+        Can be one of *'demand'* or *'demand_flexibility'*.
+    :param set/list regions: The combination of interconnection names and state
```
BainanXia: Maybe we can put `iterable` here?

lanesmith: I'm not sure I follow where you are suggesting that be added. Are you suggesting that replace `set/list`?

BainanXia: Yeah, I remember previously we did replace `list/set/tuple` by `iterable`.

lanesmith: Okay, will do!

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -52,6 +101,27 @@ def decompose_demand_profile_by_state_to_loadzone(df, save=None):
     # Convert from local hours to UTC time
     df_lz = shift_local_time_by_loadzone_to_utc(df_lz)
 
+    # Create mapping from state abbreviations to loadzone IDs
+    abv2id = {}
+    for k, v in id2abv.items():
+        abv2id.setdefault(v, []).append(k)
```
BainanXia: Or we could define `abv2id` as a `defaultdict(list)` here.

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -52,6 +101,27 @@ def decompose_demand_profile_by_state_to_loadzone(df, save=None):
...
+
+    # Determine the loadzones to be inlcuded in the profile, as specified by regions
+    loadzones = set()
+    for region in regions:
+        if len(region) == 2:
+            loadzones = loadzones | set(abv2id[region])
+        else:
+            loadzones = loadzones | interconnect2id[region]
+    loadzones = list(sorted(loadzones))
```
BainanXia: I think `sorted` will return a list by default, so maybe we don't need the `list` conversion here. Also, we might not need to sort it at all, not sure where the order matters here or not.

lanesmith: Good catch, thanks!

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -52,6 +101,27 @@ def decompose_demand_profile_by_state_to_loadzone(df, save=None):
...
+    for k, v in id2abv.items():
+        abv2id.setdefault(v, []).append(k)
+
+    # Determine the loadzones to be inlcuded in the profile, as specified by regions
+    loadzones = set()
+    for region in regions:
+        if len(region) == 2:
+            loadzones = loadzones | set(abv2id[region])
+        else:
+            loadzones = loadzones | interconnect2id[region]
```
BainanXia: Not sure whether this works in the same way or will fit in one line.
```python
loadzones = set().union(set(abv2id.get(r, {})) | interconnect2id.get(r, {}) for r in regions)
```

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Yeah, I can change it it use `defaultdict` instead. Is there a reason this is better? Also, this mapping doesn't appear to be needed elsewhere, which is why I needed to create it here. That being said, should this be added to all of the other mappings in `powersimdata\network\usa_tamu\constants\zones.py`?

prereise/gather/demanddata/nrel_efs/map_states.py
```diff
@@ -52,6 +101,27 @@ def decompose_demand_profile_by_state_to_loadzone(df, save=None):
...
+    loadzones = list(sorted(loadzones))
+
+    # Keep the appropriate loadzones
+    df_lz = df_lz[loadzones]
+
+    # Change the column headers if the profile_type is "demand_flexibility"
+    if profile_type == "demand_flexibility":
+        df_lz.columns = ["zone." + str(x) for x in loadzones]
```
BainanXia: Maybe we can simply do `f"zone.{x}"` here?

prereise/gather/demanddata/nrel_efs/map_states.py
BainanXia: Good idea, let's add it in.

BainanXia (approved):
All minor concerns. LGTM.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Okay, I'll add it in there instead.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: After looking at this for a bit, it appears that sets can't be added in this way. When you try to do so, the following error is produced: `TypeError: unhashable type: 'set'`. You see this same error even in the following similar dummy implementation:

```
a = [{1, 2, 3}, {3, 4}]
b = set()
b = b.union(i for i in a)
```

So it would appear that one-line iterative unions of sets don't work in the same way as one-line iterative summations of elements in a list.

BainanXia: Yeah, you are right. There is a work around as following, probably it isn't worth the effort though (I'm getting nerdy here)
```python
loadzones = set().union(*[set(abv2id.get(r, {})) | interconnect2id.get(r, {}) for r in regions]))

or 

from itertools import chain
loadzones = set(chain(*[set(abv2id.get(r, {})) | interconnect2id.get(r, {}) for r in regions]))
```
It's up to you. I'm okay with the current implementation as well.

lanesmith: That's really snazzy! I wasn't aware of the star operator. I'll implement this using that method. Thanks!

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Done.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Done.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Done.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Done.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Done.

prereise/gather/demanddata/nrel_efs/map_states.py
lanesmith: Done.

lanesmith:
It looks like the Pytests are failing because it isn't recognizing the new `abv2id` mapping that I added to `PowerSimData` (see [`PowerSimData` PR #568](https://github.com/Breakthrough-Energy/PowerSimData/pull/568)). I merged it in to `PowerSimData` a couple hours ago, but the test that is failing (`test_map_states.py`) is passing for me locally. Am I missing something?

danielolsen:
> It looks like the Pytests are failing because it isn't recognizing the new `abv2id` mapping that I added to `PowerSimData` (see [`PowerSimData` PR #568](https://github.com/Breakthrough-Energy/PowerSimData/pull/568)). I merged it in to `PowerSimData` a couple hours ago, but the test that is failing (`test_map_states.py`) is passing for me locally. Am I missing something?

I wonder if it's because the PreREISE CI is grabbing the latest _release_ of powersimdata (v0.4.4), not the latest copy of `develop`.

lanesmith:
> I wonder if it's because the PreREISE CI is grabbing the latest _release_ of powersimdata (v0.4.4), not the latest copy of `develop`.

Yeah, I think you're right. Looking into the pytest setup, it looks like it references `Pipfile.lock` to install the correct dependencies. In `PreREISE`, `PowerSimData` v0.4.4 is the requirement, meaning that the new change to `PowerSimData` would not be recognized. Since this isn't necessarily critical functionality, is the best course of action to wait on this PR until there is a version bump in `PowerSimData`?

PR #256, opened 2022-01-07
feat: generate solar profiles during HIFLD grid-building

danielolsen:
### Purpose
- Generate solar profiles at the same time as we build the HIFLD grid, so that we can leverage the plant-specific attributes of each solar farm to generate better profiles (partial fulfillment of #227).
- Add a module to generate profiles for concentrating solar power (CSP) plants, which represent about 5% of total solar capacity (closes #248).
- ~Simplify solar profile generation process, now that SAM can handle leap years.~ EDIT: this no longer appears possible, see https://github.com/Breakthrough-Energy/PreREISE/issues/254#issuecomment-1007800590
- Update the version of NREL's PySAM used, for compatibility with Python 3.10, and test using Python 3.10.

### What the code is doing
- Within `Pipfile`, `Pipfile.lock`, and `requirements.txt`: updating the version of PySAM. Then, we add Python 3.10 to the test matrix in `.github/workflows/test.yml`.
- Within `const.py`, we add references to the tables of 'extra' EIA data in the blob storage.
- Within `generators.py`, we make sure that the plant index gets named (unrelated to main objectives).
- We add a new module `prereise.gather.griddata.hifld.data_process.profiles`, which takes the 'extra' data table for solar, joins it with the solar plants from the plant dataframe, and passes this (with some pre-processing) to individualized profile generation function introduced in #247. Most of this new code is copy-and-pasted from the usage examples in that PR.
- Within `orchestration.py`, we call the new function from `prereise.gather.griddata.hifld.data_process.profiles`, and then save the resulting profiles to CSV at the same time as we save the rest of the Grid CSVs.
- Within `nrel_api.py`:
  - we add a method that converts the `Psm3Data` data structure back to the file format expected by SAM. For whatever reason, the way that weather data are attached to a `PySAM.PVWatts` object is different than how they're attached to a `PySAM.TcsgenericSolar` object: `TcsgenericSolar` needs a filename for a file formatted in a particular way.
  - we make the downloading more resilient to server-side errors.
- Within `sam.py`:
  - ~we remove the leap-year-specific logic, now that SAM handles lear years properly.~
  - we query the full year's weather data during leap years, since SAM can take it as input, but we still need to manually copy the Feb 28th solar data to leap days, since SAM doesn't produce power output for Feb 29th.
[...]

PR #260, opened 2022-01-15
feat: generate wind profiles during HIFLD grid-building

danielolsen:

### Purpose
Generate wind profiles at the same time as we build the HIFLD grid, so that we can leverage the plant-specific attributes of each wind farm to generate better profiles (partial fulfillment of #227).

### What the code is doing
- `build_wind` joins the wind-specific data table with the wind plants in the main plant data frame, then calls the individual calculation introduced in #249.
- `create_csvs` is extended with the additional inputs needed by `build_wind`.

### Testing
Tested manually: profiles are created which match the expected timespan, and columns match the indices of wind generators.

### Usage Example/Visuals
```python
def create_csvs(
    DESIRED_OUTPUT_CSV_DIRECTORY,
    wind_directory=YOUR_WIND_CACHE_DIR,
    year=2020,
    YOUR_NREL_EMAIL,
    YOUR_NREL_API_KEY,
    solar_kwargs={"cache_dir": YOUR_SOLAR_CACHE_DIR},
):
```

### Time estimate
30 minutes. This is a draft PR since it depends on #256, but has been tested and performs as expected.

### DIFF
--- a/prereise/gather/griddata/hifld/data_process/profiles.py
+++ b/prereise/gather/griddata/hifld/data_process/profiles.py
@@ -1,6 +1,10 @@
+import datetime as dt
+
 from prereise.gather.griddata.hifld import const
 from prereise.gather.griddata.hifld.data_access.load import get_eia_form_860
 from prereise.gather.solardata.nsrdb.sam import retrieve_data_individual
+from prereise.gather.winddata.hrrr.calculations import calculate_pout_individual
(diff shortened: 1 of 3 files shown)


danielolsen:
This has been rebased and re-tested, so is ready for a full review.

kasparm (approved):
Works as expected. Logic looks good.
Thank you!

PR #262, opened 2022-01-24
feat: add module to calculate transmission impedances from geometry

danielolsen:

### Purpose
Add a module to calculate transmission impedances from geometry. This will let us develop impedance values for 'standard' tower designs, which can be used with assumptions about tower design per voltage level to develop impedance values for the HIFLD transmission lines.

### What the code is doing
Four new classes are added. Each has the `dataclasses.dataclass` decorator, which automatically give them an `__init__` method that specifies inputs and also calls a `__post_init__` method (if one is available).
- The `Conductor` class represents a single conductor, and can perform estimations of relevant parameters as necessary. This probably won't be used much if we get parameter values from tables or spec sheets, but is included for completeness.
- The `ConductorBundle` class represents a bundle of conductors, and performs calculations of equivalent inductive and capacitive spacings. These formulas are purely geometric. Both flat or circular bundles of up to four conductors each are supported.
- The `PhaseLocations` class represents the geometry of a single-circuit tower: the (x, y) locations of each phase.
- The `Tower` class combines a `ConductorBundle` object and a `PhaseLocations` object to calculate per-kilometer impedance values. The formulas here are from _Power System Analysis_, by Grainger and Stevenson.

### Testing
Tests are added for each major class:
- `Conductor` calculations check that resistances can be approximately calculated.
- `ConductorBundle` calculations check bundle geometry.
- `Tower` calculations check tower geometry. The values tested are those presented in _A Methodology for the Creation of Geographically Realistic Synthetic Power Flow Models_; series reactance and shunt admittance match to within 1%, resistance matches to within 5% (resistivity varies significantly by temperature, and the paper doesn't list the temperature or resistivity used in their calculations).

### Usage Example/Visuals
The tests demonstrate how to instantiate these objects. Once instantiated, all relevant parameters are contained as object attributes.

### Time estimate
30-60 minutes: the calculations are all straightforward, once the formulas are known. ~This is currently a draft PR since I plan to add docstrings and per-unit calculations ~and calculations for double-circuit towers, but the logic and design for single-circuit~ nominal values are ready for review.~ **EDIT**: all the remaining to-do list items have been completed.

### DIFF
--- a/prereise/gather/griddata/transmission/geometry.py
+++ b/prereise/gather/griddata/transmission/geometry.py
@@ -1,0 +1,282 @@
+from dataclasses import dataclass, field
+from itertools import combinations
+from math import exp, log, pi, sqrt
+from statistics import geometric_mean
+
+from prereise.gather.griddata.transmission.const import (
+    epsilon_0,
+    mu_0,
+    relative_permeability,
+    resistivity,
+)
+from prereise.gather.griddata.transmission.helpers import DataclassWithValidation
+
+
+@dataclass
+class Conductor(DataclassWithValidation):
+    """Represent a single conductor (which may be a stranded compsite).
+
+    :param float radius: outer radius of conductor.
+    :param str material: material of conductor. Used to calculate ``resistance_per_km``
+        and ``gmr`` if these aren't passed to the constructor, unnecessary otherwise.
+    :param float resistance_per_km: resistance (ohms) per kilometer. Will be estimated
+        from other parameters if it isn't passed.
+    :param float gmr: geometric mean radius of conductor. Will be estimated from
+        other parameters if it isn't passed.
+    :param float area: cross-sectional area of conductor. Will be estimated from
+        other parameters if it isn't passed.
+    """
+
+    radius: float
+    material: str = None
+    resistance_per_km: float = None
+    gmr: float = None
+    area: float = None
+    permeability: float = field(init=False)
+
+    def __post_init__(self):
+        # Validate inputs
+        self.validate_input_types()  # defined in DataclassWithValidation
+        if self.gmr is None and (self.material is None):
+            raise ValueError(
+                "If gmr is not provided, material and radius are needed to estimate"
+            )
+        if self.resistance_per_km is None and self.material is None:
+            raise ValueError(
+                "If resistance_per_km is not provided, material is needed to estimate"
+            )
+        # Estimate missing inputs using the inputs which are present
+        if self.gmr is None:
+            try:
+                self.permeability = relative_permeability[self.material]
+            except KeyError:
+                raise ValueError(
+                    f"Unknown permeability for {self.material}, can't calculate gmr"
+                )
+            self.gmr = self.radius * exp(self.permeability / 4)
+
+        if self.resistance_per_km is None:
+            try:
+                self.resistivity = resistivity[self.material]
+            except KeyError:
+                raise ValueError(
+                    f"Unknown resistivity for {self.material}, "
+                    "can't calculate resistance"
+                )
+            if self.area is None:
+                self.area = pi * self.radius**2
+            # convert per-m to per-km
+            self.resistance_per_km = self.resistivity * 1000 / self.area
+
+
+@dataclass
+class ConductorBundle(DataclassWithValidation):
+    """Represent a bundle of conductors (or a 'bundle' of one).
+
+    :param int n: number of conductors in bundle (can be one).
+    :param Conductor conductor: information for each conductor.
+    :param float spacing: distance between the centers of each conductor (meters).
+    :param str layout: either 'circular' (conductors are arranged in a regular polygon
+        with edge length ``spacing``) or 'flat' (conductors are arranged in a line, at
+        regular spacing ``spacing``).
+    """
+
+    n: int
+    conductor: Conductor
+    spacing: float = None  # we need to be able to ignore spacing for a single conductor
+    layout: str = "circular"
+    resistance_per_km: float = field(init=False)
+    spacing_L: float = field(init=False)  # noqa: N815
+    spacing_C: float = field(init=False)  # noqa: N815
+
+    def __post_init__(self):
+        self.validate_input_types()  # defined in DataclassWithValidation
+        self.resistance_per_km = self.conductor.resistance_per_km / self.n
+        self.spacing_L = self.calculate_equivalent_spacing("inductance")
+        self.spacing_C = self.calculate_equivalent_spacing("capacitance")
+
+    def calculate_equivalent_spacing(self, type="inductance"):
+        if type == "inductance":
+            conductor_distance = self.conductor.gmr
+        elif type == "capacitance":
+            conductor_distance = self.conductor.radius
+        else:
+            raise ValueError("type must be either 'inductance' or 'capacitance'")
+        if self.n == 1:
+            return conductor_distance
+        elif self.n == 2:
+            return (conductor_distance * self.spacing) ** (1 / 2)
+        else:
+            if self.layout == "circular":
+                return self.calculate_equivalent_spacing_circular(conductor_distance)
+            if self.layout == "flat":
+                return self.calculate_equivalent_spacing_flat(conductor_distance)
+            raise ValueError(f"Unknown layout: {self.layout}")
+
+    def calculate_equivalent_spacing_circular(self, conductor_distance):
+        if self.n == 3:
+            return (conductor_distance * self.spacing**2) ** (1 / 3)
+        if self.n == 4:
+            return (conductor_distance * self.spacing**3 * 2 ** (1 / 2)) ** (1 / 4)
+        raise NotImplementedError(
+            "Geometry calculations are only implemented for 1 <= n <= 4"
+        )
+
+    def calculate_equivalent_spacing_flat(self, conductor_distance):
+        if self.n == 3:
+            return (conductor_distance * 2 * self.spacing**2) ** (1 / 3)
+        if self.n == 4:
+            return (conductor_distance * 12 * self.spacing**3) ** (1 / 8)
+        raise NotImplementedError(
+            "Geometry calculations are only implemented for 1 <= n <= 4"
+        )
+
+
+@dataclass
+class PhaseLocations(DataclassWithValidation):
+    """Represent the locations of each conductor bundle on a transmission tower. Each of
+    ``a``, ``b``, and ``c`` are the (x, y) location(s) of that phase's conductor(s).
+
+    :param tuple a: the (x, y) location of the single 'A' phase conductor if
+        ``circuits`` == 1, or the ((x1, y1), (x2, y2), ...) locations of the 'A' phase
+        conductors if ``circuits`` > 1. Units are meters.
+    :param tuple b: the (x, y) location of the single 'B' phase conductor if
+        ``circuits`` == 1, or the ((x1, y1), (x2, y2), ...) locations of the 'B' phase
+        conductors if ``circuits`` > 1. Units are meters.
+    :param tuple c: the (x, y) location of the single 'C' phase conductor if
+        ``circuits`` == 1, or the ((x1, y1), (x2, y2), ...) locations of the 'C' phase
+        conductors if ``circuits`` > 1. Units are meters.
+    :param int circuits: the number of circuits on the tower.
+    """
+
+    a: tuple
+    b: tuple
+    c: tuple
+    circuits: int = 1
+    equivalent_distance: float = field(init=False)
+    equivalent_height: float = field(init=False)
+    phase_self_distances: dict = field(init=False, default=None)
+    equivalent_reflected_distance: float = field(init=False)
+
+    def __post_init__(self):
+        self.validate_input_types()  # defined in DataclassWithValidation
+        if not (len(self.a) == len(self.b) == len(self.c)):
+            raise ValueError("each phase location must have the same length")
+        if self.circuits == 1 and len(self.a) == 2 and isinstance(self.a[0], float):
+            # Single-circuit specified as (x, y) will be converted to ((x, y))
+            self.a = (self.a,)
+            self.b = (self.b,)
+            self.c = (self.c,)
+        self.calculate_distances()
+
+    def calculate_distances(self):
+        self.true_distance = {
+            "ab": _geometric_mean_euclidian(self.a, self.b),
+            "ac": _geometric_mean_euclidian(self.a, self.c),
+            "bc": _geometric_mean_euclidian(self.b, self.c),
+        }
+        # 'Equivalent' distances are geometric means
+        self.equivalent_distance = geometric_mean(self.true_distance.values())
+        self.equivalent_height = geometric_mean(
+            [self.a[0][1], self.b[0][1], self.c[0][1]]
+        )
+        if self.circuits == 1:
+            self.calculate_single_circuit_distances()
+        else:
+            self.calculate_multi_circuit_distances()
+
+    def calculate_single_circuit_distances(self):
+        # The distance bounced off the ground, or 'reflected', is used for
+        # single-circuit capacitance calculations
+        self.reflected_distance = {
+            "ab": _euclidian(self.a[0], (self.b[0][0], -self.b[0][1])),  # a -> b'
+            "ac": _euclidian(self.a[0], (self.c[0][0], -self.c[0][1])),  # a -> c'
+            "bc": _euclidian(self.b[0], (self.c[0][0], -self.c[0][1])),  # b -> c'
+        }
+        self.equivalent_reflected_distance = geometric_mean(
+            self.reflected_distance.values()
+        )
+
+    def calculate_multi_circuit_distances(self):
+        self.phase_self_distances = [
+            geometric_mean(_euclidian(p0, p1) for p0, p1 in combinations(phase, 2))
+            for phase in (self.a, self.b, self.c)
+        ]
+        # Multi circuit, so we assume tall tower negligible impact from reflectance
+        self.equivalent_reflected_distance = 2 * self.equivalent_height
+
+
+@dataclass
+class Tower(DataclassWithValidation):
+    """Given the geometry of a transmission tower and conductor bundle information,
+    estimate per-kilometer inductance, resistance, and shunt capacitance.
+
+    :param PhaseLocations locations: the locations of each conductor bundle.
+    :param ConductorBundle bundle: the parameters of each conductor bundle.
+    """
+
+    locations: PhaseLocations
+    bundle: ConductorBundle
+    resistance: float = field(init=False)
+    inductance: float = field(init=False)
+    capacitance: float = field(init=False)
+
+    def __post_init__(self):
+        self.validate_input_types()  # defined in DataclassWithValidation
+        self.resistance = self.bundle.resistance_per_km / self.locations.circuits
+        self.inductance = self.calculate_inductance_per_km()
+        self.capacitance = self.calculate_shunt_capacitance_per_km()
+
+    def calculate_inductance_per_km(self):
+        denominator = _circuit_bundle_distances(
+            self.bundle.spacing_L, self.locations.phase_self_distances
+        )
+        inductance_per_km = (
+            mu_0 / (2 * pi) * log(self.locations.equivalent_distance / denominator)
+        )
+        return inductance_per_km
+
+    def calculate_shunt_capacitance_per_km(self):
+        denominator = _circuit_bundle_distances(
+            self.bundle.spacing_C, self.locations.phase_self_distances
+        )
+        capacitance_per_km = (2 * pi * epsilon_0) / (
+            log(self.locations.equivalent_distance / denominator)
+            - log(
+                self.locations.equivalent_reflected_distance
+                / (2 * self.locations.equivalent_height)
+            )
+        )
+        return capacitance_per_km
+
+
+def _euclidian(a, b):
+    """Calculate the euclidian distance between two points."""
+    try:
+        if len(a) != len(b):
+            raise ValueError("Length of a and b must be equivalent")
+    except TypeError:
+        raise TypeError(
+            "a and b must both be iterables compatible with the len() function"
+        )
+    return sqrt(sum((x - y) ** 2 for x, y in zip(a, b)))
+
+
+def _geometric_mean_euclidian(a_list, b_list):
+    """Calculate the geometric mean euclidian distance between two coordinate lists."""
+    try:
+        if len(a_list) != len(b_list):
+            raise ValueError("Length of a_list and b_list must be equivalent")
+    except TypeError:
+        raise TypeError("a_list and b_list must both be iterables")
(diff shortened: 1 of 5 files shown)


danielolsen:
Logic for multi-circuit towers has been added, along with a test example from _Power System Analysis_. The formulation isn't super intuitive, but should follow geometric first principles for any number of circuits >=2. I can hop onto a video call and walk through it if anyone is interested in the details, or you can trust the results based on the new test.

The new test is for a single-conductor, double-circuit tower, to complement the existing two-conductor bundle, single-circuit tower test from the TAMU paper. If anyone is aware of a triple- or quadruple-circuit example with relevant inputs and outputs, I can test against that as well to validate n > 2.

A couple of small per-unit calculation functions are also added, and docstrings are added for the new classes, so I think this PR is ready for a full review.

YifanLi86:
@danielolsen Great work! Here are also some data points on 345kV and above that you can use to crosscheck (more will become available soon): Let me know if I missed anything. Looks like our statistical average from TAMU before lines up pretty well with double circuit ones which is good. And compared to region, design, conductor, etc, it appears that bundling is more impactful. But we won't be able to tell each line is SC or DC one-by-one anyways...

image

danielolsen:
> @danielolsen Great work! Here are also some data points on 345kV and above that you can use to crosscheck (more will become available soon): Let me know if I missed anything. Looks like our statistical average from TAMU before lines up pretty well with double circuit ones which is good. And compared to region, design, conductor, etc, it appears that bundling is more impactful. But we won't be able to tell each line is SC or DC one-by-one anyways...
> 
> image

Nice! Where does this table come from? What's the meaning of the `Spacing` column when `Bundling` is `1`?

YifanLi86:
> > @danielolsen Great work! Here are also some data points on 345kV and above that you can use to crosscheck (more will become available soon): Let me know if I missed anything. Looks like our statistical average from TAMU before lines up pretty well with double circuit ones which is good. And compared to region, design, conductor, etc, it appears that bundling is more impactful. But we won't be able to tell each line is SC or DC one-by-one anyways...
> > image
> 
> Nice! Where does this table come from? What's the meaning of the `Spacing` column when `Bundling` is `1`?

I compiled it based on reference 1 and 2 below, will try to add more on low voltages. I am guessing the spacing value is there but it is more like a "spacing requirements" rather than actually spacing and bundling, so you can ignore those when Bundling is 1.

danielolsen:
Here's the implementation of calculating the impedance for the AEP example using the new tools in this PR:
```python
>>> from math import pi
>>> from prereise.gather.griddata.transmission.geometry import (
...     Conductor,
...     ConductorBundle,
...     PhaseLocations,
...     Tower,
... )
>>> from prereise.gather.griddata.transmission.helpers import (
...     calculate_z_base,
...     translate_to_per_unit,
... )
>>> # Define some helpful constants
>>> mm_in_one_inch = 25.4
>>> m_in_one_inch = mm_in_one_inch * 1e-3
>>> m_in_one_ft = m_in_one_inch * 12
>>> km_in_one_mile = 1.609
>>> omega = 60 * 2 * pi
>>> z_base = calculate_z_base(v_base=345, s_base=100)
>>> # Parameters for ASCR "Rail", 60 Hz, 50C
>>> conductor = Conductor(
...     radius=((1.165 / 2) * m_in_one_inch),
...     material="ASCR",
...     gmr=(.0386 * m_in_one_ft),
...     resistance_per_km=(.1092 / km_in_one_mile),
... )
>>> bundle = ConductorBundle(
...     n=2,
...     conductor=conductor,
...     spacing=(18 * m_in_one_inch),
... )
>>> # Double-circuit tower design '3L8' from the EPRI book.
>>> # Assume 100' to lowest conductors,
>>> # 24' vertical spacing,
>>> # Phases are 22' from centerline except middle conductors, which are 36'.
>>> # Center of tower at ground level is (0, 0). Units are meters.
>>> locations = PhaseLocations(
...     circuits=2,
...     a=((-22 * m_in_one_ft, 148 * m_in_one_ft), (22 * m_in_one_ft, 100 * m_in_one_ft)),
...     b=((-36 * m_in_one_ft, 124 * m_in_one_ft), (36 * m_in_one_ft, 124 * m_in_one_ft)),
...     c=((-22 * m_in_one_ft, 100 * m_in_one_ft), (22 * m_in_one_ft, 148 * m_in_one_ft)),
... )
>>> tower = Tower(locations=locations, bundle=bundle)
>>> # Once we've instantiated the Tower object, we automatically have per-km parameters
>>> per_mile_resistance_per_unit = translate_to_per_unit(
...     tower.resistance * km_in_one_mile, "r", z_base
... )
>>> per_mile_resistance_per_unit
2.293635790800252e-05
>>> per_mile_reactance_per_unit = translate_to_per_unit(
...     tower.inductance * omega * km_in_one_mile, "x", z_base
... )
>>> per_mile_reactance_per_unit
0.00024169519387211635
>>> per_mile_admittance_per_unit = translate_to_per_unit(
...     tower.capacitance * omega * km_in_one_mile, "siemens", z_base
... )
>>> per_mile_admittance_per_unit
0.017357500685635782
```
@YifanLi86 I'm getting a reactance that's about half of yours, did you model this example as single-circuit instead of double?

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+@dataclass
+class PhaseLocations:
+    """Represent the locations of each conductor bundle on a transmission tower.
+
+    :param tuple a: the (x, y) location of the single 'A' phase conductor if
+        ``circuits`` == 1, or the ((x1, y1), (x2, y2), ...) locations of the 'A' phase
+        conductors if ``circuits`` > 1. Units are meters.
+    :param tuple a: the (x, y) location of the single 'B' phase conductor if
```
jenhagg: 2nd and 3rd param names should be `b` and `c`

jenhagg: I would consider combining the descriptions of tuple parameters in the initial docstring, before the param section. E.g. `Each of a,b,c are the (x,y) location...`

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,272 @@
...
+
+def _euclidian(a, b):
+    try:
+        if len(a) != len(b):
+            raise ValueError("Length of a and b must be equivalent")
+    except TypeError:
+        raise TypeError(
+            "a and b must both be iterables compatible with the len() function"
+        )
+    return sqrt(sum((a[i] - b[i]) ** 2 for i in range(len(a))))
```
jenhagg: Could do `sqrt(sum((x-y)**2 for x,y in zip(a,b)))`

YifanLi86:
@danielolsen yes you are right...I misunderstood some data there in the EPRI book, and it ended up like the calculation is all single circuit. Updated table. The values are still a little bit different than yours, but it is mostly due to that I approximate the distance calculation using the minimum clearance data rather than the actual design. More data coming and I will get a complete table soon.

image

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,272 @@
...
+            raise ValueError("each phase location must have the same length")
+
+
+@dataclass
+class Tower:
+    """Given the geometry of a transmission tower and conductor bundle information,
+    estimate per-kilometer inductance, resistance, and shunt capacitance.
+
+    :param PhaseLocations locations: the locations of each conductor bundle.
+    :param ConductorBundle bundle: the locations of each conductor bundle.
```
jenhagg: copy/paste description?

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,272 @@
...
+        self.a = self.locations.a
+        self.b = self.locations.b
+        self.c = self.locations.c
+        self.calculate_distances()
+        self.resistance = self.bundle.resistance_per_km / self.locations.circuits
+        self.inductance = self.calculate_inductance_per_km()
+        self.capacitance = self.calculate_shunt_capacitance_per_km()
+
+    def calculate_distances(self):
+        if self.locations.circuits == 1:
```
jenhagg: Would it make sense to have `Tower` subclasses for 1 circut and (either >1 circuit, or specifically 2, not sure)?

danielolsen:
@YifanLi86 I don't think we need to go through too many manual calculations. If we can build up a 'library' of machine-readable tower designs and common ACSR conductor parameters (e.g. as a CSV that we can load into a dataframe), then we can quickly loop through the combinations and calculate the parameters of interest with the new code. If you've got time to spend on this issue, I think the most useful information you might be able to provide is 'typical' designs for each voltage, or for different voltages and different regions.

prereise/gather/griddata/transmission/geometry.py
jenhagg: Another possibility would be moving some calculations to the `PhaseLocations` class if they don't depend on `self.bundle`, in which case the subclasses would extend from there.

danielolsen: Sub-classing could work. You're right that the logic for 2 circuits should work for any plural number of circuits (although it's only been validated against a double circuit example).

The `__post_init__` would be shared among the two designs, but the other methods would differ. Would we be able to design it so that we can still instantiate a `Tower`, and the number of circuits would automatically determine which subclass of object is generated/returned? I guess at that point `Tower` becomes a factory, rather than a class, and looks at the `circuits` input to determine whether to instantiate and return a `SingleCircuitTower` or `MultiCircuitTower` object.

Either that, or maybe we can tweak the format or parsing for the single-circuit inputs, such that the multiple-circuit equations still work with only a single circuit.

danielolsen: I think you're right that the distance calculations can live in the `PhaseLocations` class. Calculating the inductance and the capacitance does need bundle info though.

prereise/gather/griddata/transmission/geometry.py
danielolsen: Done.

prereise/gather/griddata/transmission/geometry.py
danielolsen: Nice, done.

prereise/gather/griddata/transmission/geometry.py
danielolsen: Done.

prereise/gather/griddata/transmission/geometry.py
danielolsen: I refactored all of the distance calculations into `PhaseLocations` and a small helper function, so calculations in the `Tower` class are common between single-circuit and multi-circuit designs. There are still different calculations between single-circuit and multi-circuit distances within `PhaseLocations`, but they're organized behind self-contained methods and a single logic branch, which should hopefully be clearer to follow.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,282 @@
...
+        if len(a) != len(b):
+            raise ValueError("Length of a and b must be equivalent")
+    except TypeError:
+        raise TypeError(
+            "a and b must both be iterables compatible with the len() function"
+        )
+    return sqrt(sum((x - y) ** 2 for x, y in zip(a, b)))
+
+
+def _geometric_mean(a):
```
jenhagg: FYI starting in python 3.8, we could use `from statistics import geometric_mean`

danielolsen: Nice! Done.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@dataclass
+class Conductor:
+    """Represent a single conductor (which may be a stranded compsite).
+
+    :param float radius: outer radius of conductor.
+    :param str material: material of conductor. Used to calculate ``resistance_per_km``
+        and ``gmr`` if these aren't passed to the constructor, unnecessary otherwise.
+    :param float resistance_per_km: outer radius of conductor. Will be estimated from
```
BainanXia: `outer radius of conductor` -> `resistance per kilometer`?

danielolsen: Done. I guess I got pretty sloppy with my docstrings on this one.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,273 @@
...
+
+    radius: float
+    material: str = None
+    resistance_per_km: float = None
+    gmr: float = None
+    area: float = None
+
+    def __post_init__(self):
+        # Validate inputs
+        if self.gmr is None and (self.material is None or self.radius is None):
```
BainanXia: According to the definition, `self.radius` is a positional argument, which can't be `None` or must be provided?

danielolsen: Good catch, fixed. That was from an older iteration of the code where I didn't realize that I always need radius in order to be able to calculate capacitance.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,273 @@
...
+        if self.resistance_per_km is None:
+            try:
+                self.resistivity = resistivity[self.material]
+            except KeyError:
+                raise ValueError(
+                    f"Unknown resistivity for {self.material}, "
+                    "can't calculate resistance"
+                )
+            if self.area is None:
+                self.area = pi * self.radius**2
```
BainanXia, line 67: `self.area` is useful only if `self.resistance_per_km` is not provided and needs to be calculated? Not sure whether we should put this up to the same level of other attributes check.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,273 @@
...
+        other parameters if it isn't passed.
+    :param float area: cross-sectional area of conductor. Will be estimated from
+        other parameters if it isn't passed.
+    """
+
+    radius: float
+    material: str = None
+    resistance_per_km: float = None
+    gmr: float = None
+    area: float = None
```
BainanXia, line 34: AFAIK, these variable types are annotations only, not sure whether we should perform type checks upon instantiation if it is important.

prereise/gather/griddata/transmission/geometry.py
danielolsen: We always have `radius`, so we can always estimate the cross-sectional area if necessary (as long as we also have the `material`, which we already check for).

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,273 @@
...
+            self.material is None or self.radius is None
+        ):
+            raise ValueError(
+                "If resistance_per_km is not provided, "
+                "material and radius are needed to estimate"
+            )
+
+        if self.gmr is None:
+            try:
+                self.permeability = relative_permeability[self.material]
```
BainanXia, line 51: Not sure whether we would like these attributes which are assigned/calculated in post initialization to be shown once the data object is printed, if so, we could do `permeability: float = field(init=False) ` together with other class attributes.

prereise/gather/griddata/transmission/geometry.py
danielolsen: You're right, if the user provides silly inputs they'll get an unintuitive error here:
```python
>>> from prereise.gather.griddata.transmission.geometry import Conductor
>>> c = Conductor(radius="foo", resistance_per_km="bar", material="aluminum")
  File "<string>", line 8, in __init__
  File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\transmission\geometry.py", line 52, in __post_init__
    self.gmr = self.radius * exp(self.permeability / 4)
TypeError: can't multiply sequence by non-int of type 'float'
```

I added a parent class with a validation method, so now instead we get:
```python
>>> c = Conductor(radius="foo", resistance_per_km="bar", material="aluminum")
  File "<string>", line 8, in __init__
  File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\transmission\geometry.py", line 46, in __post_init__
    self._validate_input_types()
  File "C:\Users\DanielOlsen\repos\bes\PreREISE\prereise\gather\griddata\transmission\geometry.py", line 18, in _validate_input_types
    raise TypeError(
TypeError: for radius: expected <class 'float'>, but got <class 'str'>
```

**EDIT**: wait, the current commit doesn't work with good inputs since it doesn't like the default None values, I need to fix that.

danielolsen: Okay, the updated code does the trick.

prereise/gather/griddata/transmission/geometry.py
danielolsen: TIL! I've updated everything so that all these intermediately-calculated or looked-up values are contained in the `repr`:
```python
>>> from prereise.gather.griddata.transmission.geometry import Conductor
>>> c = Conductor(radius=0.1, material="aluminum")
>>> c
Conductor(radius=0.1, material='aluminum', resistance_per_km=0.0008435211983870453, gmr=0.12840254166877416, area=0.031415926535897934, permeability=1)
```

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,286 @@
...
+            if self.layout == "flat":
+                return self.calculate_equivalent_spacing_flat(conductor_distance)
+            raise ValueError(f"Unknown layout: {self.layout}")
+
+    def calculate_equivalent_spacing_circular(self, conductor_distance):
+        if self.n == 3:
+            return (conductor_distance * self.spacing**2) ** (1 / 3)
+        if self.n == 4:
+            return (conductor_distance * self.spacing**3 * 2 ** (1 / 2)) ** (1 / 4)
+        raise NotImplementedError("Geometry calculations not implemented for n > 4")
```
BainanXia: Maybe we can say `Geometry calculations are implemented for 1 <= n <= 4` to cover corner cases for non-positive integers.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def calculate_equivalent_spacing_flat(self, conductor_distance):
+        if self.n == 3:
+            return (conductor_distance * 2 * self.spacing**2) ** (1 / 3)
+        if self.n == 4:
+            return (conductor_distance * 12 * self.spacing**3) ** (1 / 8)
+        raise NotImplementedError("Geometry calculations not implemented for n > 4")
```
BainanXia: Same here.

prereise/gather/griddata/transmission/geometry.py
danielolsen: Done.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,286 @@
...
+        self.phase_self_distances = [
+            geometric_mean(
+                [
+                    _euclidian(p0, p1)
+                    for i, p0 in enumerate(phase)
+                    for j, p1 in enumerate(phase)
+                    if i > j
+                ]
+            )
+            for phase in (self.a, self.b, self.c)
```
BainanXia: Not sure whether this is equivalent to:
```python
geometric_mean( _euclidian(p0, p1) for p0, p1 in combinations(phase, 2) for phase in (self.a, self.b, self.c)
```

prereise/gather/griddata/transmission/helpers.py
```diff
@@ -0,0 +1,50 @@
...
+            raise TypeError(
+                "_validate_input_types can only be called on dataclassed objects."
+            )
+        for (name, specified_type) in self.__annotations__.items():
+            # Ignore attributes that aren't passed via the __init__ method
+            if not self.__dataclass_fields__[name].init:
+                continue
+            # Check whether the attribute is the specified or default type
+            # This supports inputs which have a default value of None
+            default_type = type(self.__dataclass_fields__[name].default)
```
BainanXia, line 18: Brilliant!

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -0,0 +1,290 @@
...
+    :param int circuits: the number of circuits on the tower.
+    """
+
+    a: tuple
+    b: tuple
+    c: tuple
+    circuits: int = 1
+    equivalent_distance: float = field(init=False)
+    equivalent_height: float = field(init=False)
+    phase_self_distances: dict = field(init=False)
```
BainanXia: Not sure whether we could initialize `phase_self_distances` to be `None` given `circuits` is initialized to be 1, then l.199 can be removed.

prereise/gather/griddata/transmission/helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    """Calculate base impedance from voltage and system base apparent power.
+
+    :param int/float v_base: base voltage (kV).
+    :param int/float s_base: base apparent power (MVA).
+    :return: (*float*) -- base impedance (ohms).
+    """
+    return (v_base**2) / s_base
+
+
+def translate_to_per_unit(x, nominal_unit, z_base):
```
BainanXia, line 36: It seems this function is only used in tests, i.e. called the corresponding test function to test it. Does it used elsewhere in the codebase?

prereise/gather/griddata/transmission/tests/test_helpers.py
```diff
@@ -0,0 +1,55 @@
...
+    test_object = NotDataclass(a=2, b=3.0)
+    with pytest.raises(TypeError):
+        test_object.validate_input_types()
+
+
+def test_calculate_z_base():
+    assert calculate_z_base(345, 100) == 1190.25
+
+
+def test_calculate_to_per_unit():
```
BainanXia: `test_calcuate_to_per_unit()` -> `test_translate_to_per_unit()`?

prereise/gather/griddata/transmission/geometry.py
danielolsen: Nice! Yeah `combinations` is definitely the cleaner way to do this. Done.

prereise/gather/griddata/transmission/helpers.py
danielolsen: Eh, I had to figure it out the hard way... `__dir__` plus guess-and-check.

prereise/gather/griddata/transmission/geometry.py
danielolsen: Yeah! Done.

prereise/gather/griddata/transmission/helpers.py
danielolsen: Not yet, but it will be used in subsequent code, since we eventually need to translate nominal impedances to per-unit for use in powerflows.

prereise/gather/griddata/transmission/tests/test_helpers.py
danielolsen: Good catch, done.

BainanXia (approved):
Very nice. Thanks for adding in-line explanations and tests, which are super helpful for both reviewers and future users.

PR #266, opened 2022-02-08
feat: generate profiles for immediate charging

isond:

### Purpose
To add the latest refactored code for immediate charging.

### What the code is doing
How is the purpose executed?
This code calculates the charging profiles for either LDVs or LDTs by iterating through the NHTS data. It then adjusts the charging profiles by applying weighting factors to yield the final results.

### Testing
This code was tested by comparing the output against the output from the original matlab version.

### Where to look
The new code can be found in "PreREISE/prereise/gather/demanddata/transportation_electrification/".

`immediate.py` is the most important file since it contains the main algorithm.

### Usage Example/Visuals

### Time estimate
It should take a few days for the reviewers to understand.


prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,114 @@
+def get_charging_power(power, trip_strategy, location_strategy, kwh, trip_data):
+    """Determines the charging power. Returns either param power or 0.
+
+    :param int power: charger power, EVSE kW.
+    :param int trip_strategy: a flag that determines which trips to consider, 1-anytrip number, 2-last trip.
+    :param int location_strategy: where the vehicle can charge-1, 2, 3, 4, or 5; 
+        1-home only, 2-home and work related, 3-anywhere if possibile, 
+        4-home and school only, 5-home and work and school.
+    :param float kwh: kwhmi * veh_range, amount of energy needed to charge vehicle.
```
danielolsen: Is this equal to the total battery energy capacity? If so, won't the battery SOC at the end of a trip always be less than this capacity?

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,114 @@
...
+            trip_strategy, trip_data["total vehicle trips"], trip_data["trip number"]
+        )
+    )
+
+    battery = int(get_battery_SOC(trip_data["trip end battery charge"], kwh))
+
+    if (dwelling * location * trip * battery != 1) or not (
+        (dwelling * location == 1)
+        or (trip_data["total vehicle trips"] == trip_data["trip number"])
+    ):
```
danielolsen: What's the purpose of the second big condition in this `if`? It seems like if any of `dwelling`, `location`, `trip`, `battery` are 0, then the first condition will trigger, and the second condition is redundant since it will only trigger if at least one of {`dwelling` or `location`} are 0 and the trip isn't the last one, but if at least one of {`dwelling` or `location`} are 0 then we're already covered by the first big condition. Maybe I'm missing something though, since boolean algebra isn't my strong suit.

As a side note, I think the boolean algebra would be clearer if we didn't convert the booleans to integers. In that case, `dwelling * location * trip * battery != 1` would be clearer as `not all([dwelling, location, trip, battery])`.

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,114 @@
...
+        charging_power = 0
+
+    return charging_power
```
danielolsen: I think we can change these to:
```python
    return 0

return power
```
and remove the `charging_power = power` line at the start of the function.

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,114 @@
...
+    elif location_strategy == 4:
+        # 21 is 'go to school as student'
+        return dwell_location in {1, 21}
+
+    # home and work and school
+    elif location_strategy == 5:
+        if dwell_location in {1, 11, 12, 21}:
+            return True
+        else:
+            return False
```
danielolsen: I think this can be changed to just `return dwell_location in {1, 11, 12, 21}`

prereise/gather/demanddata/transportation_electrification/charging.py
```diff
@@ -0,0 +1,114 @@
...
+    """Determines if the vehicle can be charged given location strategy and dwelling location
+
+    :param int location_strategy: where the vehicle can charge-1, 2, 3, 4, or 5; 
+        1-home only, 2-home and work related, 3-anywhere if possibile, 
+        4-home and school only, 5-home and work and school.
+    :param int dwell_location: location the vehicle dwells
+    :return: (*bool*) -- a boolean that represents whether or not the vehicle can charge
+    """
+    # only home
+    if location_strategy == 1:
```
danielolsen: One way this could be refactored is to use a dictionary:
```python
location_allowed = {
    1: dwell_location == 1,  # only home
    2: dwell_location < 13,  # home and go to work related
    3: True,  # anywhere
    4: dwell_location in {1, 21},  # home and school only
    5: dwell_location in {1, 11, 12, 21},  # home and work and school
}
return location_allowed[location_strategy]
```
Besides being more compact, this has the advantage of raising an error if we give an unexpected strategy number, rather than silently returning `None`.

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,117 @@
...
+    return data["Date"].dt.month
+
+
+def get_day_of_week(data: pd.DataFrame):
+    """Get day of week designation value from data.
+
+    :param pandas.DataFrame data: the data to get day of week from.
+    :return: (*numpy.array*) -- indicates the days of the week.
+    """
+    return np.array(data["Day of Week"])
```
danielolsen: Are these functions used anywhere?

prereise/gather/demanddata/transportation_electrification/mileage.py
```diff
@@ -0,0 +1,117 @@
+from prereise.gather.demanddata.transportation_electrification import const
+from scipy.io import loadmat
+import numpy as np
+import pandas as pd
+
+
+def get_model_year_dti(model_year: int):
```
danielolsen: I don't really understand the naming convention of this new module. None of these functions seem like they do anything related to mileage, they seem like they're helper functions for time or for loading/filtering data.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+    weekday_electricload, weekend_electricload = np.zeros(4800), np.zeros(4800)
+
+    l = len(newdata)
+
+    # weekday first
+    trip_num = 1
+    start_point = 0
+    end_point = 0
+
+    for i in range(l):
```
danielolsen: These two big loops seem nearly identical, with the only difference being that one operates on weekdays only and the other operates on weekends only. I think it would be clearer to define a function that takes in an arbitrary dataframe and calculates the new column values, and then to call this function twice: once on the weekday data and once on the weekend data.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
+from prereise.gather.demanddata.transportation_electrification import const
+from prereise.gather.demanddata.transportation_electrification import mileage
+from prereise.gather.demanddata.transportation_electrification import charging
+import numpy as np
+
+
+def immediate_charging(
```
danielolsen, line 103: This function could use a parameter that specifies the location of the data file, and passes this through to `load_data`, since otherwise it can only be run within a directory that contains the data file with the exact name.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+                )
+                trip_num += 1
+                newdata.iloc[i, newdata.columns.get_loc("trip number")] = trip_num
+
+            else:
+                newdata.iloc[
+                    i, newdata.columns.get_loc("trip start battery charge")
+                ] = kwh
+                trip_num = 1
+                newdata.iloc[i, newdata.columns.get_loc("trip number")] = trip_num
```
danielolsen: This code looks like we're grouping the trips by sample vehicle number, and adding values for `"trip number"` and `"trip start battery charge"`. I think we can do this in a clearer way by defining a function that calculates these values for all trips of a given vehicle at once, like we did with the `calculate_dwell_time` function within #243.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+                newdata.iloc[i, newdata.columns.get_loc("trip end battery charge")] = (
+                    newdata.iloc[
+                        i, newdata.columns.get_loc("trip start battery charge")
+                    ]
+                    - newdata.iloc[i, newdata.columns.get_loc("Miles traveled")]
+                    * kwhmi
+                    * const.ER
+                )
+            else:
+                newdata.iloc[i, newdata.columns.get_loc("BEV could be used")] = 0
```
danielolsen: Once we have the values of `"trip end battery charge"`, can we calculate the values for the new `"BEV could be used"` and `"trip end battery charge"` columns for all columns at once, rather than within a loop? E.g. something like:
```python
newdata["BEV could be used"] = (newdata["total vehicle miles traveled"] < veh_range * const.ER).map(int)
newdata.loc[newdata["BEV could be used"] == 1, "trip end battery charge"] = (
    newdata.loc[newdata["BEV could be used"] == 1, "trip start battery charge"]
    - newdata.loc[newdata["BEV could be used"] == 1, "Miles traveled"]
    * kwhmi
    * const.ER
)
```
If we keep the `"BEV could be used"` column as a boolean, it gets even simpler, since we can use the boolean column directly for slicing:
```python
newdata["BEV could be used"] = newdata["total vehicle miles traveled"] < veh_range * const.ER
newdata.loc[newdata["BEV could be used"], "trip end battery charge"] = (
    newdata.loc[newdata["BEV could be used"], "trip start battery charge"]
    - newdata.loc[newdata["BEV could be used"], "Miles traveled"]
    * kwhmi
    * const.ER
)
```

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+        axis="columns",
+        fill_value=0,
+    )
+
+    input_day = mileage.get_input_day(mileage.get_model_year_dti(model_year))
+
+    TRANS_charge = np.zeros(24 * len(input_day))
+    data_day = mileage.get_data_day(newdata)
+
+    trip_strategy = 1
```
danielolsen: Do we want the user to be able to change this? If so, it might make more sense as an input with a default value, rather than hard-coded within the function.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+    kwh = kwhmi * veh_range
+    weekday_electricload, weekend_electricload = np.zeros(4800), np.zeros(4800)
+
+    l = len(newdata)
+
+    # weekday first
+    trip_num = 1
+    start_point = 0
+    end_point = 0
```
danielolsen: Are these values ever used, or are they always overwritten within the loop before they're accessed?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+
+            newdata.iloc[
+                i, newdata.columns.get_loc("charging time")
+            ] = charging.get_charging_time(
+                newdata.iloc[i, newdata.columns.get_loc("charging power")],
+                kwh,
+                newdata.iloc[i, newdata.columns.get_loc("trip end battery charge")],
+                const.charging_efficiency,
+                trip_data,
+            )
```
danielolsen: These functions could also be used outside the loop via an `apply` call which applies a function to each row/column of a dataframe. E.g.:
```python
newdata["charging power"] = newdata.apply(
    lambda x: charging.get_charging_power(
        power,
        trip_strategy,
        location_strategy,
        kwh,
        trip_data=x,
    ),
    axis=1,
```
The `lambda` function allows concise creation of one-time functions, in this case the function takes a single input (`x`) and returns the output of `charging.get_charging_power` called with `x` for `trip_data`, plus the other parameters. `axis=1` means that the function is called on each row of data, rather than each column.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+
+            newdata.iloc[i, newdata.columns.get_loc("charging consumption")] = (
+                newdata.iloc[i, newdata.columns.get_loc("charging power")]
+                * newdata.iloc[i, newdata.columns.get_loc("charging time")]
+                * const.charging_efficiency
+            )
```
danielolsen: Same comment as above: this can be down for all rows at once by operating on entire columns.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+    for k in range(48):
+        if k == 0:
+            weekday_output_electricload[k] = sum(weekday_electricload[:100]) / 100
+        elif k == 47:
+            weekday_output_electricload[k] = sum(weekday_electricload[4700:]) / 100
+        else:
+            weekday_output_electricload[k] = (
+                sum(weekday_electricload[(k + 1) * 100 - 150 : (k + 1) * 100 - 50])
+                / 100
+            )
```
danielolsen: I think this block can also be defined as a function that takes in a length-4800 array and down-samples to length 48, so that it doesn't need to get repeated for both weekdays and weekends.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+
+    outputs = [weekend_output_electricload, weekday_output_electricload]
+
+    for day_iter in range(len(input_day)):
+        # weekday = 1 if weekday, 0 if not weekday
+        weekday = input_day[day_iter] - 1
+
+        if day_iter == len(input_day) - 1:
+            TRANS_charge[day_iter * 24 :] += outputs[weekday][:24].conj().T
```
danielolsen: Is `conj` the complex conjugate? Does that make a difference here, where I think all of our values are purely real?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+def adjust_BEV(TRANS_charge, adjustment_values):
+    """Adjusts the charging profiles by applying weighting factors based on 
+    seasonal/monthly values
+
+    :param (*numpy.ndarray*) TRANS_charge: normalized charging profiles
+    :param (*pandas.DataFrame*) adjustment_values: weighting factors for each 
+        day of the year loaded from month_info_nhts.mat.
+    :return: (*numpy.ndarray*) -- the final adjusted charging profiles.
+    """
+    adj_vals = adjustment_values.conj().T
```
danielolsen: Same comment here about `conj`

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+                    i, newdata.columns.get_loc("trip start battery charge")
+                ] = kwh
+                trip_num = 1
+                newdata.iloc[i, newdata.columns.get_loc("trip number")] = trip_num
+
+            if (
+                newdata.iloc[i, newdata.columns.get_loc("total vehicle miles traveled")]
+                < veh_range * const.ER
+            ):
+                newdata.iloc[i, newdata.columns.get_loc("BEV could be used")] = 1
```
danielolsen: Does the `"BEV could be used"` column get used anywhere?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,298 @@
...
+                newdata.iloc[i, newdata.columns.get_loc("trip end battery charge")] = (
+                    newdata.iloc[
+                        i, newdata.columns.get_loc("trip start battery charge")
+                    ]
+                    - newdata.iloc[i, newdata.columns.get_loc("Miles traveled")]
+                    * kwhmi
+                    * const.ER
+                )
```
danielolsen: When the total daily distance is within the vehicle's range, it seems like we calculate the state of charge at the end of the trip (`"trip end battery charge"`) here within the `if` block. When the total distance is greater than the vehicle's range, it seems we don't set this value, so it remains at the default value of 0 (which it got when we first added this column). Won't this artificially inflate the total charging profile, since the charging profile for each of these trips will be built assuming that the vehicle arrives at the charger completely empty?

prereise/gather/demanddata/transportation_electrification/charging.py
isond: Yes, so if the battery SOC at the end of a trip is less than this kwh value, then it'll indicate that the vehicle needs to be charged.

prereise/gather/demanddata/transportation_electrification/charging.py
isond: In the original matlab code, one of the if-statements was to check if `dwelling` and `location` are equal to 1 while being on the last trip, then it would return the charging power. However, for this version, all the if-statements were consolidated (hence, the first condition and the second big condition) and we took their inverse. If they're not satisfied, then it returns 0.

Thinking more about it, it does seem redundant because whether or not it's on the last trip, if `dwelling` and `location` are 0, then it would be handled by the first condition as you mentioned. So I decided to remove the second portion and update it per your suggestion.

prereise/gather/demanddata/transportation_electrification/immediate.py
isond: I changed `trip_strategy` to a default argument as you suggested since we're focusing on charging after any trip (a value of 1). But in the future, if someone may want to use the last-trip capability (value of 2), then they'll be able to do so.

prereise/gather/demanddata/transportation_electrification/charging.py
danielolsen: Doesn't the vehicle always need to be charged at least a little bit at the end of a every trip? Even if there are zero-energy/zero-mileage trips, won't that end up getting handled downstream because the charging time will be zero (even if the charging power isn't)?

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -0,0 +1,251 @@
...
+        weekday = input_day[day_iter] - 1
+
+        if day_iter == len(input_day) - 1:
+            TRANS_charge[day_iter * 24 :] += outputs[weekday][:24].transpose()
+            TRANS_charge[:24] += outputs[weekday][24:48].transpose()
+
+        else:
+            TRANS_charge[day_iter * 24 : day_iter * 24 + 48] += outputs[
+                weekday
+            ].transpose()
```
danielolsen: Do we need to transpose at all here? We're adding one 1D array to another 1D array.

danielolsen (approved):
The code runs through all zones in just over a minute. If the data are as expected, I think this is ready to go!

PR #267, opened 2022-02-10
refactor: improve mapping of generators to buses

danielolsen:

### Purpose
In preliminary testing, the `hifld` grid model had shown a lot of infeasibilities caused by mismatches between inflexible generators (namely coal, nuclear, and hydro) and local transmission capacities. This PR aims to improve this in two primary ways:
- Aggregating multiple hydro generating units within one plant into a single generator with equivalent Pmax and Pmin sums
- Improving how generators are mapped to substations
- Improving how generators are mapped to buses within substations

Hydro aggregation is relevant to the `improving how generators are mapped to buses within substations` step, since we decide which bus within a substation to map to based on generator capacity.

The most complicated part by far is the how generators are mapped to substations. The old logic was:
- Are coordinates available for the generator?
  - **Yes**: Are there any substations with ZIP codes that match the generator?
    - **Yes**: return the closest substation within that ZIP
    - **No**: return the closest substation within the 'closest' 200 ZIP codes
  - **No**: Are there any substations with ZIP codes that match the generator?
    - **Yes**: return an arbitrarily chosen one
    - **No**: return nothing

The new logic is:
- Are coordinates available for the generator?
  - **Yes**: Find the closest substation within that generator's state & interconnection (with some edge-case handling)
  - **No**: Are there any substations with ZIP codes that match the generator?
    - **Yes**: return an arbitrarily chosen one
    - **No**: return nothing

The no-coordinates case is a fraction of a percent; the big changes are from the cases where we do have coordinates and now use those as the primary substation mapping tool, rather than ZIP codes.

### What the code is doing
Aggregating hydro units: a new `aggregate_hydro_generators_by_plant_id` function is added, which groups hydro units by plant ID, sums Pmin and Pmax, and returns all other attributes and index from the first unit in the group.

Improving the mapping of generators to substations:
- We remove the old `map_generator_to_sub_by_location` and add a replacement `map_generators_to_sub_by_location`. These function names differ because the previous one operated on one `generator` at a time as a part of an `apply` call, while the new one operates on the entire `generators` dataframe.
- The new function translates lats & lons of the `substations` and generators dataframes into (x, y, z) pairs in 3d space, where the center of the earth is (0, 0, 0) and the radius is 1. We use a new `latlon_to_xyz` function that we add to prereise, rather than the existing `ll2uv` implementation in powersimdata to loosen the coupling of the two packages and to avoid ambiguity between (lat, lon) and (lon, lat). Then, a `KDTree` is instantiated for each combination of (interconnect, state). If there exists one or more generators labelled with (interconnect, state) for which there are no substations with corresponding (interconnect, state):
  - if we believe the generator in that state and physically connected to the other interconnection, then we create a `KDTree` for that entire interconnection (e.g. a plant that's physically located in Oklahoma may have a single transmission line that connects it to ERCOT).
  - If we believe that the generator is in that state but we don't trust the interconnection information (e.g. plants in Virginia or Pennsylvania that claim to be connected to WECC), then we create a KDTree for the state.
  - These assumptions are printed for the user to examine.
- We have a small internal function which queries into the appropriate KDTree for each generator and translates the result to a substation ID.
- For any generators whose locations can't be found this way (i.e. they don't have latitudes and longitudes defined), we fall back to the old ZIP-code matching logic, where we try to grab a random substation within a matching ZIP code.
- Any remaining generators which can't be mapped to a substation will be filtered out of the final outputs.

Improving how generators are mapped to buses within substations: `map_generator_to_bus_by_sub` is refactored from the previous logic (always map to the lowest-voltage bus within the substation) to branching logic:
- If there's only one bus within the substation or the generator's Pmax is less than 200 MW, we connect to the lowest-voltage bus.
- If the generator is between 200 and 500 MW, connect to the second-lowest voltage bus
- If the generator is 500 MW or great, connect to the highest-voltage bus

There's also a small unrelated fix to prevent duplicate `'interconnect'` columns in some output CSVs, which was preventing the `grid.mat` files created from REISE.jl from being read back into PowerSimData for post-simulation analysis.

### Testing
Tested manually. The printouts for the edge cases of mapping generators to substations looks like:
```
no substations within (Western, MA), will map generators to substations within MA instead
no substations within (Western, MO), will map generators to substations within MO instead
no substations within (Eastern, CA), will map generators to substations within CA instead
no substations within (Western, VA), will map generators to substations within VA instead
no substations within (Western, PA), will map generators to substations within PA instead
no substations within (Western, KS), will map generators to substations within Western instead
no substations within (ERCOT, OK), will map generators to substations within ERCOT instead
```

With this change, about 25% of generators end up mapped to a different substation (2,837 out of 12,735). In addition, 29 generators can be mapped which were not mapped with the previous logic. The most impactful change may have been how Palo Verde's generator were mapped: previously they ended up at a 69 kV substation within metro Phoenix; now they're appropriately connected to the 500 kV substation at their true location. Many other large inflexible generators are relocated as well, and generators in WECC seem to be particularly affected.

Running powerflows with the results shows drastic improvement in WECC, where infeasibilities were initially the worst. Previously, the total amount of transmission line limit violation energy that was required was about 25% of the total demand. After the change, this is down to 0.45%. In addition, transmission violations occur at fewer than half as many lines as before. Eastern powerflows are still running but I expect to see significant improvements there as well. **EDIT**: Eastern is done. The improvements aren't quite as good as in Western, but are still big improvements. Transmission violation energy is down to 5.1% (from 18.2% before) and at 0.19% of branches (from 0.61% before). That puts the overall USA-wide number of violating branches down to around 200, less than 1 out of every 400 branches, and about a third as many as before the refactor.

### Usage Example/Visuals
All code is still launched via:
```python
from prereise.gather.griddata.hifld import create_csvs
create_csvs(output_folder_name)
```

### Time estimate
1 hour. Most of the new code is pretty straightforward, but about half of it is designed to combat edge cases caused by strange data inputs from the original EIA Form 860 data.


prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
+    latitude and longitude, an attempt will be made to match via ZIP code, and failing
+    that a pandas.NA value will be returned.
+
+    :param pandas.DataFrame generators: generator data. Required columns:
+        'interconnect', 'lat', 'lon', 'state', 'ZIP'.
+    :param pandas.DataFrame substations: substation data. Required columns:
+        'interconnect', 'lat', 'lon', 'STATE', 'ZIP'.
+    :param bool inplace: whether to modify the generator table inplace with a new
+        'sub_id' column or to return a new one. If ``inplace`` is `True`, entries in
+        `generators` which have non-sensical combinations of 'state' and 'interconnect'
```
BainanXia: To clarify the logic in the PR description: we will always trust the `state` tag of the generator. For boundary states, generators connect to other interconnections (different from the one specified in the `interconnect` column) is allowed, otherwise, the corresponding interconnect of the generator will be fixed, right?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
-        lambda x: haversine((x.lat, x.lon), (generator.lat, generator.lon)),
+        grouper_key = (generator["interconnect"], generator["state"])
+        _, array_index = state_trees[grouper_key].query(generator["xyz"])
+        sub_index = subs_state_lookup[grouper_key][array_index]
+        return sub_index
+
+    # Translate lat/lon to 3D positions (assume spherical earth, origin at center)
+    substations_with_xyz = substations.assign(
+        xyz=substations.apply(lambda x: latlon_to_xyz(x["lat"], x["lon"]), axis=1)
+    )
```
BainanXia, line 78: I think we can simply do 
`substations_with_xyz = substations.assign(xyz=lambda x: latlon_to_xyz(x.lat, x.lon))` here.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
+    generators_with_xyz = generators.assign(
+        xyz=generators.apply(
+            lambda x: (
+                pd.NA
+                if pd.isna(x["lat"]) or pd.isna(x["lon"])
+                else latlon_to_xyz(x["lat"], x["lon"])
+            ),
+            axis=1,
+        )
+    )
```
BainanXia, line 88: I think we don't need to filter out NaN here given any calculations will end up with NaN by default. Hence `generator_with_xyz = generators.assign(xyz=lambda x: latlon_to_xyz(x.lat, x.lon))` might work here.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: For 'boundary' states, we trust that the generator may actually be in the listed state, but that it may be connected to a substation in the matching interconnection but within another state. E.g. a generator in Oklahoma may actually be connected to the bulk power grid via an ERCOT substation in Texas. So from the grid's perspective, we trust the `interconnect` column and lat/lon more than the `state` column (since the grid only really cares where the power gets injected), although we do use the `state` column to determine when to apply this logic.

For non-boundary states, we trust the `state` column and the lat/lon more than the `interconnect` column, which we ignore.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: I think we might get an AttributeError, since `NaN` has no `lat`.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Nice! I haven't used callables with `assign` before, TIL.

prereise/gather/griddata/hifld/data_process/generators.py
BainanXia: My question is more on what will be potentially changed with `inplace` set, i.e. for boundary states, we will trust the `interconnect`, thus both `state` and `interconnect` won't be changed, whereas for non-boundary states, we will trust the `state` only then `interconnect` will be changed if does seem to be correct, right?

danielolsen: Right. If the generator is in Virginia, then we don't trust the `interconnect` column that says `Western` (which comes from EIA Form 860 when it says it's connected to CISO), and we change it to 'Eastern' instead.

prereise/gather/griddata/hifld/data_process/generators.py
BainanXia: Never mind...I looked into this and found it doesn't work given `assign` + `lambda` work with columns only. Let's keep the current implementation.

prereise/gather/griddata/hifld/data_process/generators.py
BainanXia: Same here. Let's keep the current implementation.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
+    state_trees = {
+        key: KDTree(np.array(substations_with_xyz.loc[sub_ids, "xyz"].tolist()))
+        for key, sub_ids in subs_state_lookup.items()
+    }
+    # Ensure that we have a tree for every generator
+    gens_state_groupby = generators_with_xyz.groupby(["interconnect", "state"])
+    missing_groups = set(gens_state_groupby.groups) - set(state_trees)
+    if len(missing_groups) > 0:
+        # There are some combinations of generator (interconnect, state) without subs
+        allowable_border_states = {"KS", "NE", "OK"}
```
BainanXia: Are these states coming from EIA references?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
+                subs_state_lookup[(interconnect, state)] = new_subs.index
+            else:
+                # Assume that the state is correct, the interconnect is wrong
+                print(
+                    f"no substations within ({interconnect}, {state}), "
+                    f"will map generators to substations within {state} instead"
+                )
+                (assumed,) = {
+                    interconnect
+                    for interconnect, state_list in const.interconnect2state.items()
```
BainanXia: We have `ignore` and `split` as keys in `const.interconnect2state`. Reading the std outputs, it doesn't seem to matter given generators in those states can find corresponding substations.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: State information for generators comes from the EIA. The 'allowable' border states come from my personal eyeballing. I should probably add South Dakota too.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
+                print(
+                    f"no substations within ({interconnect}, {state}), "
+                    f"will map generators to substations within {interconnect} instead"
+                )
+                # Find all substations for the interconnection
+                new_subs = substations_with_xyz.query("interconnect == @interconnect")
+                # Extend the 'true' combinations of (interconnect, state) with fakes
+                state_trees[(interconnect, state)] = KDTree(
+                    np.array(new_subs["xyz"].tolist())
+                )
```
BainanXia: Not sure how many such cases we have. We can pre-construct interconnect_trees in a dict as well and query corresponding ones here.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +24,108 @@ def floatify(value, default=float("nan")):
...
+    sub_ids = generators_with_xyz.apply(
+        lambda x: get_sub_id_of_closest_substation(x, state_trees, subs_state_lookup),
         axis=1,
     )
-    return distance_to_subs.idxmin()
+    # For generators without coordinates, try to pick a substation with a matching ZIP
+    for g in generators.loc[sub_ids.isnull()].index:
+        try:
+            candidates = subs_zip_groupby.get_group(generators.loc[g, "ZIP"])
+            sub_ids.loc[g] = candidates.index[0]  # arbitrary choose the first one
```
BainanXia: Just curious how many such cases? Our current logic flow is matching generator to substation first then to buses within the substation. If we match generator to buses directly, we can choose buses with proper `baseKV` here. Just thinking loud here maybe it's not doable.

danielolsen: 28 plants with no lat/lon in the 2019 Form 860, out of a total of 11,833.

We could definitely combine the logic of 'map generators to substation' and 'map generators to buses within substation' if we wanted to, since these are called one after another within `build_plant`.

BainanXia: 28/11833 is pretty minor. We can ignore it if they are not causing similar troubles that are described [here](https://app.zenhub.com/workspaces/renewable-energy-integration-5bfdd1184b5806bc2bf88976/issues/breakthrough-energy/renewableenergyproject/361) (I wasn't able to referencing the exact comment, but the one you uses Oconee nuclear plant as an example).

danielolsen: The Oconee example is not caused by this block, but by not making use of the Grid Voltage entries for each plant in Form 860.

I think I can replace the grouping of substations by state with segmentation by voltage classes (classes rather than absolute numbers, since there are may unique values but only a few 'clusters'), rather than try to group on both. The Grid Voltage data seem a bit noisy too (e.g. I bet the 1.5 MW individual wind turbines aren't actually connected at the listed 620 kV), so maybe I'll add a sanity check on the distance to the mapped substation that can simultaneously catch errors in interconnection and grid voltage.

danielolsen:
I've refactored this so that it ignores states completely, and uses the voltage(s) available at each substation to help ensure that generators don't get mapped to substations with inadequate transmission capacity, based on their listed 'grid voltage' within Form 860. In a quick one-day test on Eastern, it seems to help: previously the transmission violations were > 17 GW in every hour of the year, but in the test day we have transmission violations as low as 1.3 GW. I'll run the full year over the weekend and report more detailed results.

The nonsensical-interconnection listing is at least partially fixed by using the Balancing Authority column of each generator within Form 860 to map to interconnects, rather than the NERC region. For whatever reason, BA seems more reliable, and we only use the NERC region as a fall-back. The generators which end up getting mapped more than 50 miles away now all seem to be either:
- very small generators
- generators in very sparse areas (e.g. the Florida keys, southwest Texas)
- wind generators in the Texas panhandle (which indicates that we probably need to revise how we draw the border between ERCOT/EI there)

danielolsen:
Results for full-year runs with the [new voltage-class and interconnection mapping](https://github.com/Breakthrough-Energy/PreREISE/pull/267#issuecomment-1045341600) of generators to substations:
- Eastern interconnection: transmission violation energy reduced significantly (0.451% of annual demand, vs. 5.1% before) and at significantly fewer branches (0.10% of branches vs. 0.19% before)
- Western interconnection: slight improvement in transmission violation energy (0.31% of annual demand, vs. 0.45% before) and number of branches (0.27% before vs. 0.29% before)
- ERCOT: transmission violation energy reduced significantly (0.16% of annual demand, vs. 2.0% before) and slight improvement in number of branches (0.30% now vs. 0.43% before)

We're down to only 138 branches with violations across the whole USA. The next thing I'm going to try is revising the configuration of transformers within substations, which should hopefully increase the effective impedance between the higher-voltage buses at which large generators are connected and the lower-voltage buses connected to low-voltage branches with transmission violation.

danielolsen:
The 'cascade' configuration (every bus within a substation is connected via a transformer to the next-highest bus) reduces transmission violation energy only barely compared to the previous configuration (every bus is connected to the substation's highest-voltage bus), but has a larger impact on reducing the number of branches at which transmission violations occur:
- ERCOT: violation energy goes from 0.16% to 0.14%, and the number of branches goes from 0.29% to 0.26%
- Western: violation energy goes from 0.305% to 0.304%, and the number of branches goes from 0.27% to 0.26%
- Eastern: violation energy goes from 0.451% to 0.447%, and the number of branches goes from 0.10% to 0.08%

The remaining transmission violations are pretty heavily concentrated among a few branches, with what I believe are a few common root causes:
- There are some large generators for which there's a transmission line in the original HIFLD dataset originating within the plant and going to a nearby substation, but no substation listed for the plant itself. As a result, the plant-end of this transmission line gets mapped to another nearby substation, potentially causing a connection where there shouldn't be one and/or resulting in the generator getting connected to the grid at a different substation with lower transfer capacity (although this is partially mitigated by the new code in this branch which looks at substation voltages and plant-level grid connection voltages). This is essentially #234. We could manually add these substations, or change how we build the transmission network topology so that transmission line endpoints with no substation in the vicinity create their own substations. Manually adding is a little less elegant, but has a lower potential for topological side-effects.
- There is only a single assumed reactance and power rating for a line of a given voltage and length, ignoring the fact that some lines are double-circuit, triple-circuit, etc. This can result in a generator being connected at a transmission line which has only a third of the minimum generation that the generator is constrained to (e.g. the Catawba Nuclear station). This also pops up for some of the longer transmission lines in the West and Texas. **EDIT**: We don't currently have any code that can take in a subset of lines  and alter their parameters; it would be fairly trivial to do so by assuming two (or three, etc.) identical parallel lines, but properly modeling different sorts of towers/lines would mean we'd need to make some assumptions about designs and start using the new code introduced in #262 and #268.

prereise/gather/griddata/hifld/const.py
```diff
@@ -194,6 +194,79 @@
...
+    "SWPP": "Eastern",
+    "TAL": "Eastern",
+    "TEC": "Eastern",
+    "TEPC": "Western",
+    "TIDC": "Western",
+    "TPWR": "Western",
+    "TVA": "Eastern",
+    # "WACM": "Western",  # can be Western or Eastern
+    # "WALC": "Western",  # can be Western or Eastern
+    # "WAUW": "Western",  # can be Western or Eastern
```
rouille, line 264: So those are not considered?

prereise/gather/griddata/hifld/const.py
```diff
@@ -438,6 +511,18 @@
     "west": -125,
 }
```
rouille, line 512: Should we then delete these lines?

prereise/gather/griddata/hifld/const.py
```diff
@@ -438,6 +511,18 @@
...
 
+proxy_substations = [
+    {"LATITUDE": 35.0514, "LONGITUDE": -81.0694, "NAME": "Catawba", "STATE": "SC"},
+    {
+        "LATITUDE": 44.2853,
+        "LONGITUDE": -105.3826,
+        "NAME": "Neil Simpson II",
+        "STATE": "WY",
+    },
+]
```
rouille, line 522: What are these substations used for?

danielolsen: Sometimes there are short transmission lines for which one endpoint seems to connect directly to a generator, but there's no substation there. In this case, that end of the line gets connected to the next closest substation, which sometimes modifies the transmission network topology and creates infeasibilities (i.e. a power plant with inflexible generation gets mapped to a low-capacity line which can't support the minimum power being pushed.

By adding proxy substations at the endpoints of these lines, we avoid making these connections, and the generator pushes its power only to the high-voltage network to which it's really connected to.

prereise/gather/griddata/hifld/const.py
danielolsen: They are not. The comments are here as an indication that we know the WAPA balancing authorities are missing, and it's an intentional choice.

prereise/gather/griddata/hifld/const.py
danielolsen: The bounds? We still use them within `get_hifld_electric_power_transmission_lines`

rouille: Sorry,, I was referring to `nercregion2interconnect`. The comment says: Usage of this is deprecated, since these data seem noisier than Balancing Authorities

prereise/gather/griddata/hifld/const.py
rouille: Sounds good

prereise/gather/griddata/hifld/const.py
danielolsen: We do still use `nercregion2interconnect` if `balancingauthority2interconnect` doesn't work for all generators (e.g. the `WACM`, `WALC`, and `WAUW` BAs). I think when I wrote the comment I thought we could use `balancingauthority2interconnect` alone, so I should probably update the comment to make it a little clearer that this is still used, but only as a fallback.

BainanXia: @danielolsen That's consistent with my understanding. The fallback option got eventually used in some corner cases, which previous logic failed to catch.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+                if pd.isna(x["lat"]) or pd.isna(x["lon"])
+                else latlon_to_xyz(x["lat"], x["lon"])
+            ),
+            axis=1,
+        )
+    )
+
+    # Bin voltages into broad classes
+    substations_with_xyz["voltage_class"] = substations["MAX_VOLT"].map(
+        lambda x: classify_voltages(x, voltage_ranges)
```
BainanXia: Is the column `voltage_class` in data frame `substations_with_xyz` used anywhere?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+    )
+    generators_with_xyz["voltage_class"] = generators["Grid Voltage (kV)"].map(
+        lambda x: classify_voltages(x, voltage_ranges)
+    )
+
+    # Group substations by voltage to build KDTrees
+    subs_voltage_lookup = {
+        (interconnect, voltage_level): substations_with_xyz.loc[
+            (substations_with_xyz["interconnect"] == interconnect)
+            & (substations_with_xyz["MAX_VOLT"] >= voltage_range["min"])
```
BainanXia, line 99: So we only care about lower bound here instead of the exact `voltage_range` defined in the dict `voltage_ranges`?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+        ].index
+        for interconnect in generators["interconnect"].unique()
+        for voltage_level, voltage_range in voltage_ranges.items()
+    }
+    # Group substations by ZIP code for a fallback for generators without coordinates
+    subs_zip_groupby = substations_with_xyz.groupby(["interconnect", "ZIP"])
+
+    # Create a KDTree for each combination of state and interconnect
+    voltage_trees = {
+        key: KDTree(np.array(substations_with_xyz.loc[sub_ids, "xyz"].tolist()))
```
BainanXia: Are we using `numpy.ndarray` for performance concern? If so, maybe we can do `substations_with_xyz.loc[sub_ids, "xyz"].to_numpy()` directly?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+        for key, sub_ids in subs_voltage_lookup.items()
+        if len(sub_ids) > 0
+    }
+    # Create a KDTree for each interconnect (all voltages)
+    subs_interconnect_groupby = substations_with_xyz.groupby("interconnect")
+    for interconnect in generators["interconnect"].unique():
+        tree_subs = subs_interconnect_groupby.get_group(interconnect)
+        voltage_trees[interconnect] = KDTree(np.array(tree_subs["xyz"].tolist()))
```
BainanXia: Same comment as above.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
-                [
-                    substation_groupby.get_group((generator.loc["interconnect"], z))
-                    for z in zip_range_strings
-                    if (generator.loc["interconnect"], z) in substation_groupby.groups
-                ]
-            )
-        except ValueError:
-            # If no matching subs within the given interconnection and ZIPs, give up
+
+    def get_closest_substation(generator, state_trees, subs_state_lookup):
```
BainanXia: Shall we rename the input variables into `voltage_trees` and `subs_voltage_lookup` instead?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+    # Query the appropriate tree for each generator to get the closest substation ID
+    mapping_results = generators_with_xyz.apply(
+        lambda x: get_closest_substation(x, voltage_trees, subs_voltage_lookup),
         axis=1,
     )
-    return distance_to_subs.idxmin()
+    # For generators without coordinates, try to pick a substation with a matching ZIP
+    for g in generators.loc[mapping_results["sub_id"].isnull()].index:
+        try:
+            candidates = subs_zip_groupby.get_group(generators.loc[g, "ZIP"])
```
BainanXia: Will this work? `subs_zip_groupby` is grouped based on combinations of `interconnect` and `ZIP` rather than `ZIP` only.

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+    that a pandas.NA value will be returned.
+
+    :param pandas.DataFrame generators: generator data. Required columns:
+        'interconnect', 'lat', 'lon', 'state', 'ZIP'.
+    :param pandas.DataFrame substations: substation data. Required columns:
+        'interconnect', 'lat', 'lon', 'STATE', 'ZIP'.
+    :param bool inplace: whether to modify the generator table inplace with a new
+        'sub_id' column or to return a new one. If ``inplace`` is `True`, entries in
+        `generators` which have non-sensical combinations of 'state' and 'interconnect'
+        columns will have their 'interconnect' entries modified.
```
BainanXia: Does this logic still apply given `state` is no longer used? We don't modify `interconnect` any more, do we?

prereise/gather/griddata/hifld/data_process/generators.py
```diff
@@ -22,54 +26,131 @@ def floatify(value, default=float("nan")):
...
+    generators, substations, inplace=True, report_worst=None
+):
+    """Determine the closest substation to each generator. For generators without
+    latitude and longitude, an attempt will be made to match via ZIP code, and failing
+    that a pandas.NA value will be returned.
+
+    :param pandas.DataFrame generators: generator data. Required columns:
+        'interconnect', 'lat', 'lon', 'state', 'ZIP'.
+    :param pandas.DataFrame substations: substation data. Required columns:
+        'interconnect', 'lat', 'lon', 'STATE', 'ZIP'.
```
BainanXia: Should we remove `state` and `STATE`?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -822,6 +826,13 @@ def build_transmission(method="line2sub", **kwargs):
         lambda x: estimate_branch_rating(x, bus["baseKV"]), axis=1
     )
 
+    # Update substation max & min voltages using bus data (from lines)
+    substations.replace(
+        {"MAX_VOLT": -999999, "MIN_VOLT": -999999}, float("nan"), inplace=True
```
BainanXia, line 831: Should `MIN_VOLT` initialized by 999999 instead?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -822,6 +826,13 @@ def build_transmission(method="line2sub", **kwargs):
...
+    )
```
BainanXia: Not sure whether we need this or not, `update` will override everything based on the info from `bus` table if there is no `np.NA`, which shouldn't be the base given each substation contains at least one bus?

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: It appears not, I'll remove this and double-check that everything still works.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: The primary goal of this PR was to avoid large generators getting mapped to substations without at least one high-enough voltage bus (and therefor probably too low of a transfer capacity). I guess we could make it more strict by ensuring that there's at least one bus that's truly within the range, which will have the tradeoff that the distance to the connection location may sometimes increase. What do you think?

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: If I remember correctly, each entry in the `"xyz"` column is a list/tuple, so calling `.to_numpy()` directly gives us a 1D array of list/tuple objects, not the 2D array that we really want. I'll double-check to be sure. I agree that it's ugly.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Good catch. As it was written, those parameter were never being used, and the same-name versions from the parent scope were being used instead (which still behaved identically, but this function call signature didn't really make sense).

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Good catch! It 'works', but only since the `KeyError` always gets thrown and handled. Will fix.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: I think you're right. Now that we assign generators to interconnects based on the `Balancing Authority Code`, not the `NERC Region`, the interconnect info seems okay. Will update.

prereise/gather/griddata/hifld/data_process/generators.py
BainanXia: Yeah, you are right. Having more strict filter will potentially give us farther mapping if there is no match nearby. Which side in the trade-off is more important, the location or the voltage range? If the voltage range turns out to be more important, let's go with the more strict way, otherwise, let's keep what we have.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: You're right. This was needed before when the update calls used a combination of the bus voltages and the existing `MAX_VOLT`, `MIN_VOLT` values. Will update.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Hard to say really. Locations will give us more representative renewable profiles, but voltage ranges can help ensure that generators aren't hooked up to substations with more transmission capacity than they should be (will impact the renewable curtailment). I'm leaning towards locations, but could probably be convinced otherwise.

prereise/gather/griddata/hifld/data_process/generators.py
BainanXia: Yeah, I tried it and found the issue: if the value of each cell is `list/tuple`, `.to_numpy()` will give us a 1D array with entries in a weird format, i.e. `array([list([a, b, c]), list([d, e, f])])`. Here is one solution I found from google which works for values being same length across the column: `np.vstack(substations_with_xyz.loc[sub_ids, "xyz"])`.

prereise/gather/griddata/hifld/data_process/generators.py

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Nice! Done.

prereise/gather/griddata/hifld/data_process/generators.py
danielolsen: Also done.

prereise/gather/griddata/hifld/data_process/generators.py

prereise/gather/griddata/hifld/data_process/generators.py

prereise/gather/griddata/hifld/data_process/generators.py

prereise/gather/griddata/hifld/data_process/generators.py

prereise/gather/griddata/hifld/data_process/transmission.py

BainanXia (approved):
I think this is good to go. Thanks!

PR #268, opened 2022-02-11
feat: translate per-length impedance values to whole-line parameters

danielolsen:

### Purpose
Add logic to translate per-distance impedance values to lumped-element values suitable for power system modeling, and to calculate surge impedance loading.

### What the code is doing
A new `Line` class is added, which is a combination of a `Tower` object (which includes all conductor/bundle information) a `length`, a `voltage`, and a `freq` (frequency). With these, impedances which depend on the system frequency are calculated, and the voltage and surge impedance are used to calculate the surge impedance loading. Per-kilometer shunt admittance and series impedance are used along with the `length` parameter to calculate equivalent π-model lumped element parameters using the long-line distributed impedance model. The differences between the long-line model and the short- or medium-line model are insignificant for shorter lines, but important for longer ones, and the computational burden to calculate all of them the more complicated way is low compared to everything else the grid-building code does.

There are also a few minor changes to clean up small lingering things from #262, which impact prints but not core functionality.

### Testing
A new unit test is added which tests this functionality against an example from a power systems textbook.

### Usage Example/Visuals
The new unit test shows how lowest-level parameters are combined to calculate whole-line impedances. This example is single-circuit single-conductor, but other tests demonstrate how multi-circuit and/or bundled conductors can be specified.

### Time estimate
15 minutes to understand the code, longer if you want to crack open your own power system textbook and brush up on the logic or try a few examples.

### DIFF
--- a/prereise/gather/griddata/transmission/geometry.py
+++ b/prereise/gather/griddata/transmission/geometry.py
@@ -280,11 +280,11 @@ class Line(DataclassWithValidation):
                 setattr(self, attr, float(getattr(self, attr)))
         self.validate_input_types()  # defined in DataclassWithValidation
         # Calculate second-order electrical parameters which depend on frequency
-        self.omega = 2 * pi * self.freq
+        omega = 2 * pi * self.freq
         self.series_impedance_per_km = (
-            self.tower.resistance + 1j * self.tower.inductance * self.omega
+            self.tower.resistance + 1j * self.tower.inductance * omega
         )
-        self.shunt_admittance_per_km = 1j * self.tower.capacitance * self.omega
+        self.shunt_admittance_per_km = 1j * self.tower.capacitance * omega
         self.surge_impedance = cmath.sqrt(
             self.series_impedance_per_km / self.shunt_admittance_per_km
         )
--- a/prereise/gather/griddata/transmission/geometry.py
+++ b/prereise/gather/griddata/transmission/geometry.py
@@ -1,3 +1,4 @@
+import cmath
 from dataclasses import dataclass, field
 from itertools import combinations
 from math import exp, log, pi, sqrt
@@ -32,7 +33,7 @@ class Conductor(DataclassWithValidation):
     resistance_per_km: float = None
     gmr: float = None
     area: float = None
-    permeability: float = field(init=False)
+    permeability: float = field(init=False, default=None)
 
     def __post_init__(self):
         # Validate inputs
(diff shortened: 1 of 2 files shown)


prereise/gather/griddata/transmission/geometry.py
```diff
@@ -32,7 +33,7 @@ class Conductor(DataclassWithValidation):
     resistance_per_km: float = None
     gmr: float = None
     area: float = None
-    permeability: float = field(init=False)
+    permeability: float = field(init=False, default=None)
```
BainanXia: I think simply do `permeability: float = None` is equivalent here? I might be wrong.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -250,6 +251,58 @@ def calculate_shunt_capacitance_per_km(self):
...
+    propogation_constant_per_km: complex = field(init=False)
+    surge_impedance: complex = field(init=False)
+    series_impedance: complex = field(init=False)
+    shunt_admittance: complex = field(init=False)
+
+    def __post_init__(self):
+        # Convert integers to floats as necessary
+        for attr in ("freq", "length", "voltage"):
+            if isinstance(getattr(self, attr), int):
+                setattr(self, attr, float(getattr(self, attr)))
```
BainanXia, line 280: For simplicity, maybe we can cast everything into `float` regardlessly. The current implementation is good to go as well.

prereise/gather/griddata/transmission/geometry.py
```diff
@@ -250,6 +251,58 @@ def calculate_shunt_capacitance_per_km(self):
...
+        self.validate_input_types()  # defined in DataclassWithValidation
+        # Calculate second-order electrical parameters which depend on frequency
+        self.omega = 2 * pi * self.freq
```
BainanXia: Not sure whether we would like to define `omega` as an attribute of a `Line` object or it is just something intermediate.

prereise/gather/griddata/transmission/tests/test_geometry.py
```diff
@@ -22,7 +23,6 @@ def test_conductor():
     rated_dc_resistance_per_km = rated_dc_resistance_per_1000_ft * 3.2808
     # Calculate for solid-aluminum (resistance should be lower than rated)
     conductor = Conductor(radius=(outer_diameter / 2), material="aluminum")
-    print(conductor.resistance_per_km)
```
BainanXia, line 25: Good catch:)

BainanXia (approved):
LGTM. Thanks for adding the test. I trust you on the formulas. 

One side note is maybe we should put down the formulas with references of this module in the future. For now, the entire `griddata` module has not been documented yet.

prereise/gather/griddata/transmission/geometry.py
danielolsen: Good point, it's only really intermediate, and anyone who would want to look at the value should know how to calculate it manually. Removed.

prereise/gather/griddata/transmission/geometry.py
danielolsen: I think you're right! Done.

prereise/gather/griddata/transmission/geometry.py
danielolsen: The intention here is to be a little flexible about the inputs before the validation method, but not too flexible. Currently, if we try to pass e.g. a `str` as any of these inputs, the `validate_input_types` method will catch it and return an informative error: `TypeError: for voltage: expected <class 'float'>, but got <class 'str'>`. If we remove the `if` block, the error is less informative: `ValueError: could not convert string to float: 'a'`.

BainanXia: I've been exploring how to support union types in this situation which also works with the validation, but got  `TypeError: Subscripted generics cannot be used with class and instance checks` when using the `Union[int, float]` trick. Let's keep it as it is then.

Issue #259, opened 2022-01-13
Create function which translates profiles defined for one zone set to another zone set


danielolsen:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish I could align zones that were created in one project (e.g. PUMAs for the buildings project, census-defined Urban Areas for the transportation project) with a different set of zone shapes for the grid model (e.g. states, ISOs/RTOs, state+ISOs/RTOs combinations).

### Describe your proposed implementation
The function should take in a set of zones to be transformed (as Shapely geometries), a set of profiles corresponding to those zones, and a set of reference zones, and output profiles which correspond to the reference zones. When a zone-to-be-transformed overlaps with multiple reference zones, the profile should be disaggregated somehow; the simplest way would be based on intersection area, but a more correct way could include some sort of weighting (e.g. by population).

### Describe alternatives you've considered, if relevant
Aligning all zones together within PreREISE is probably the simplest way to include these new profiles into the software framework, but gives the user less flexibility from the Scenario object. For example, if a small urban area is contained entirely within a larger grid model zone, we would not longer be able to specify that we want to model high uptake of electric vehicles just within that zone, but not within others.

PR #270, opened 2022-02-27
data: create remapped shape files and matching fractions

limingzhou2004:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
What is the larger goal of this change?
The code generate a map of Control Areas, ISO subregions, and geographical zones, for spatial load assignment.
### What the code is doing
How is the purpose executed?
From the existing BA_County_map.json, we add the subregions of NYISO, ISONE, ERCOT, CISO, PJM. The CA/subregional level load is matched to this map. A matching table is created to allocate the load based on the overlapping area.
### Testing
How did you test this change (unit/functional testing, manual testing, etc.)? 
To test, run python -m prereise.gather.data.remap_ba_area.remap_ba_area
### Where to look
* It's helpful to clarify where your new code lives if you moved files around or there could be confusion/
All the code and data files are in PreREISE/prereise/gather/data/remap_ba_area.
* What files are most important?
remap_ba_area.py
### Usage Example/Visuals
How the code can be used and/or images of any graphs, tables or other visuals (not always applicable). 
Three files will be generated:  remap_ba_area.shp, zenodo_map_to_ca.csv, remap_ba_area.tif
### Time estimate
How long will it take for reviewers and observers to understand this code change?
20 min

PR #271, opened 2022-03-02
feat: Create function which translates profiles defined for one zone set to another zone set

merrielle:

### Purpose
Addresses https://github.com/Breakthrough-Energy/PreREISE/issues/259
"Create function which translates profiles defined for one zone set to another zone set"

**DEMO JUPYTER NOTEBOOK W/ DATA:**
https://www.dropbox.com/s/srbbkxirykug843/zone_translation.ipynb?dl=0

### What the code is doing
`translate_zone_set()` has three key steps:
- Overlay map of `prev_zones` with `new_zones`
- Pivot table to make it a matrix of `prev_zones` x `new_zones`
- Scale the matrix such that each row sums to 1, then return it

Also included some plotting and results analysis functions. Not sure if these belong in PostREISE?

- `plot_zone_map()` can do both interactive maps and static PNGs
- `filter_interesting_zones()` takes the output of `translate_zone_set()` and filters it to only show when a single zone has been distributed to several new zones

### Testing
TODO

### Where to look
`translate_zone_set()` is the most important

### Usage Example/Visuals
Example of `plot_zone_map()` using Liming's BA shapefiles:
liming_ba_map

Example of `plot_zone_map()` using Merrielle's BA shapefiles:
merrielle_ba_map

Example of `filter_interesting_zones()`, showing translation from Liming's BA shapefiles to Merrielle's BA shapefiles:
**NOTE**: We're rounding the results to three decimal places to make them more readable, so not all rows add up to 1.

Screen Shot 2022-03-01 at 7 25 10 PM

### Time estimate
30 mins?


requirements.txt
```diff
@@ -16,3 +16,6 @@ black
 pytest
 coverage
 pytest-cov
+folium
+geopandas
+rtree
```
danielolsen, line 22: We'll want to add these to the Pipfile so that we can sync all necessary dependencies.

requirements.txt
```diff
@@ -16,3 +16,6 @@ black
 pytest
 coverage
 pytest-cov
+folium
```
danielolsen: It would be nice if we could add this as an optional requirement (and avoiding importing if `interactive=False`), since it seems like it's not strictly necessary.

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+        prev_to_new_matrix["sum"].round(3) > 1
+    ]
+    if len(sum_greater_than_one > 0):
+        print(
+            f"WARNING: Detected rows with a sum greater than one. This most likely means that several {name_new} zones overlap each other."
+        )
+        if verbose:
+            print("\nROWS WITH SUM GREATER THAN ONE:")
+        if verbose:
+            print(sum_greater_than_one.round(3))
```
danielolsen: Can these be combined into one `if` block?

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+    prev_to_new_matrix["to_distribute"] = 1 - prev_to_new_matrix["sum_without_none"]
+    prev_to_new_matrix["scale_factor"] = (
+        1 + prev_to_new_matrix["to_distribute"] / prev_to_new_matrix["sum_without_none"]
+    )
+
+    if verbose:
+        print("\nMATRIX BEFORE SCALING")
+    if verbose:
+        print(prev_to_new_matrix.round(3))
```
danielolsen: Same comment here.

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+        columns = scaled_matrix.columns
+        for zone in list(new_zones[name_new]):
+            if not zone in columns:
+                scaled_matrix[zone] = 0
+
+    if verbose:
+        print(
+            "\nINTERESTING ZONES: rows and cols that include values that are not 1 or 0"
+        )
+        with pd.option_context("display.max_rows", None, "display.max_columns", None):
```
danielolsen, line 218: Nice

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
+import folium
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from folium.features import DivIcon
+
+# IMPORTANT: rtree is required for overlay function, but not imported
```
danielolsen, line 5: I'm guessing we'll get some sort of Exception if we try to use `overlay` without `rtree`? Can we add a `try`/`except` block around the `overlay` call that provides a more direct warning to the user if `rtree` is missing?

BainanXia:
> Also included some plotting and results analysis functions. Not sure if these belong in PostREISE?

I think put these plot functions closer to its main context is good. The reason we have most visualizations in PostREISE is because the package is designed for post analysis of simulation results rather than a place where we should put all visualizations in. Open to discussion.

Not sure whether we will use functionalities here in scenario analysis in the future though. But if that's the case, we will definitely need more than geographical definitions, thus more complicated plots in PostREISE will be implemented anyways.

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def plot_zone_map(gdf, interactive=False):
```
BainanXia: The map looks great in the demo! Not sure whether we would like the plot parameters in this function to be hardcoded (size, color, etc.) or not, given we might want to look into a smaller area rather than the entire nation some day. Maybe this is future work.

BainanXia:
A higher level question for the team: should we put support modules like this in `gather` or should we create another folder, say `utility` next to `gather` as we did in `PowerSimData`?

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+    :return: (*pandas.DataFrame*) -- subset of original matrix
+    """
+    df_copy = df.round(rounding)
+    df_copy = df_copy.replace({1.0: 0.0})
+    df_copy.loc["sum"] = df_copy.sum()  # sum each col
+    df_copy["sum"] = df_copy.sum(axis=1)  # sum each row
+
+    interesting_zones = df_copy.loc[df_copy["sum"] > 0, df_copy.loc["sum"] > 0]
+    interesting_zones = interesting_zones.drop(index=["sum"], columns=["sum"])
+    interesting_zones = interesting_zones.replace({0.0: "-"})
```
BainanXia: Maybe we can do following here:
```python
interesting_zones = df_copy.loc[(df_copy != 0).any(axis=1), (df_copy != 0).any()].replace({0.0: "-"})
```

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+
+    return interesting_zones
+
+
+def format_zone_df(df, name):
+    """copies df, checks if it contains valid geometry, and adds col for zone area
```
BainanXia: `valid` -> `invalid`?

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+    is_valid_df = df_copy.geometry.is_valid.to_frame()
+    invalid_geom = list(is_valid_df.loc[is_valid_df[0] == False].index)
+    if len(invalid_geom) > 0:
+        print(f"WARNING: {name} contains invalid geometries: {invalid_geom}")
+
+    df_copy.index.name = name
+    df_copy = df_copy.reset_index()
+
+    # get zone area
+    df_copy[f"{name}_area"] = df_copy["geometry"].area
```
BainanXia, line 94: What would be the `area` value if the `geometry` is invalid?

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+    :param geopandas.geodataframe.GeoDataFrame df: GeoDataFrame where
+        index = zone names, columns = ['geometry']
+    :param str name: the name of the data set
+    :return: (*geopandas.geodataframe.GeoDataFrame*) -- the formatted df
+    """
+    df_copy = df.copy()
+
+    # Check for invalid geometries
+    is_valid_df = df_copy.geometry.is_valid.to_frame()
+    invalid_geom = list(is_valid_df.loc[is_valid_df[0] == False].index)
```
BainanXia: Not sure whether we can do
```python
invalid_geom = df_copy.loc[~df_copy.geometry.is_valid].index.tolist()
```
here.

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def translate_zone_set(
+    prev_zones, new_zones, name_prev="prev_zones", name_new="new_zones", verbose=False
+):
+    """returns matrix of prev_zones to new_zones. Scales rows proportionally so
+    that each row adds up to 1
+    NOTE: We recommend using EPSG:4269 as your CRS (coordinate reference system).
+    It triggers a warning but it is able to handle invalid geometries, unlike
+    Mercator Projections (EPSG:3395).
```
danielolsen: Do we really want to be recommending `EPSG:4269` here? I think this one is defined in lat/lon, but if we're doing area calculations then I think we want an equal-area projection. As a test, we can look at a lat/lon rectangle, with its bottom edge at the equator, and translate it into two sub-areas, one for the 'lower' half and one for the 'upper' half. We should see that the 'lower' half is larger.

As a side note: do we want to handle invalid geometries?

BainanXia: A follow up question regarding invalid geometries: how do we define invalid geometries in our situation? It seems any shape doesn't form a polygon is treated as invalid from the [documentation](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.is_valid.html), however, we can have bowtie like shapes of areas (BAs or utilities?), right?

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+    # if old zones aren't completely covered by new zones, we have < 100% coverage
+    # Note: we could technically simplify this to just scale_factor = 1/(sum - none) but
+    # the extra cols are nice for verbose mode
+    prev_to_new_matrix["sum_without_none"] = (
+        prev_to_new_matrix["sum"] - prev_to_new_matrix["none"]
+    )
+    prev_to_new_matrix["to_distribute"] = 1 - prev_to_new_matrix["sum_without_none"]
+    prev_to_new_matrix["scale_factor"] = (
+        1 + prev_to_new_matrix["to_distribute"] / prev_to_new_matrix["sum_without_none"]
+    )
```
BainanXia, line 192: Thanks for putting down detailed explanation in doc strings! Just copying our discussions from Slack to here regarding how to redistribute `over counts` to make rows summing up to 1:

> Another approach to bring it down to 1 is looking at the overlaps among columns and distribute the overlap portion uniformly cross columns that causing duplications rather than all areas.

This will make things messy for sure but gives a relative better projection when dealing with self overlaps. I'm good with the current implementation. Just posting the idea here for future references.

prereise/gather/translate_zones.py
```diff
@@ -0,0 +1,229 @@
...
+        columns=["none", "sum", "sum_without_none", "to_distribute", "scale_factor"]
+    )
+
+    # Add any rows that got dropped because they do not overlap. Fill with zeros.
+    # NOTE: not sure if this will get slow with larger datasets
+    if len(scaled_matrix) < len(prev_zones):
+        index = scaled_matrix.index.values
+        for zone in list(prev_zones[name_prev]):
+            if not zone in index:
+                scaled_matrix.loc[zone] = 0
```
BainanXia: Maybe we can do following here:
```python
scaled_matrix.loc[list(set(prev_zones[name_prev]) - set(scaled_matrix.index))] = 0
```
Similar for adding columns below.

prereise/gather/translate_zones.py
merrielle: Sure! That sounds like a good idea.

requirements.txt
merrielle: Honestly it might be better for me to figure out how to do this in Bokeh so we don't need the extra library at all

prereise/gather/translate_zones.py
merrielle: Actually, the exception from geopandas is very explicit and I don't want to accidentally mask other issues that may come up. Is it okay if I leave this as-is?

danielolsen: Nice! I'm too used to unhelpful exceptions. If the `geopandas` one is already good, we can keep it.

prereise/gather/translate_zones.py
merrielle: In the case of a bowtie, the shape is better represented with a MultiPolygon. MultiPolygons are also used for BAs with several discrete areas.

prereise/gather/translate_zones.py
merrielle: I think you're right about this, but it will be quite a bit more work if we want to check for three or more BAs overlapping. To handle three or more simultaneous overlapping BAs we need to recursively check for overlaps until there are none left.

We also need to think about how we handle gaps. For now I think distributing demand proportionally works best to fill in any gaps in the new shapefiles.

Here's an article I've been reading about gaps and overlaps:
https://medium.com/@achm.firmansyah/an-approach-for-checking-overlaps-and-gaps-in-polygons-using-geopandas-ebd6606e7f70

prereise/gather/translate_zones.py
BainanXia: So `is_valid()` will return `True` in this case?

prereise/gather/translate_zones.py
merrielle: I believe assigning with loc only works if you do one row at a time. If one is missing we get:
`"None of [Index(['c'], dtype='object', name='gray_zones')] are in the [index]"`

Using a for loop for now, but let me know if you know a better way!
```python
new_rows = list(set(prev_zones[name_prev]) - set(scaled_matrix.index))
for row in new_rows:
    scaled_matrix.loc[row] = 0
```

danielolsen: `pandas.concat` will let you add an arbitrary number of rows at once, but doesn't modify inplace like a `loc` or `append` addition.

BainanXia: Yeah, you are right. Adding multiple rows in-place in one shot is not supported. Other than using a loop, using `append`/`concat` should work as well, not in-place operation though:
```python
scaled_matrix  = pd.concat([scaled_matrix, pd.DataFrame(0, index=new_rows, columns=scaled_matrix.columns)])
```

merrielle:
Looks like using Pseudo-Mercator does actually affect the results -- mostly by a couple thousandths. Included a quick screenshot of partial new results on top and partial old results on the bottom.

Screen Shot 2022-03-16 at 4 53 00 PM

prereise/gather/translate_zones.py
merrielle: If it's a bowtie as a multipolygon then it's valid. If it's a single polygon that crosses itself, it's invalid.

danielolsen: Do we see any bowties in the current draft of the BA shapes?

BainanXia: I see. So it depends on the object type within the inputs. Not sure whether our existing shapefile runs into this issue though.

requirements.txt
merrielle: Update: I'd like to just ditch the interactive graph. It'd take me a while to refresh my bokeh knowledge and I feel like since we already have a static graph the interactive one is not a huge value add anyway.

prereise/gather/translate_zones.py
merrielle: Okay, so I've tested bowties specifically, and it looks like the area is calculated incorrectly. I've searched for a bunch of different ways to fix invalid polygons, and it seems like there is no single reliable way. I think it's best to throw an exception if any invalid polygon is detected.

Here's a bowtie with "zero" area. :(
Screen Shot 2022-03-17 at 2 35 58 PM

merrielle: Here's the same bowtie as a multipolygon with the area calculated correctly:
Screen Shot 2022-03-17 at 2 49 35 PM

Note that I only found one invalid geometry in Liming's shapefiles -- most of the BAs were already multipolygons.

danielolsen: Is there any sort of helpful utility that converts bowties to multipolygons?

merrielle: I tried a couple different options from stack overflow, and the best I've seen is `make_valid()`. I had looked it up previously, and I thought I saw someone mention some issues with it, but I can't find that post anymore. I tested it with two types of invalid polygons (a bowtie and a flag) and it seemed to work. If we want to use it, we need to add shapely 1.8 as a requirement.

merrielle: Here's a screenshot of my tests if you're interested. I'm not sure if screenshots like this are helpful but I figure it's less work for you than sending over a notebook.
Screen Shot 2022-03-20 at 3 16 44 PM

BainanXia: > I tried a couple different options from stack overflow, and the best I've seen is `make_valid()`. I had looked it up previously, and I thought I saw someone mention some issues with it, but I can't find that post anymore. I tested it with two types of invalid polygons (a bowtie and a flag) and it seemed to work. If we want to use it, we need to add shapely 1.8 as a requirement.

Not sure whether the issue you saw is among the replies of [this](https://stackoverflow.com/questions/20833344/fix-invalid-polygon-in-shapely) post. One of them mentioned dependency conflicts due to shapely 1.8a3 at that time wasn't a stable release, which shouldn't be a problem given 1.8.1 is released now. While, the very last post mentioned a problem that `make_valid` might return a multi-polygon of length greater than 1, which seems to be the normal behavior as expected (not sure whether my read is correct though)? All in all, I think this solution should work unless running into some corner cases which we might be able to fix it case by case.

merrielle: I wonder if we should be trying to fix other people's input data though. Maybe I can throw an error if there's invalid geometry but include a separate helper function that people can use to fix their shapefiles? To me it seems like cleaning the data is a separate concern.

danielolsen: I think an Exception that tells the user why their geometries can't be ingested as passed is fine within this PR, assuming we've got a set of valid geometries that comes out of #270 so that our use case is fully automated.

merrielle:
@BainanXia @danielolsen
Aside from pytest failing (it passes locally so I'm investigating) I think this is ready for a final pass. I've addressed all the PR comments, added tests, and added a helper function `make_zones_valid()` that can be used to clean any imported gdf.

Regarding Bainan's comment about changing how we distribute overlaps:
> Another approach to bring it down to 1 is looking at the overlaps among columns and distribute the overlap portion uniformly cross columns that causing duplications rather than all areas.

I've been investigating this more closely and there are many many edge cases that we may or may not want to handle. The rabbit hole goes deep. I'll send you a summary of some of my findings and address this in a future PR.

danielolsen:
Are the tests failing because we add geopandas to the `Pipfile` but we haven't generated a new `Pipfile.lock`?

merrielle:
Possibly! Updating it now.

danielolsen (approved):
Thanks!

PR #274, opened 2022-03-08
Add pygrib package to requirements file

rouille:

### Purpose
Add missing third party package to requirements file

### What the code is doing
N/A

### Testing
N/A

### Where to look
`pygrib` is added to the *requirements.txt*

### Usage Example/Visuals
N/A

### Time estimate
2min

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ black
 pytest
 coverage
 pytest-cov
+pygrib~=2.1.4


requirements.txt
```diff
@@ -16,3 +16,4 @@ black
 pytest
 coverage
 pytest-cov
+pygrib~=2.1.4
```
BainanXia, line 19: Not familiar with the versions of this package. Is it safe to do `~=` rather than `==`?

danielolsen: `~=` on a patch version _shouldn't_ cause any problems. The Pipfile just specifies `pygrib = "*"` so we're being more cautious here already.

rouille: Good point. I was leaning on the side of some flexibility but I cannot guarantee we won't run into problems with later version. I am fine with locking the version. Any suggestion @jon-hagg?

BainanXia: @danielolsen I checked `Pipfile.lock` which chooses 2.1.4 and brought up the conversation. Maybe it doesn't matter.

danielolsen (approved):

requirements.txt
jenhagg: I think `~=` is a good middle ground - safer than `*` and less work to maintain than `==`. That's what i'd use by default, and only use `==` for special cases or unstable packages.

BainanXia (approved):

PR #275, opened 2022-03-09
Update packages

jenhagg:

### Purpose
Original goal was to bump pandas to be consistent with powersimdata. This required installing openpyxl which is now used to read excel files instead of xlrd (maybe this can be removed?), and updating the version of python-dateutil. There were also a bunch of warnings to fix, and 2 tests failing in `test_decompose_profile.py`. I fixed these by setting the index to plant_id, but that caused other tests in that file to fail (if done at the module level), so I could use some help to determine if that's fine, or there is a better approach. 

### What the code is doing
See above. Other changes are mainly to omit the code coverage locally while still having it run by github actions.

### Testing
tox

### Time estimate
10-20 min?


prereise/gather/hydrodata/eia/tests/test_decompose_profile.py
```diff
@@ -76,7 +77,8 @@ def test_get_profile_by_plant_argument_value():
 
 
 def test_get_profile_by_plant():
-    plant_df = mock_plant_df[mock_plant_df.type == "hydro"]
+    mdf = mock_plant_df.set_index("plant_id")
+    plant_df = mdf[mdf.type == "hydro"]
```
BainanXia: Not sure why this test passed for old version of `pandas` maybe it sets the index as the first column by default? Looking through `get_profile_by_plant`, we do need the index of `plant_df` to be `plant_id`, otherwise it will give profile data frame with wrong column names (plant ids). We can still make it one liner though
```python
plant_df = mock_plant_df.set_index("plant_id").query("type == 'hydro'")
```

The reason it fails other tests if we change it at the module level is because `MockGrid` will look for `plant_id` column and set it as index when building up plant data frame of an attribute of the mock grid. Hence it won't work if `plant_id` is already the index.

prereise/gather/hydrodata/eia/tests/test_decompose_profile.py
```diff
@@ -102,7 +104,8 @@ def test_get_normalized_profile_argument_value():
 
 
 def test_get_normalized_profile():
-    plant_df = mock_plant_df[mock_plant_df.type == "hydro"]
+    mdf = mock_plant_df.set_index("plant_id")
+    plant_df = mdf[mdf.type == "hydro"]
```
BainanXia: Same comment as above.

prereise/gather/solardata/helpers.py
```diff
@@ -29,7 +29,7 @@ def to_reise(data):
         if i == 1:
             profile = data_tmp.T
         else:
-            profile = profile.append(data_tmp.T, sort=False, ignore_index=True)
+            profile = pd.concat([profile, data_tmp.T], sort=False, ignore_index=True)
```
BainanXia, line 32: Just curious, what does warning it produce here? It seems to be equivalent.

BainanXia (approved):
LGTM. Thanks for the fixes!

prereise/gather/solardata/helpers.py
danielolsen: `append` seems to deprecated in favor of `concat`.

rouille: I think @lanesmith encountered the same issue.

danielolsen:
> This required installing openpyxl which is now used to read excel files instead of xlrd (maybe this can be removed?)

It seems that we still use Excel-file-reading in two places:
- Reading data from a copy of EIA Form 923 data in the repo (`EIA923_Schedules_2_3_4_5_M_12_2016_Final_Revision.xlsx`), used for generating hydro profiles (?)
- Reading Excel files downloaded from the EIA as part of demand profile building (`prereise.gather.demanddata.eia.get_eia_data`).

I think both of these functions _could_ be refactored inplace to perform the same functions without Excel, by converting the sheet(s) of EIA Form 923 that we care about into CSVs, and by grabbing EIA data in JSON format via their API rather than as Excel files, but we are also in the process of developing replacements to these anyway so I'm not sure whether it's worth it.

rouille:
> > This required installing openpyxl which is now used to read excel files instead of xlrd (maybe this can be removed?)
> 
> It seems that we still use Excel-file-reading in two places:
> 
>     * Reading data from a copy of EIA Form 923 data in the repo (`EIA923_Schedules_2_3_4_5_M_12_2016_Final_Revision.xlsx`), used for generating hydro profiles (?)
> 
>     * Reading Excel files downloaded from the EIA as part of demand profile building (`prereise.gather.demanddata.eia.get_eia_data`).
> 
> 
> I think both of these functions _could_ be refactored inplace to perform the same functions without Excel, by converting the sheet(s) of EIA Form 923 that we care about into CSVs, and by grabbing EIA data in JSON format via their API rather than as Excel files, but we are also in the process of developing replacements to these anyway so I'm not sure whether it's worth it.

It seems that `xlrd` no longer allow to read anything but *xls* files. This means that the loading of  *xlsx* files into data frame through `pandas.read_excel` will fail depending of the version of `xlrd` installed. We likely need refactor the code to use `openpyxl` which seems to be more flexible. See list of files below where `pandas.read_excel` is used:
```
prereise/gather/demanddata/eia/demo/eastern_demand_v6_demo/spp/split_spp_into_subareas.ipynb:    "county_list_owner = pd.read_excel('spp_counties_owners.xlsx', index_col=0)\n",
prereise/gather/demanddata/eia/demo/eastern_demand_v6_demo/miso/generate_miso_subarea_profile.ipynb:    "lrz_demand_2016 = pd.read_excel('20161231_dfal_hist.xls', skiprows=5)\n",
prereise/gather/demanddata/eia/demo/eastern_demand_v6_demo/miso/generate_miso_subarea_profile.ipynb:    "lrz_demand_2015 = pd.read_excel('20151231_dfal_hist.xls', skiprows=5)\n",
prereise/gather/demanddata/eia/get_eia_data.py:        df = pd.read_excel(
prereise/gather/helpers.py:    eia_form_923 = pd.read_excel(
prereise/gather/hydrodata/eia/demo/eastern_hydro_v3_demo/eastern_hydro_v3_demo.ipynb:    "eastern_hps = pd.read_excel(io='./hps_plants_eastern.xlsx',sheet_name = 'all_plantIDs',header = 0)\n",
```

jenhagg:
> We likely need refactor the code to use openpyxl which seems to be more flexible. 

Seems like pandas will automatically use openpyxl if installed. One of the tests was actually failing and was fixed by doing this.

rouille:
> > We likely need refactor the code to use openpyxl which seems to be more flexible.
> 
> Seems like pandas will automatically use openpyxl if installed. One of the tests was actually failing and was fixed by doing this.

Then, I would say we can remove `xlrd` from the requirements

PR #276, opened 2022-03-10
feat: use transmission line physical parameters to calculate impedances and ratings for HIFLD grid

danielolsen:

### Purpose
To build branch ratings and impedances for the HIFLD grid, use assumptions which start with the physical parameters of typical line designs (with non-typical overrides as necessary), rather than relying on processed averages from the TAMU grid.

### What the code is doing
- We add a table of 'typical' tower/conductor designs (`line_designs.csv`). Within `transmission.py`, the default choice for a given voltage level is the single-circuit design with the minimum number of conductors per bundle.
- Within `const.py`:
  - we add the system power base that will be used to calculate per-unit values (`s_base`).
  - ~we make a change to the impedance of 161/500 kV transformers. The existing value produces severe transmission infeasibilities, while the new value fixes them (and is consistent with examination of real world data).~ **EDIT**: removed for now in favor of more extensive transformer re-working within #286.
  - we add a new mapping of lines to voltage (`line_voltage_assumptions`), to be used to override the algorithms within `transmission.py` when problems are encountered.
  - we add a set of assumptions about lines which should be built using the non-default line designs.
- Within `transmission.py`:
  - We apply the line voltage assumptions from `const.py` before performing the neighbor search algorithm to assign voltage to unknown-voltage lines.
  - We add a new function `add_impedance_and_rating` which builds `Tower` objects for each of the designs in `line_designs.csv`, and associates each transmission line in the HIFLD dataset with one of these designs (using either the default or the overrides in `const.py`) before passing this augmented dataframe (original data plus Tower designs) to the existing lower-level `estimate_branch_impedance` and `estimate_branch_rating` functions.
  - We modify the `estimate_branch_impedance` and `estimate_branch_rating` functions to use the `Tower` objects plus the line lengths to calculate the appropriate values for transmission lines (transformer logic is unchanged).
- **EDIT**: within `test_transmission.py`: several new tests aimed at improving the overall test coverage of the `transmission` module.

### Testing
Tested manually to build the HIFLD grid. Preliminary tests show the built grid having much less transmission violation energy and many fewer lines violating, compared to when the grid is built using the assumed values from the TAMU data. Full results to be added later.

~Unit tests still need to be updated.~ EDIT: done.

### Usage Example/Visuals
All code changes will be used by calling the `build_transmission` function (either directly or via `create_csvs` or `create_grid` in the `orchestration` module).

### Time estimate
30 minutes to an hour, provided they already have a good understanding of the line parameter calculation logic, or are willing to trust the previous reviews of that code (#272). ~This branch depends on #272, so is a draft for now.~

### DIFF
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -594,6 +594,10 @@ line_interconnect_assumptions = {
     "ERCOT": {305330},
 }
 
+line_voltage_assumptions = {
+    128842: 230,
+}
+
 line_design_assumptions = {
     # branch index: (voltage, circuits, bundle count)
     101536: (161, 2, 4),  # TVA Cumberland (between substations)
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -125,43 +125,6 @@ volt_class_defaults = {
     "500": 500,
 }
 
-line_reactance_per_mile = {  # per-unit
-    69: 0.0096,
-    100: 0.0063,
-    115: 0.0039,
-    138: 0.0026,
-    161: 0.0021,
-    230: 0.0011,
-    345: 0.0005,
-    500: 0.0002,
-    765: 0.0001,
-}
-line_rating_short = {  # MVA
-    69: 86,
-    100: 181,
-    115: 239,
-    138: 382,
-    161: 446,
-    230: 797,
-    345: 1793,
-    500: 2598,
-    765: 5300,
-}
-line_rating_short_threshold = 50  # miles
-line_rating_surge_impedance_loading = {  # MVA
-    69: 13,
-    100: 30,
-    115: 35,
-    138: 50,
-    161: 69,
-    230: 140,
-    345: 375,
-    500: 1000,
-    765: 2250,
-}
-line_rating_surge_impedance_coefficient = 43.261
-line_rating_surge_impedance_exponent = -0.6678
-
 transformer_reactance = {  # per-unit
     (69, 115): 0.14242,
     (69, 138): 0.10895,
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -1,3 +1,5 @@
+s_base = 100  # MVA
+
 mile_to_meter = 1609.34
 meter_to_mile = 1.0 / mile_to_meter
 
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -594,6 +594,43 @@ line_interconnect_assumptions = {
     "ERCOT": {305330},
 }
 
+line_design_assumptions = {
+    # branch index: (voltage, circuits, bundle count)
+    101536: (161, 2, 4),  # TVA Cumberland (between substations)
+    103726: (161, 2, 4),  # TVA Cumberland (plant to substation)
+    110220: (115, 1, 2),  # Riverside Generating Station
+    112166: (230, 1, 2),  # Waterford 3 Nuclear Generating Station
+    116493: (230, 1, 2),  # Herbert A. Wagner Generating Station
+    132671: (230, 1, 3),  # Florida Power & Light St. Lucie
+    130105: (161, 1, 3),  # TVA Cumberland (downstream)
+    132804: (230, 1, 2),  # Waterford 3 Nuclear Generating Station
+    139216: (230, 4, 3),  # Catawba Nuclear Station
+    140651: (230, 1, 3),  # Florida Power & Light St. Lucie
+    147054: (115, 1, 2),  # RE Ginna Nuclear Power Plant
+    158263: (100, 1, 4),  # G. G. Allen Steam Station
+    160723: (230, 1, 3),  # Florida Power & Light St. Lucie
+    203381: (230, 1, 2),  # Midpoint to Boise Bench
+    203382: (230, 1, 2),  # Midpoint to Boise Bench
+    203383: (230, 1, 2),  # Midpoint to Boise Bench
+    300029: (69, 2, 1),  # Ormat Brawley geothermal
+    301320: (138, 2, 2),  # Collin/Frisco substation connection
+    301847: (115, 1, 2),  # Coso Power Station
+    303557: (138, 2, 2),  # West Levee substations connection (Dallas)
+    304908: (69, 2, 1),  # Heber geothermal
+    305018: (138, 2, 2),  # TH Wharton Generating Station
+    305464: (230, 2, 2),  # The Geysers
+    305990: (100, 2, 1),  # southeast Salton Sea geothermal plants (several)
+    306949: (230, 2, 2),  # The Geysers
+    307569: (230, 2, 2),  # The Geysers
+    308434: (230, 2, 2),  # The Geysers
+    309268: (69, 2, 1),  # Ormat Brawley geothermal
+    309441: (69, 2, 2),  # substation connection near Lake Creek Reservoir, outside Waco
+    309966: (138, 2, 2),  # Garfield to Pilot Knob (south of Austin)
+    312241: (100, 2, 1),  # southeast Salton Sea geothermal plants (several)
+    313473: (69, 2, 2),  # Holtville geothermal?
+    313474: (69, 2, 2),  # Holtville geothermal?
+}
+
 b2b_ratings = {  # MW
     "BLACKWATER TIE": 200,  # a.k.a. 'Clovis'/'Roosevelt County' (Eastern/Western)
     "EDDY AC-DC-AC TIE": 200,  # a.k.a. 'Artesia' (Eastern/Western)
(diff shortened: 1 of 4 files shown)


danielolsen:
With #272 merged and tests updated, this branch is now ready for a full review.

prereise/gather/griddata/hifld/const.py
```diff
@@ -1,3 +1,5 @@
+s_base = 100  # MVA
```
rouille, line 1: Why not using `baseMVA` as we do in **PowerSimData**?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -813,14 +902,9 @@ def build_transmission(method="line2sub", **kwargs):
     first_new_id = ac_lines.index.max() + 1
     transformers.index = pd.RangeIndex(first_new_id, first_new_id + len(transformers))
     ac_lines["type"] = "Line"
-    ac_lines["length"] = ac_lines.apply(calculate_branch_mileage, axis=1)
+    ac_lines["length"] = ac_lines.apply(calculate_branch_mileage, axis=1) * 1.6093  # km
```
rouille: should we add `mile_to_kilometer` to the `const` module?

prereise/gather/griddata/hifld/const.py
danielolsen: Good idea.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: It's already there in `prereise.gather.griddata.transmission.const`. Done.

prereise/gather/griddata/hifld/const.py
danielolsen: Done. We also currently use `s_base` elsewhere in PreREISE though, and it follows a more pythonic naming convention, whereas I think the `baseMVA` in PowerSimData originates with Matpower. Not sure which approach is best.

rouille: Yes it originates from Matpower. If we already use `s_base` elsewhere in PreREISE then my consistency argument is no longer valid and I let you make an executive decision

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -535,11 +557,102 @@ def create_transformers(bus):
     return pd.DataFrame(bus_pairs, columns=["from_bus_id", "to_bus_id"])
 
 
+def add_impedance_and_rating(branch, bus_voltages, line_overrides={}):
```
BainanXia: Do we want to initialize `line_overrides` as None following the conventions?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -535,11 +557,102 @@ def create_transformers(bus):
...
+        {
+            voltage: (voltage, 1, group["bundle_count"].min())
+            for voltage, group in tower_designs.groupby("voltage")
+        }
+    )
+    # Find the default line design for each voltage
+    closest_voltage_design = {
+        v: default_designs.iloc[default_designs.index.get_loc(v, method="nearest")]
+        for v in branch["VOLTAGE"].unique()
+        if not pd.isnull(v)
```
BainanXia: We can reduce a line here by `for v in branch["VOLTAGE"].dropna().unique()` 😉

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -535,11 +557,102 @@ def create_transformers(bus):
...
+    }
+    # Add meaningful index for easier lookups using design tuples
+    tower_designs.set_index(["voltage", "circuits", "bundle_count"], inplace=True)
+    # Map each transmission line to its corresponding Tower design
+    branch_plus_lines = branch.assign(
+        Line=branch.query("type == 'Line'").apply(
```
BainanXia: Shall we use `LINE` as column name to reduce confusion here?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -535,11 +557,102 @@ def create_transformers(bus):
...
+            lambda x: Line(
+                tower=tower_designs.loc[
+                    line_overrides[x.name]
+                    if x.name in line_overrides
+                    else closest_voltage_design[x["VOLTAGE"]],
```
BainanXia: Maybe we could do `line_overrides.get(x.name, closest_voltage_design([x["VOLTAGE"]))` here, which works for both `pandas.Series` and `dict`.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -535,11 +557,102 @@ def create_transformers(bus):
     return pd.DataFrame(bus_pairs, columns=["from_bus_id", "to_bus_id"])
 
 
+def add_impedance_and_rating(branch, bus_voltages, line_overrides={}):
+    """Estimate branch impedances and ratings based on voltage and length (for lines),
+    and add these to the branch data frame (modified inplace).
+
+    :param pandas.DataFrame branch: branch data table. Required columns are:
+        'VOLTAGE' (in kV) and 'length' (in km).
```
BainanXia, line 565: We might want to mention the index of `branch` data frame should be branch ids.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -559,12 +672,8 @@ def _euclidian(a, b):
         )
         return const.transformer_reactance[closest_voltage_tuple]
     elif branch.loc["type"] == "Line":
```
BainanXia, line 673: This is outside this PR: given branch is a series, I believe we don't need `loc` here. Same comment below.

prereise/gather/griddata/hifld/data/line_designs.csv
```diff
@@ -0,0 +1,23 @@
...
+161,2,4,Bluejay,0.5,-5,15,-5,17,-5,19,5,19,5,17,5,15,,,,,,,,,,,,,Two parallel single-poles
+230,1,1,Falcon,,-7,20,0,20,7,20,,,,,,,,,,,,,,,,,,,H-frame design
+230,1,2,Drake,0.5,-7,20,0,20,7,20,,,,,,,,,,,,,,,,,,,H-frame design
+230,1,3,Drake,0.5,-7,20,0,20,7,20,,,,,,,,,,,,,,,,,,,H-frame design
+230,2,2,Drake,0.5,-3,19,-3,17,-3,15,3,15,3,17,3,19,,,,,,,,,,,,,Double-circuit lattice
+230,4,3,Drake,0.5,-7,20,0,20,7,20,18,20,25,20,32,20,43,20,50,20,57,20,68,20,75,20,82,20,Four parallel H-frames
+345,1,2,Falcon,0.5,-8,25,0,25,8,25,,,,,,,,,,,,,,,,,,,H-frame design
+500,1,3,Rail,0.5,-10,30,0,30,10,30,,,,,,,,,,,,,,,,,,,lattice
+500,2,3,Drake,0.5,-5,40,-8,30,-5,20,5,20,8,30,5,40,,,,,,,,,,,,,lattice
+765,1,4,Martin,0.5,-15,40,0,40,15,40,,,,,,,,,,,,,,,,,,,H-frame design
```
BainanXia, line 23: Should we mention the data source somewhere and/or do we need to do data-intake for this?

BainanXia (commented):
This is novelty in grid building.

prereise/gather/griddata/hifld/data/line_designs.csv
danielolsen: These are self-created, based on compiling a variety of sources and rough assumptions, so I don't think we need to go through the data intake process. I'm going to write more about this in the paper describing our process. Do you think we need more documentation in the repo itself?

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: That's true. This is taken care of in #286, so it should become a moot point soon.

prereise/gather/griddata/hifld/data_process/transmission.py

prereise/gather/griddata/hifld/data/line_designs.csv
BainanXia: Sounds good. We will definitely put some paragraphs from the paper into the readme of the repo later and reference the paper for sure.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Less code! Done.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I see what you mean. I think in the past I've tried to add new columns without all caps, to distinguish from the source data columns which are all caps. I was using `Line` to denote that what's in the column is a `Line` object. What do you think about some thing like `line_object` instead?

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Nice! Done.

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: Anything other than the same name with `Line` class should be fine.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: I _think_ that this function doesn't actually care what's in the index of this dataframe, as long any entries in `line_overrides` correspond to indices in the frame. I added some text to the docstring to attempt to clarify this. Do you think it makes sense?

BainanXia: That's exactly what I mean to.

BainanXia (approved):
Thanks for adding those tests!

danielolsen:
> Thanks for adding those tests!

I was bored on a plane with no wifi.

PR #281, opened 2022-03-25
Clean up tests

rouille:

### Purpose
Clean up ***tox.ini*** file and fix a SettingWithCopyWarning raised by a test.

### What the code is doing
N/A

### Testing
Run successfully `tox`

### Where to look
* Remove unnecessary environment variables
* Combine a chained operation into a single operation using `iloc` so that we can ensure the original data frame is set

### Usage Example/Visuals
N/A

### Time estimate
2 min


BainanXia (approved):
Thanks!

PR #283, opened 2022-03-29
feat: when loading HIFLD transmission lines, join disconnected line segments

danielolsen:

### Purpose
The HIFLD transmission line paths are defined by either: a list of lat/lon points (which I'll call a 'segment' from hereon out), or a list of such lists. The Jul 2020 dataset values seem to _always_ be a list of segments, but the February 2020 dataset is either a single segment (not wrapped in a list) or a list of segments.

Previously, when we had a list of segments, we always grabbed the first one, discarding all the rest. This PR changes the logic so that _if_ we get a list of segments of length greater than one, we make an assumption about how to reconnect them. The logic is a little complex because the segments aren't 'drawn' in any particular order, don't necessary connect, and there are some small noisy segments that need to be discarded.

### What the code is doing
Within `helpers.py`, we copy over a couple of helper functions for calculating distances between coordinate pairs.

Within `load.py`, we add a new function `_join_line_segments` to take a list of segments and parse them.
- There is a private function `_generate_linear_spanning_tree` which:
  - creates a graph with each segment's endpoints as nodes and an edge between the start and end of each segment
  - creates edges between each endpoint of each segment and each endpoint of each other segment, weighted by distance
  - find the minimum spanning tree of the graph
  - raises an error if the minimum spanning tree isn't a chain (not sure if this is the right graph theory terminology or not)
  - returns the minimum spanning tree if not
- The list of segments is passed to `_generate_linear_spanning_tree`, and if the tree isn't a chain, the smallest segment is dropped and the process is repeated until we have a chain tree.
- The final list of segments is processed to put all of the coordinate pairs into the proper order, by traversing the tree and flipping the original segments if they go from 'end' to 'start' rather than 'start' to 'end'.

### Testing
Tested manually. Calling `get_hifld_electric_power_transmission_lines` works, as well as `create_grid`.

### Time estimate
30-60 minutes. The code is not too complex, but the logic is, since it needs to handle many edge cases found in the source data.

### DIFF
--- a/prereise/gather/griddata/hifld/data_access/load.py
+++ b/prereise/gather/griddata/hifld/data_access/load.py
@@ -1,3 +1,4 @@
+import itertools
 import json
 import os
 import tempfile
@@ -6,6 +7,7 @@ from urllib.parse import urlparse
 from urllib.request import urlopen
 from zipfile import ZipFile
 
+import networkx as nx
 import pandas as pd
 from tqdm import tqdm
 
@@ -14,6 +16,7 @@ from prereise.gather.griddata.hifld.const import (
     contiguous_us_bounds,
     heat_rate_estimation_columns,
 )
+from prereise.gather.helpers import angular_distance, latlon_to_xyz
 
 
 def get_eia_form_860(path):
@@ -186,6 +189,94 @@ def get_hifld_electric_substations(path):
     )
 
 
+def _join_line_segments(segments, threshold=0.001):
+    """Given a list of line segments, join them together as best as possible.
+
+    :param list segments: list of segments, each defined as a list of (lon, lat)
+        coordinate pairs.
+    :param float threshold: minimum angular distance for a segment to be considered.
+        Defined as the angular distance from the segment start to segment end (degrees).
+    :return: (*list*) -- consolidated list of points.
+    """
+
+    def _generate_linear_spanning_tree(segments):
+        """Given a list of segments, create a minimum spanning tree and ensure that it
+        is linear (i.e. there are two nodes of degree 1, all other nodes are degree 2).
+
+        :param list segments: list of segments, each defined as a list of (lon, lat)
+            coordinate pairs.
+        :return: (*networkx.Graph*) -- a minimum spanning tree from the segments.
+        :raises ValueError: if the minimum spanning tree is not linear
+        """
+        # Build a Graph representing the existing connections between segment endpoints
+        g = nx.Graph()
+        # add a 'start' and 'end' node for each segment, and a zero-cost edge b/w them
+        for i in range(len(segments)):
+            g.add_edge((i, "start"), (i, "end"), weight=0)
+        # Add edge weights by the angular distance between different segment endpoints
+        for (seg1, which1), (seg2, which2) in itertools.combinations(g.nodes, 2):
+            if seg1 == seg2:
+                # no edge needed, since there's already a zero-cost edge there
+                continue
+            # Extract latitude and longitude from the appropriate end of each segment
+            lon1, lat1 = segments[seg1][0 if which1 == "start" else -1]
+            lon2, lat2 = segments[seg2][0 if which2 == "start" else -1]
+            edge_angle = angular_distance(
+                latlon_to_xyz(lat1, lon1), latlon_to_xyz(lat2, lon2)
+            )
+            g.add_edge((seg1, which1), (seg2, which2), weight=edge_angle)
+        # Identify edges to connect the segments
+        tree = nx.algorithms.tree.minimum_spanning_tree(g)
+        # Ensure that the minimum spanning tree is linear
+        if max(dict(nx.degree(tree)).values()) > 2:
+            raise ValueError("The minimum spanning tree isn't linear")
+        return tree
+
+    if len(segments) == 1:
+        # There's only one segment in the list, so no joining is necessary
+        return segments[0]
+
+    segments_df = pd.Series(segments).to_frame(name="segments")
+    segments_df["distance"] = segments_df["segments"].map(
+        lambda x: angular_distance(
+            latlon_to_xyz(x[0][1], x[0][0]),  # first point
+            latlon_to_xyz(x[-1][1], x[-1][0]),  # last point
+        )
+    )
+    filtered_segments = segments_df.loc[segments_df["distance"] > threshold]
+    if len(filtered_segments) == 1:
+        # There's only one segment left in the list, so no joining is necessary
+        return filtered_segments["segments"].iloc[0]
+
+    # Sort the segments by length, in case the minimum spanning tree isn't linear
+    sorted_segments = filtered_segments.sort_values("distance")
+    # Make a minimum spanning tree, dropping small segments as necessary for linearity
+    for i in range(len(sorted_segments)):
+        try:
+            linear_segments = sorted_segments.iloc[i:]["segments"].tolist()
+            tree = _generate_linear_spanning_tree(linear_segments)
+            # We found a good minimum spanning tree, so we can exit
+            break
+        except ValueError:
+            # The minimum spanning tree wasn't good, go to the next iteration
+            pass
+
+    # Identify an endpoint of the linear minimum spanning tree
+    node_degrees = dict(nx.degree(tree))
+    endpoint = min(k for k, v in node_degrees.items() if v == 1)
+    # Traverse the tree, adding segments in order
+    joined_segments = []
+    traversal = nx.algorithms.traversal.dfs_edges(tree, source=endpoint)
+    for (seg1, which1), (seg2, _) in traversal:
+        if seg1 == seg2:
+            if which1 == "start":
+                joined_segments += linear_segments[seg1]
+            else:
+                joined_segments += linear_segments[seg1][::-1]
+
+    return joined_segments
+
+
 def get_hifld_electric_power_transmission_lines(path):
     """Read the HIFLD Electric Power Transmission Lines json zip file and keep in
     service lines.
@@ -216,8 +307,14 @@ def get_hifld_electric_power_transmission_lines(path):
     )
 
     properties["COORDINATES"] = [
-        line["geometry"]["coordinates"][0] for line in data["features"]
+        line["geometry"]["coordinates"] for line in data["features"]
     ]
+
+    # Join multiple distinct segments as necessary to get one single list
+    properties["COORDINATES"] = properties["COORDINATES"].map(
+        lambda x: _join_line_segments(x) if not isinstance(x[0][0], float) else x
+    )
+
     # Flip [(lon, lat), (lon, lat), ..] points to [(lat, lon), (lat, lon), ...]
     properties["COORDINATES"] = properties["COORDINATES"].map(
         lambda x: [y[::-1] for y in x]
--- a/prereise/gather/griddata/hifld/data_access/tests/test_load.py
+++ b/prereise/gather/griddata/hifld/data_access/tests/test_load.py
@@ -1,0 +1,78 @@
+from prereise.gather.griddata.hifld.data_access.load import _join_line_segments
+
+
+def test_join_line_segments_single_segment():
+    # Seattle to Tacoma via Seatac
+    segments = [[(-122.34, 47.60), (-122.30, 47.44), (-122.44, 47.25)]]
+    assert _join_line_segments(segments) is segments[0]
+
+
+def test_join_line_segments_single_segment_after_filtering():
+    segments = [
+        # Long line with extremely similar start & end in Tacoma, should get filtered
+        [(-122.45, 47.24), (-122.90, 47.04), (-122.45, 47.240001)],
+        # Seattle to Tacoma via Seatac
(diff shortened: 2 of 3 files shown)


BainanXia:
This is interesting! I'm wondering the performance of this algorithm. What is the maximum number of segments for a line in the dataset? We are creating a pretty dense graph by adding edges between all end-points of segments for a line. The overall time complexity is bounded by O(N*V^2logV) where N is the number of lines, V is the number of maximum segments. Not sure how long it takes to go through the current transmission dataset.

Just another random thought (might be a bad idea;)): assuming all the line segments roughly form a straight line rather than circles, would it be sufficient to project one of the end points of each segment onto an axis in a 3D cartesian coordinates system (`latlong_to_xyz`) to get some sort of partial ordering? For example, consider three segments with start points being (x1, y1, z1), (x2, y2, z2), (x3, y3, z3), sort the points based on the projections of the axis with maximum span (suppose `max(x1, x2, x3) - min(x1, x2, x3) > max(y1, y2, y3) -min(y1, y2, y3) > max(z1, z2, z3) - min(z1, z2, z3)`, then x axis is chosen). In this way, a partial ordering of segments is determined. It won't tell us the start/end info and not able to fill gaps between segments though. Also, not sure how it performs over the noisy source data.

danielolsen:
For the July 2020 dataset, we have 89,744 total lines, of which 80 need joining. On my laptop, this step takes about three seconds to get through all the lines (including the single-segment lines where we're just identifying that the list has length one and returning the first element). The vast majority of these lines have either 2 or 3 segments, although there's one line with 5, one line with 10, and one line with 33. I believe that when I was looking at the February 2022 dataset, I didn't see a 33-segment line, but I did see a 14-segment line (maybe in its place?).

BainanXia:
@danielolsen Thanks for looking into the statistics, <0.1% is super light. Then I presume this is minor. Probably it isn't worth the time to explore other possibilities. Given what we have works for now, let's stick with it.

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -186,6 +189,104 @@ def get_hifld_electric_substations(path):
...
+        for i in range(len(segments)):
+            (lon1, lat1), (lon2, lat2) = segments[i][0], segments[i][-1]
+            segment_angular_distance = angular_distance(
+                latlon_to_xyz(lat1, lon1), latlon_to_xyz(lat2, lon2)
+            )
+            if segment_angular_distance < threshold:
+                # Segment is too small, so we won't add this to the graph
+                continue
+            g.add_nodes_from([(i, "start"), (i, "end")])
+            g.add_edge((i, "start"), (i, "end"), weight=0)
```
rouille, line 215: Looking at the documentation of `add_edge`, it seems that the nodes will be automatically added if they are not already in the graph. So I guess we can discard the previous line.

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -186,6 +189,104 @@ def get_hifld_electric_substations(path):
...
+            tree = _generate_linear_spanning_tree(linear_segments)
+            # We found a good minimum spanning tree, so we can exit
+            break
+        except ValueError:
+            # The minimum spanning tree wasn't good, go to the next iteration
+            pass
+
+    if len(segments) == 1:
+        # There's only one segment in the list, so no joining is necessary
+        return segments[0]
```
rouille: Should this condition go at the top, right after the docstring of `_join_line_segments`?

danielolsen: It's down here because we may have one segment to start, or we may have one segment after filtering out small segments within `_generate_linear_spanning_tree`. Maybe it would be clearer to move that step out of `_generate_linear_spanning_tree`, and then check after filtering but before tree generation?

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -186,6 +189,104 @@ def get_hifld_electric_substations(path):
...
+
+    # Identify an endpoint of the minimum spanning tree
```
rouille: How will the code below behave if `tree` is made of one segment, i.e., all the other ones in the line have been dropped to fulfill the linearity condition?

prereise/gather/griddata/hifld/data_access/load.py
rouille: the condition is on `segments` (the input parameter) not on `linear_segments`

danielolsen: Ah, yeah, good catch. I think it should be on `linear_segments`.

rouille: If the number of single segment line in the dataset is not ridiculous, we could have:
* one condition on `segments` at the top and the one on `linear_segments` here
*  in `get_hifld_electric_power_transmission_lines` filter the unique segment line before the mapping

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: I think if we update above to operate on `linear_segments` instead of `segments`, a single-segment tree should never make it here.

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: Single-segment lines are the vast majority in the latest version of the dataset, ~99.9%. Multi-segment lines are the exceptions.

rouille: I guess we can speed up the code by avoiding going into `_generate_linear_spanning_tree` if we don't have to.

prereise/gather/griddata/hifld/data_access/load.py
rouille: I agree.

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: Okay, this is updated in the latest commit. I added some tests as well that check whether the returned object `is` the entry from the original list, to ensure that we're not entering the tree-building and joining logic when we don't have to, plus a couple other tests for what happens when we do enter the tree-building/joining logic.

**EDIT**: testing locally, I see a 10-15x speedup, from 3 seconds to get through all lines down to 0.2.

rouille: Thanks

rouille (approved):
Very nice

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -186,6 +189,102 @@ def get_hifld_electric_substations(path):
...
+    # Make a minimum spanning tree, dropping small segments as necessary for linearity
+    for i in range(len(sorted_segments)):
+        try:
+            linear_segments = sorted_segments.iloc[i:]["segments"].tolist()
+            tree = _generate_linear_spanning_tree(linear_segments)
+            # We found a good minimum spanning tree, so we can exit
+            break
+        except ValueError:
+            # The minimum spanning tree wasn't good, go to the next iteration
+            pass
```
BainanXia, line 262: Another random thought to get more predictable results here: I'm wondering for invalid trees (nonlinear) whether it is always the case that there is main linear branch plus a few small branches along the way instead of multiple main branches. If so, probably we can remove nodes with [betweenness_centrality](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.betweenness_centrality.html) being 1 which connect to a node with `betweenness_centrality` being greater or equal to 3.

danielolsen: It seems that this approach could produce ties in this `betweenness_centrality` metric. What would we do there? In most of the cases in the latest dataset where we connect segments, there are either two segments (can't produce a non-linear tree) or three, which would tend to produce either a linear tree or a Y-shaped tree with some level of symmetry. Plus, the structure of the graph in how it connects three segments whose ends are all very close could be very sensitive to a small movement in the coordinates of one of the endpoints.

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -186,6 +189,102 @@ def get_hifld_electric_substations(path):
...
+        (seg1, which1), (seg2, _) = endpoint, next_endpoints[0]
+        if seg1 == seg2:
+            if which1 == "start":
+                joined_segments += linear_segments[seg1]
+            else:
+                joined_segments += linear_segments[seg1][::-1]
+        if node_degrees[next_endpoint] == 1:
+            break
+        last_endpoint = endpoint
+        endpoint = next_endpoint
```
BainanXia: Not sure whether we can use [dfs_preorder_nodes](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.traversal.depth_first_search.dfs_preorder_nodes.html#networkx.algorithms.traversal.depth_first_search.dfs_preorder_nodes) (should be equivalent with `dfs_postorder_nodes` given it is a path graph) here

prereise/gather/griddata/hifld/data_access/load.py
```diff
@@ -216,8 +315,14 @@ def get_hifld_electric_power_transmission_lines(path):
     )
 
     properties["COORDINATES"] = [
-        line["geometry"]["coordinates"][0] for line in data["features"]
+        line["geometry"]["coordinates"] for line in data["features"]
     ]
+
+    # Join multiple distinct segments as necessary to get one single list
+    properties["COORDINATES"] = properties["COORDINATES"].map(
+        lambda x: _join_line_segments(x) if not isinstance(x[0][0], float) else x
```
BainanXia, line 315: So it is possible we have single segment in following forms `[[start, end]]` and `[start, end]`?

prereise/gather/griddata/hifld/data_access/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from prereise.gather.griddata.hifld.orchestration import create_csvs  # noqa: F401
```
BainanXia: Why we need this import here?

danielolsen: Oops. I was lazy and just did a copy/paste from an existing `__init__.py` file assuming it was empty. Fixed.

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: Yes, some versions of the hifld dataset have it one way and other have it the other.

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: TIL! That should definitely simplify things.

prereise/gather/griddata/hifld/data_access/load.py
BainanXia: If there is a Y-shaped end of the main branch, i.e. the central node has `betweenness_centrality`== 3, we can drop the node with shorter edge (following the current convention). If it is in the middle, i.e. the central node has `betweenness_centrality` > 3, we simply drop all of them which shouldn't be a `tie` by definition. But if there is a Y-shaped structure in the middle, i.e. more than one main branches, this approach won't work since in the original process, we will keep removing edges until one main branch is completely removed.

Not sure this will cover all the corner cases though. If we are diving deep into a rabbit hole, let's stop here and keep whatever works as always:)

danielolsen: I think you're right, in a three-segment non-linear connected tree, there is symmetry for the two 'longer' branches of the Y but there will be one that's shorter, just looking at graph topology.

I did a visual inspection on the latest dataset, and everything looks as reasonable as can be expected given an automated process and messy data. The logic for dropping the shortest one comes from this case (original on the left, joined on the right): 
image

It's hard to tell from visual inspection what the minimum spanning tree would look like, since all the ends are in roughly in the same location, but the short spur seems like the least important part of the line.

The worst output is actually a two-segment line, but one where once of the segments contains a doubled-back section, so the 'endpoint' is essentially in the middle:
image

danielolsen: This one is my other favorite. What's going on with that big loop that seems to start and end at the same point?
image

prereise/gather/griddata/hifld/data_access/load.py
danielolsen: Done. I ended up using [df_edges](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.traversal.depth_first_search.dfs_edges.html) instead, which I think should be equivalent, and verified that the behavior is the same as what I had previously.

prereise/gather/griddata/hifld/data_access/load.py
rouille: Weird.

BainanXia (approved):
Great. We are taking better use of the data! Thanks for adding those tests.

PR #286, opened 2022-04-05
refactor: update values and logic for representative transformer designs

danielolsen:

### Purpose
Update the logic used to generate transformers and assign parameter values for the HIFLD grid. These parameters are generated by examining real data sets to identify representative values for different voltage combinations (thank you @danlivengood and @YifanLi86 for helping to identify these values).

### What the code is doing
- We add two new data files to contain values which are defined based on voltage pairs (`transformer_params.csv`) and high-side voltages (`transformer_impedance_ratios.csv`).
- In `load.py`, we add a helper function to load these files and return a joined dataframe.
- In `transmission.py`:
  - We refactor the `create_transformers` function so that it no longer returns just bus pairs, but also adds parameter values. The other big change is that we also take the `lines` dataframe, which already has capacities assigned, and uses this information to determine the highest-capacity line connected to each transformer. This capacity is used to estimate how many parallel transformers to add to be able to meet this capacity. This is a deviation from the previous logic, which generated a single transformer, keeping the impedance constant based on the previous hard-coded values, and modifying only the transformer capacity to meet the lines capacity, using an assumption about the thermal capacity of connected lines based on their voltage alone.
  - A lower-level helper function `estimate_transformers` is used to support the refactored `create_transformers` function. `estimate_transformers` returns a single Series for each voltage pair describing the parameters of each transformer and the number required, while `create_transformers` used this information to create N parallel transformers, each matching the given parameters.
  - Based on the new behavior of `create_transformers`, we make some changes to the existing `add_impedance_and_rating` function (renaming it to `add_lines_impedances_ratings` now that it only deals with transmission lines, not transformers) and to `build_transmission`, so that we have the right parameters in the order that we need them.
- We remove the now-obsolete functions `estimate_branch_impedance` and `estimate_branch_rating`.
- In `test_transmission.py` we update the values expected from the refactored functions, and remove the checks on the now-obsolete functions.

### Testing
Unit testing.

### Usage Example/Visuals
Usage of `build_transmission` is unchanged, only internal logic is refactored.

### Time estimate
30-60 minutes. ~This is a draft PR for now, since it relies on #276, but all logic should be complete and ready for review.~ **EDIT**: #276 is merged, so this is ready to go.


danielolsen:
#276 is merged, so this is now ready for a full review.

rouille:
Do we need to create entries in the ***Attribution*** file for the two CSV files?

danielolsen:
> Do we need to create entries in the _**Attribution**_ file for the two CSV files?

They are self-created values, made by examining a few different data sources, so I don't think so. @BainanXia had the same question about transmission lines: https://github.com/Breakthrough-Energy/PreREISE/pull/276#discussion_r843294526.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -538,12 +539,13 @@ def create_buses(lines):
...
 
-def create_transformers(bus):
+def create_transformers(bus, lines, transformer_designs):
     """Add transformers between buses within the same substation. The assumed topology
     is that the highest-voltage bus in each substation is connected via a tansformer to
     every other voltage.
 
-    :param pandas.DataFrame bus: columns 'sub_id' and 'baseKV'
+    :param pandas.DataFrame bus: columns 'sub_id' and 'baseKV'.
+    :param pandas.DataFrame lines: columns 'from_bus_id', 'to_bus_id', and 'rateA'.
```
rouille, line 551: entry for `transformer_designs` is missing

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -538,12 +539,13 @@ def create_buses(lines):
...
     :return: (*pandas.DataFrame*) -- each row is one transformer, columns are
         ["from_bus_id", "to_bus_id"].
```
rouille: I guess the data frame is more complex now

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Done.

prereise/gather/griddata/hifld/data_process/transmission.py

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -628,55 +641,51 @@ def build_tower(series):
...
     )
+    branch["rateA"] = branch_plus_line_designs["line_object"].map(
+        lambda x: x.power_rating
+    )
 
 
-def estimate_branch_impedance(branch, bus_voltages):
-    """Estimate branch impedance using transformer voltages or line voltage and length.
+def estimate_transformers(bus_pair, lines, bus_voltages, transformer_designs):
+    """Determine the design of a transformer between two bus pairs, and how many of them
```
jenhagg: Should it be "between two buses" (or "between a pair of buses")?

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -538,32 +539,44 @@ def create_buses(lines):
...
+    repeated_transformers = single_transformers.drop(columns="number").reindex(
+        single_transformers.index.repeat(single_transformers["number"])
+    )
 
-    return pd.DataFrame(bus_pairs, columns=["from_bus_id", "to_bus_id"])
+    return repeated_transformers.reset_index(drop=True)
 
 
-def add_impedance_and_rating(branch, bus_voltages, line_overrides=None):
+def add_lines_impedances_ratings(branch, line_overrides={}):
```
jenhagg: Why switch default to `{}`?

prereise/gather/griddata/hifld/data_process/transmission.py

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Oops, I think that got messed up during a rebase. Fixed.

danielolsen:
Thinking a bit more about the logic for determining how many transformers to add, I wonder if our current logic is adding too many. If we have a situation where there are a couple of higher-voltage lines and one lower-voltage line connected to a substation, adding enough transformer capacity to support the highest capacity line perhaps doesn't make sense, if this transformer capacity ends up being much higher than the lower-voltage line. This could also effectively reduce the impedance between the high-voltage side and the low-voltage side, which could lead to unrealistic low-voltage line congestion/violations.

If we trust the transmission line capacities that are being added, perhaps we only need enough transformer capacity to support the smaller of either: the highest-capacity high-voltage line, or the sum of line capacities on the lower-voltage side. Just speculating here, but I think it would be rare to end up with a situation where the total line capacity on the low-voltage side is so large that it requires transformers capable of handling more power than any one high-voltage line.

E.g. one of the current worst congested lines in the HIFLD Western grid is a 66 kV line connected to a substation with 4x 230 kV lines and one other 66 kV line. The 2x 66 kV lines could be fully served by 1x 69/230 kV transformer (each 66 kV line has a capacity of about 82 MW, and each representative transformer is rated for 190 MW), but based on the current logic we add enough transformers to handle the full capacity of one of the 230 kV lines (552 MW), which equals 3x transformers and 1/3 of the effective impedance between the 230 and 66 kV parts of the network, which is liable to create more local congestion issues.

**EDIT**: this approach seems to only sort of work. There won't be any transformers which are congested _due to hitting their power limits before the connected lines_, but this doesn't solve the problem of meeting demand which exists at the low-voltage bus itself. We don't know the demand at this stage, because the demand doesn't get added until after the entire transmission network is built, and the plant data is built, such that we don't add demand at plant substations.

This could be a use case for the code in #279: we can go through our current grid-building process (i.e. within `create_grid`: `build_transmission()`; `build_plant()`; `assign_demand_to_buses();`), and then run the code in #279 to identify any bottlenecks, and if they're transformer-caused, then we add the requisite number of additional transformers in parallel to be able to serve the demand. The reduction in the number of transformers added with this approach is pretty dramatic though: 10,237 fewer transformers (from a starting number of 28,977, or a reduction of 35%). We don't yet have the real bus-level demands to be able to identify real bottlenecks (see #293), but the `Pd` values assigned by `assign_demand_to_buses` will hopefully be close enough to make this approach generally work. If the existing `const.demand_per_person` value is an under-estimation, we can increase it (since the absolute values of `Pd` aren't ever used anywhere, just the ratios within a zone).

danielolsen:
The last four commits implement the main idea discussed in https://github.com/Breakthrough-Energy/PreREISE/pull/286#issuecomment-1113472044. Briefly: instead of just looking at the capacity to support the highest-capacity-line (probably on the high-voltage side), we also look at the required capacity by _all_ of the lines on the low-voltage side, and we add enough transformer capacity to meet the smaller of those two. There are a few select locations where this results in inadequate capacity where there's demand at the low-voltage bus, but that's difficult or impossible to solve a priori at grid-build time (since not all of these bottlenecks are on the radial part of the grid that #279 analyzes), so we add one more data file of assumptions for which substations need additional transformer capacity. These assumptions are identified pretty quickly in ERCOT with only an iteration or two of power flow solves, compared to the line upgrades which currently require many iterations and are therefore the main effort required to getting to a feasible grid (although maybe having more reasonable impedances between the higher- and lower-voltage parts of the grid will help mitigate this).

The numbers:
- Transformers to support the highest-capacity line power: 28,977
- Transformers once we limit to the sum of capacities on the low-voltage side: 18,740
- Additional transformers added via explicit assumptions: 8
- Other adjustments required to regain ERCOT feasibility using this new transformer logic:
  - 1 additional transmission line
  - 1 more line voltage assumption
  - 5 more line design assumptions
  - 2 more transformers automatically created due to new lines/voltage assumptions.

BainanXia:
Probably I need to refresh my brain on the current logic of adding transformers: taking the example above, one substation is connected by 6 lines, 2x 66kV (82MW each) and 4x 230kV (552MW each). In this case, how many buses are there in this substation? Should it be 2 given there are 2 different voltage levels among the lines connected to it? If so, there should be only one transformer added, i.e. 66kV/230kV (forgot about the capacity calculation, did we choose the `max(82, 552)` before?)

Another question not related to transformers, how do we figure out that new transmission line needs to be added during feasibility tweaking? In the `adjustments required to regain ERCOT feasibility`, I understand the last 3 bullets but a little surprise on the first one.

danielolsen:
> Probably I need to refresh my brain on the current logic of adding transformers: taking the example above, one substation is connected by 6 lines, 2x 66kV (82MW each) and 4x 230kV (552MW each). In this case, how many buses are there in this substation? Should it be 2 given there are 2 different voltage levels among the lines connected to it? If so, there should be only one transformer added, i.e. 66kV/230kV (forgot about the capacity calculation, did we choose the `max(82, 552)` before?)

You are correct that there are 2 buses since there are 2 voltage levels.

- Before this PR: every transformer's power rating is a multiple of 800 MW where the multiple is determined based on the thermal rating of the default tower/conductor design for the higher voltage.
- At the start of this PR: we introduce representative transformer designs (x, r, rateA) for each voltage pair. We first determine the real power rating for each line using line geometries, then we pick the highest-capacity line connected to either end of the bus pair and add enough of the representative transformers to meet that level of power transfer.
- Now: to determine the maximum amount of power that we need to support between bus pairs, instead of just looking at the highest-voltage capacity line, we look at the smaller of a) the highest capacity line or b) the sum of powers of all lower-voltage lines.

> Another question not related to transformers, how do we figure out that new transmission line needs to be added during feasibility tweaking? In the adjustments required to regain ERCOT feasibility, I understand the last 3 bullets but a little surprise on the first one.

This line wasn't strictly necessary, but when looking at real maps I noticed that there should definitely be a line here, but wasn't. So it was manually noticing a data gap, rather than inferring a data gap or data mistake based on power flow results.

BainanXia:
@danielolsen Very clear. Thanks for the detailed explanation. Now I remember everything:)

danielolsen:
> @danielolsen Very clear. Thanks for the detailed explanation. Now I remember everything:)

I had to go back and look at the before/after code to remind myself. I also realized that we hadn't deleted the old parameters which are now unused, so I just did that too.

prereise/gather/griddata/hifld/data_process/tests/test_transmission.py
```diff
@@ -185,71 +180,49 @@ def test_create_transformers():
...
-    transformers = pd.DataFrame(
-        {"from_bus_id": [0, 1, 2], "to_bus_id": [1, 2, 3], "type": ["Transformer"] * 3}
+    lines = pd.DataFrame(
+        [
+            # One branch to low-voltage side of buses (0, 1) transformer
+            {"from_bus_id": 100, "to_bus_id": 0, "rateA": 75},
+            # Two branches to high-voltage side of buses (0, 1) transformer
+            {"from_bus_id": 1, "to_bus_id": 101, "rateA": 100},
+            {"from_bus_id": 1, "to_bus_id": 102, "rateA": 300},
+            # Three branches to low-voltage side of buses (2, 3) transformer
```
BainanXia: Three branches -> Two branches?

prereise/gather/griddata/hifld/data_process/tests/test_transmission.py
```diff
@@ -185,71 +180,49 @@ def test_create_transformers():
...
+            {"from_bus_id": 103, "to_bus_id": 2, "rateA": 50},
+            {"from_bus_id": 104, "to_bus_id": 2, "rateA": 100},
+            # One branch to high-voltage side of buses (2, 3) transformer
+            {"from_bus_id": 3, "to_bus_id": 105, "rateA": 300},
+            # One branch to each side of buses (5, 6) transformer
+            # The second of these is also on the high-voltage side of buses (4, 5) xfmr
```
BainanXia, line 204: We don't have any lines on the low-voltage side of buses (4, 5), right?

prereise/gather/griddata/hifld/data_process/tests/test_transmission.py
```diff
@@ -185,71 +180,49 @@ def test_create_transformers():
         },
```
BainanXia, line 180: Not in this PR, it seems bus with index 7 is not used in this test case?

prereise/gather/griddata/hifld/data_process/tests/test_transmission.py
danielolsen: Correct. This is a situation that I don't think we'd ever encounter in our current HIFLD grid-building, since we only add buses where there's a line of that voltage connected to the substation, but the code logic is resilient to this edge case since it uses `int(required_power / single_transformer_power) + 1` so we get `int(0 / x) + 1 = 1` transformer between these two buses.

prereise/gather/griddata/hifld/data_process/tests/test_transmission.py
danielolsen: Correct. It's there to make sure that the code can handle substations with only one bus in them.

prereise/gather/griddata/hifld/data_process/tests/test_transmission.py
danielolsen: Good catch, fixed.

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -538,32 +540,62 @@ def create_buses(lines):
...
+            "sub_id": sub,
+            "from_bus_id": b,
+            "to_bus_id": volt_series.sort_values().index[i + 1],
+            "from_kV": bus.loc[b, "baseKV"],
+            "to_kV": bus.loc[volt_series.sort_values().index[i + 1], "baseKV"],
+        }
         for sub, volt_series in bus.groupby("sub_id")["baseKV"]
         for i, b in enumerate(volt_series.sort_values().index[:-1])
         if len(volt_series) > 1
     ]
```
BainanXia: Maybe we could do following instead:
```python
    bus_pair_data = [
        {
            "sub_id": sub,
            "from_bus_id": from_id,
            "to_bus_id": to_id,
            "from_kV": bus.loc[from_id, "baseKV"],
            "to_kV": bus.loc[to_id, "baseKV"],
        }
        for sub, v_series in bus.sort_values("baseKV").groupby("sub_id")["baseKV"]
        for from_id, to_id in zip(v_series.index, v_series.index[1:])
    ]
```

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -628,55 +660,59 @@ def build_tower(series):
...
-    :return: (*float*) -- impedance for that branch (per-unit).
+    :param iterable bus_pair: a pair of bus IDs, corresponding to indices in
+        ``bus_voltages`` and entries in the 'from_bus_id' and 'to_bus_id' columns of
+        ``lines``.
+    :param pandas.DataFrame lines: data frame of transmission lines containing columns:
+        'from_bus_id', 'to_bus_id', and 'rateA'.
+    :param pandas.Series bus_voltages: index is bus IDs, values are voltages (kV).
+    :param pandas.DataFrame transformer_designs: index is a tuple of voltage pairs in
+        (low_kV, high_kV) order, columns are 'x', 'r', and 'MVA'.
+    :return: (*pandas.Series*) -- indices are: 'x', 'r', 'MVA', and 'number'.
```
BainanXia, line 689: values are 'number'?

BainanXia (approved):
Thanks for the test which helps a lot to verify the logic.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: It's a Series with a single (not meaningfully ordered) index `['x', 'r', 'MVA', 'number']`, and the values are the value of each parameter. Essentially a dictionary. We call this from an `apply` so that the Series get grouped together and eventually form a dataframe where these are the columns.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Nice! Done (with a couple of tweaks).

prereise/gather/griddata/hifld/data_process/transmission.py
```diff
@@ -538,32 +540,61 @@ def create_buses(lines):
...
+            "sub_id": sub,
+            "from_bus_id": from_id,
+            "to_bus_id": to_id,
+            "from_kV": volt_series.loc[from_id],
+            "to_kV": volt_series.loc[to_id],
+        }
         for sub, volt_series in bus.groupby("sub_id")["baseKV"]
-        for i, b in enumerate(volt_series.sort_values().index[:-1])
-        if len(volt_series) > 1
+        for from_id, to_id in zip(volt_series.index[:-1], volt_series.index[1:])
```
BainanXia, line 571: So we no longer need to sort by `baseKV` first? The default behavior of `zip` will stop until it reaches the end of the shorter list, then why we would like to exclude the last element explicitly?

prereise/gather/griddata/hifld/data_process/transmission.py
BainanXia: Got it. It's actually a series formed by a single row of a data frame.

prereise/gather/griddata/hifld/data_process/transmission.py
danielolsen: Aha, I missed that your refactor included the sorting before the groupby. The previous version works already because within `create_buses` we create buses within substations in sorted voltage order, so the buses within each substation are already sorted when they get here, but I think it's clearer to maintain the (redundant) sorting for the sake of being more flexible with the inputs.

I agree that we don't need to explicitly shorten both, but in my mind it's a little clearer what's going on here if we do so, without having to remember how `zip` functions for iterables of mixed length.

PR #287, opened 2022-04-08
feat: generate profiles for smart charging

isond:

### Purpose
To add the Python translation of the smart charging code

### What the code is doing
How is the purpose executed?
This calculates the smart charging profiles for either LDVs or LDTs by iterating through the NHTS data. For each vehicle, all of its trips will help determine the optimized charging by considering various factors such as dwelling activity and charging costs.

### Testing
This has not been tested yet but once it's been optimized to run faster, we can compare the output against the matlab output.

### Where to look
The new code is located here: "PreREISE/prereise/gather/demanddata/transportation_electrification/".

There are two new files. `smart_charging.py` is the main file with the algorithm. `dwelling.py` contains helper functions to calculate the charging power during a vehicle's dwelling period.

### Usage Example/Visuals

### Time estimate
It could about a few days for the reviewers to understand this code change.


prereise/gather/demanddata/transportation_electrification/dwelling.py
```diff
@@ -0,0 +1,98 @@
...
+    """Determines the energy limit
+
+    :param int power: charger power, EVSE kW.
+    :param segment: the amount of the rates segments the dwelling activity possess.
+    :param float dwelling_start_time: dwelling start time.
+    :param float dwelling_length: dwelling end time.
+    :param float total_dwell_period: the total dwelling period.
+    :return: (*list*) -- list of energy limits during the time span of available charging
+    """
+    t1 = dwelling_start_time + 1 - dwelling_start_time
```
danielolsen: This seems equivalent to `t1 = 1`. Is there something else we're trying to do here?

danielolsen:
There are a couple of functions within `dwelling.py` that look similar to functions that we re-factored out of the immediate charging code in #266 (`get_location`, `consider_trip_number`, and `get_charging_power`). It seems like we could move the code from lines [156-171](https://github.com/isond/PreREISE/blob/849b3d5c1423eb94a053b3c4fd342def032e9569/prereise/gather/demanddata/transportation_electrification/immediate.py#L156) into its own function which take `trips` as input and returns a boolean column which represents if charging is allowed or not, and then use this function both within the existing `immediate_charging` and the new `smart_charging` functions. What do you think?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+
+    TRANS_charge = np.zeros(24 * len(input_day))
+    data_day = data_helper.get_data_day(newdata)
+
+    daily_vmt_total = data_helper.total_daily_vmt(
+        census_region, comm_type, location_strategy, input_day
+    )
+
+    kwh = kwhmi * veh_range
+    emfacvmt = 758118400
```
danielolsen, line 303: What does this constant represent?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+            ]
+
+        cost = LOAD_adj
+
+        G2Vload = np.zeros((100, 48))
+        individualG2Vload = np.zeros(nd_len, 48)
+
+        i = 0
+
+        while i < nd_len:
```
danielolsen, line 328: This nested loop is probably the source of the slow performance, since we end up with (number_of_trips * number_of_days_in_year) iterations, and we perform an optimization solve within each one. Are we performing any repeated calculations within here? Can any of the calculations be batched to work on a group of trips, rather than individual ones?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                                k = k - m
+                                # ones part of the column
+                                a[i - 1 : m, segcum[k] - seg[k] + 1 : segcum[k]] = 1
+
+                        A_coeff[(m - 1) * j : (m - 1) * j, :] = -a
+
+                    # try to use linopy instead
+                    linprog_result = scipy.optimize.linprog(
+                        f, A_coeff, B_coeff, Aeq, Beq, zip(lb, ub)
+                    )
```
danielolsen: `scipy.optimize.linprog` seems like a fine choice here if we are alright with the performance of the open-source `HiGHS` solver, which seems to be the most efficient backend for this function according to the latest `scipy` documentation: https://docs.scipy.org/doc/scipy/reference/optimize.linprog-highs.html#optimize-linprog-highs.

`linopy` and `Pyomo` have the benefit of being able to interface with proprietary solvers, which may or may not be necessary depending on the performance of HiGHS.

prereise/gather/demanddata/transportation_electrification/dwelling.py
```diff
@@ -0,0 +1,98 @@
...
+
+    elif segment == 2:
+        energy_limit = [power * t1, power * tn]
+
+    else:
+        energy_limit = [power * t1]
+        energy_limit += [power for s in range(segment - 1)]
+        energy_limit += [power * tn]
+
+    return energy_limit
```
danielolsen, line 50: If every output of this function is some value times `power`, do we even need the `power` input here, or are we essentially just calculating some amount of time?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
+import data_helper
+import const
+import dwelling
```
danielolsen: This will need to be modified to refer to `prereise.gather.demanddata.transportation_electrification.*`

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                    charging_consumption = (
+                        []
+                    )  # zero the consumption, this is a vector, the consumption for each trip
+                    rates = (
+                        []
+                    )  # zero the rates of all segments, which will be used as cost function later
+                    elimit = (
+                        []
+                    )  # zero the charging energy limits of all segments, which will be used as the upper bound
+                    seg = []
```
danielolsen: If comments are above the functional line, rather than inline, `black` won't mangle the code so badly. E.g.
```python
# zero the consumption, this is a vector, the consumption for each trip
charging_consumption = []
...
```
rather than starting with:
```python
charging_consumption = [] # zero the consumption, this is a vector, the consumption for each trip
```

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                        ]
+                        dwelling_length = individual.iloc[
+                            trip_num,
+                            individual.columns.get_loc("Dwell time (hour decimal)"),
+                        ]
+                        total_dwell_period = dwelling_start_time + dwelling_length
+
+                        segment = dwelling.get_segment(
+                            dwelling_start_time, total_dwell_period
+                        )
```
danielolsen: It seems that the `dwelling.get_segment` function just calculates `total_dwell_period - dwelling_start_time + 1`. If                         `total_dwell_period = dwelling_start_time + dwelling_length`, then are we essentially just calculating `dwelling_length + 1` here?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                        energylimit = dwelling.get_energy_limit(
+                            power,
+                            segment,
+                            dwelling_start_time,
+                            dwelling_length,
+                            total_dwell_period,
+                        )
+
+                        # get the rates and the extra cost related to the infrasctructures; used as cost function
```
danielolsen: `infrasctructures` -> `infrastructures`

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                        for _ in range(dwelling_start_time, dwelling_length + 1):
+                            rates.append(cost[_])
+
+                        # get the energy limit; used as constraint
+                        for s in range(segment):
+                            elimit.append(energylimit)
+
+                        seg.append(
+                            segment
+                        )  # record the amount of the segments in the ith dwelling activity
```
danielolsen: Same comment as above re: inline comments vs. comments on their own lines.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                            power,
+                            segment,
+                            dwelling_start_time,
+                            dwelling_length,
+                            total_dwell_period,
+                        )
+
+                        # get the rates and the extra cost related to the infrasctructures; used as cost function
+                        for _ in range(dwelling_start_time, dwelling_length + 1):
+                            rates.append(cost[_])
```
danielolsen: Is the equivalent to: `rates += cost[dwelling_start_time : dwelling_length + 1]`?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+
+                        # get the energy limit; used as constraint
+                        for s in range(segment):
+                            elimit.append(energylimit)
```
danielolsen: Is this equivalent to `elimit += [energylimit] * s`?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+                        A_coeff[(m - 1) * j : (m - 1) * j, :] = -a
+
+                    # try to use linopy instead
+                    linprog_result = scipy.optimize.linprog(
+                        f, A_coeff, B_coeff, Aeq, Beq, zip(lb, ub)
+                    )
+                    # fval is the value of the final cost, exitflag is the reason why the optimization terminates
+                    # 0-success, 1-limit reached, 2-problem infeasible, 3-problem unbounded, 4-numerical difficulties
+                    x = np.array(linprog_result.x)
+                    fval = linprog_result.fun
```
danielolsen: This value never appears to be used. If its unnecessary, can we also remove the `f = rates / const.charging_efficiency` step and optimize against `rates` directly, since multiplying the objective function by a constant shouldn't change the optimal value of the decision variables?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+            ]
+
+            if data_day[i] == input_day[day_iter]:
+
+                # only home based trips
+                if (
+                    newdata.iloc[i, newdata.columns.get_loc("why from")]
+                    * newdata.iloc[i + total_trips, newdata.columns.get_loc("why to")]
+                    == 1
+                ):
```
danielolsen, line 344: There's a lot that's going on inside this block. It seems that the major functions are:
* preparing input data for the optimization
* running the optimization
* extracting the optimization results.
* others?

If we can move some/all of these into their own functions, it should make it easier to understand everything that's going on within each function and within the overall block, and make testing each component easier.

prereise/gather/demanddata/transportation_electrification/data_helper.py
```diff
@@ -89,3 +90,56 @@ def update_if_weekend(data: pd.DataFrame):
...
+
+    daily_vmt_total = np.zeros(len(input_day), 2)
+
+    for day_iter in range(len(input_day)):
+        for i in range(n):
+            if data_day[i] == input_day[day_iter]:
+                daily_vmt_total[day_iter][0] += data.iloc[
+                    i, data.columns.get_loc("Miles traveled")
+                ]
+                daily_vmt_total[day_iter][1] += 1
```
danielolsen: This nested loop seems unnecessary. Can we do this as a two step, where the first step sums by weekday/weekend and then the second step applies the weekday/weekend sums to the input days?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
danielolsen: If `dwelling_start_time`, `dwelling_length`, or `segment` are floats, we may end up with errors here since `range` expects integers.

prereise/gather/demanddata/transportation_electrification/data_helper.py
```diff
@@ -89,3 +90,56 @@ def update_if_weekend(data: pd.DataFrame):
...
+
+
+def get_total_daily_vmt(
+    data: pd.DataFrame,
+    comm_type: int,
+    locationstrategy: List[int],
+    input_day: List[int],
+    data_day: np.array,
+):
+    """Load data and use the parameters to calculate total_daily_vmt.
```
rouille: What do we calculate exactly? Can we be more explicit here.

prereise/gather/demanddata/transportation_electrification/data_helper.py
```diff
@@ -89,3 +90,56 @@ def update_if_weekend(data: pd.DataFrame):
...
+
+    :param pandas.DataFrame data: the data returned from :func:`load_data`.
+    :param int comm_type: the type of Commute
+    :param List[int] locationstrategy: strategy for each location
+    :param List[int] input_day: day of the week for each day in the year derived from
+        first_func
```
rouille: Can we use `list` for the type in the docstring to follow our standard. Also what is `first_func`?

prereise/gather/demanddata/transportation_electrification/dwelling.py
```diff
@@ -0,0 +1,98 @@
+def get_segment(dwelling_start_time, total_dwell_period):
+    """Get dwelling activity segment.
+
+    :param float s1: dwelling start time.
+    :param float sn: dwelling end time.
```
rouille: `s1` and `sn` are not parameters

prereise/gather/demanddata/transportation_electrification/dwelling.py
```diff
@@ -0,0 +1,98 @@
...
+        after last trip (1-anytrip number, 2-last trip)
+    :param int total_trips: total trips that the vehicle makes
+    :param int trip_num: the trip number of the current trip
+    :return: (*bool*) -- boolean that represents if the vehicle should charge
+    """
+    if trip_strategy == 1:
+        return True
+
+    elif trip_strategy == 2:
+        return total_trips == trip_num
```
rouille: If `trip_strategy` is neither 1 or 2 the function will return `None`. Is that ok?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+    trip_strategy=1,
+):
+    """Smart charging function
+
+    :param int census_region: any of the 9 census regions defined by US census bureau.
+    :param int model_year: year that is being modelled/projected to, 2017, 2030, 2040, 2050.
+    :param int veh_range: 100, 200, or 300, represents how far vehicle can travel on single charge.
+    :param int kwhmi: fuel efficiency, should vary based on vehicle type and model_year.
+    :param int power: charger power, EVSE kW.
+    :param int location_strategy: where the vehicle can charge-1, 2, 3, 4, or 5; 1-home only, 2-home and work related, 3-anywhere if possibile, 4-home and school only, 5-home and work and school.
```
rouille: this line is too long, should be <89 characters

prereise/gather/demanddata/transportation_electrification/data_helper.py
```diff
@@ -89,3 +90,56 @@ def update_if_weekend(data: pd.DataFrame):
     data.loc[data["Day of Week"].isin(range(2, 7)), "If Weekend"] = 2
 
     return data
+
+
+def get_total_daily_vmt(
+    data: pd.DataFrame,
+    comm_type: int,
+    locationstrategy: List[int],
```
rouille: `location_strategy` is used in the `smart_charging` function. Why not using the same name for the variable

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -0,0 +1,365 @@
...
+    :param str veh_type: determine which category (LDV or LDT) to produce charging profiles for
+    :param str filepath: the path to the nhts mat file.
+    :param int comm_type: the type of Commute
```
rouille: It would be useful to list here the allowed values for `comm_type`

rouille:
It would be helpful to assess the validity of the input parameters, e.g., `LDT` vs `ldt`, etc. Raising some `ValueError` would be helpful to guide the user.

prereise/gather/demanddata/transportation_electrification/data_helper.py
```diff
@@ -89,3 +90,56 @@ def update_if_weekend(data: pd.DataFrame):
...
+    :param int comm_type: the type of Commute
+    :param List[int] locationstrategy: strategy for each location
+    :param List[int] input_day: day of the week for each day in the year derived from
+        first_func
+    :param np.array data_day: indicates weekend or weekday for every day.
+    :return: (*np.array*) -- an array where each element is a year of entries for each vehicle
+        type
+    """
+    # get the data
+    n = len(data)
```
rouille: Do we use `n`?

prereise/gather/demanddata/transportation_electrification/data_helper.py
```diff
@@ -89,3 +90,56 @@ def update_if_weekend(data: pd.DataFrame):
...
+    if comm_type == 1:
+        # removes VMT for trips from work->home and home-> work
+        # (not exactly correct due to chained trips involving work and home but no
+        # way to remove the data)
+        if all([i != 2 for i in locationstrategy]):
+            locationstrategy = 2
+            # putting this part in function means rest of code won't be able to see
+            # change in locationstrategy value.
+            # will have to use global or return locationstrategy and reassign
+            print('"locationstrategy" changed to "Home and Work" for comm_type == 1')
```
rouille: Do we use `locationstrategy`?

prereise/gather/demanddata/transportation_electrification/dwelling.py
```diff
@@ -0,0 +1,98 @@
...
+    return power
+
+
+def get_energy_limit(
+    power, segment, dwelling_start_time, dwelling_length, total_dwell_period
+):
+    """Determines the energy limit
+
+    :param int power: charger power, EVSE kW.
+    :param segment: the amount of the rates segments the dwelling activity possess.
```
rouille: missing type of parameter `segment`

prereise/gather/demanddata/transportation_electrification/dwelling.py
```diff
@@ -0,0 +1,60 @@
...
+def get_energy_limit(
+    power, segment, dwelling_start_time, dwelling_length, charging_efficiency
+):
+    """Determines the energy limit for the entire dwelling period. Takes into
+    consideration if charging does not start or end directly on the hour.
+
+    :param int power: charger power, EVSE kW.
+    :param int segment: the amount of the rates segments the dwelling activity possess.
+    :param float dwelling_start_time: dwelling start time.
+    :param float dwelling_length: dwelling end time.
```
dmuldrew: Based on the context, seems like this should dwelling length?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: placeholder number...amount of vehicle mile traveled for the population of electric vehicles

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: will consider for later

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: was updated

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: dealt with

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: working on restructuring

prereise/gather/demanddata/transportation_electrification/data_helper.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: fixed

prereise/gather/demanddata/transportation_electrification/data_helper.py
dmuldrew: more specific now

prereise/gather/demanddata/transportation_electrification/data_helper.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/dwelling.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/dwelling.py
dmuldrew: updated

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: done

prereise/gather/demanddata/transportation_electrification/data_helper.py
isond: Removed from function since it's not needed anymore

prereise/gather/demanddata/transportation_electrification/smart_charging.py
isond: This also has been removed since it's not needed.

prereise/gather/demanddata/transportation_electrification/dwelling.py
isond: Yup! That was an old description, it's been updated now.

prereise/gather/demanddata/transportation_electrification/dwelling.py
dmuldrew: appears to be resolved

dmuldrew (approved):

PR #289, opened 2022-04-08
feat/data: update HIFLD data to April 2022 version, refactor line assignment

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
- Update the data we're pulling to use the latest available (April 2022 for both transmission lines and substations)
- Refactor the logic to account for the fact that many lines don't have substations at their endpoints
- Update our set of line/substation interconnect assumptions in light of the above.

### What the code is doing
- In `load.py`, we're a bit more accommodating of the inconsistent naming of the Shape Length columns that we want to drop.
- In `topology.py`, we make sure that we tag **all** of the separated/reconnected lines with their appropriate interconnects.
- In `transmission.py`:
  - We modify the behavior of the `map_lines_to_substations_using_coords` function so that if there are no substations 'close' to a line's endpoint (as defined by our new optional parameter), we create a new proxy substation for it. We do the KDTree/query process twice, where the first time just collects a list of transmission line endpoints with no substation, then we add proxy substations, and the second time we actually re-map all the lines.
  - We avoid filtering out substations which are tagged as having `LINES` = 0, since this information seems to be very noisy.

### Testing
Tested manually to generate static grid components. ~The resulting static grid components haven't been tested for network feasibility yet, so we may also need to update some of the line assumptions that were introduced in #276.~ **EDIT**: update: the results of this new code have been tested alongside the new code in #286 and new line/substation assumptions have been added which seem to ensure generator-side feasibility.

### Usage Example/Visuals
The call signature to `map_lines_to_substations_using_coords` remains the same, we just add one more optional parameter that controls how far away from any substation a line's endpoint can be before we create a new proxy substation for it (instead of connecting it to an existing one).

### Time estimate
30 minutes? The code changes are fairly light, most of the changes are in the constants which required a tedious guess-and-check process. ~This is a draft PR for now, since the functionality depends partially on #283, and since we may want to make some modifications to the underlying logic, but the code that's here creates the interconnections as we expect.~ **EDIT**: #283 is merged, so this is ready to go.

PR #292, opened 2022-04-22
feat: add lines to HIFLD dataset before further processing

danielolsen:

### Purpose
Add the ability to append the existing HIFLD lines dataset with additional manually-defined lines. This allows us to fill in parts of the grid where we know there are transmission lines, but these lines are missing for whatever reason. This should be at least partial fulfillment of #234; we get the capability to manually add lines, but we don't yet have a full set of required lines to connect urban areas with lots of undergrounding, or logic to estimate their connectivity.

### What the code is doing
- We add a new function `add_proxy_lines` which takes data from const.py and uses it to populate required information to add to the `branch` data frame. It adds one new column `"line_design_assumptions"` which will get used in the refactored version of `add_lines_impedances_ratings`.
- Within `add_lines_impedances_ratings`, we modify the logic so that we first identify the design for each transmission line (using either the default for that voltage or the defined overrides), then we check for a `"line_design_assumptions"` column and if this is present, override the design with the design tuple from this column, and then finally use the aggregated design mapping to create a `Line` object for each branch so that we can estimate rating and impedance. This way, the data in const.py can have non-default designs without needing to use the existing `{branch_id: design}` mapping schema in const.py, since we won't know the ID of these newly-added lines in advance (their IDs get generated based on the max ID of the other branches).
- Within `build_transmission`, we call `add_proxy_lines` before we connect the branches via the minimum spanning tree, so that manually-defined branches which connect islands will avoid the logic which would otherwise create new branches. This way, even if a branch would get added via the minimum spanning tree logic, it can be entirely replaced with a different branch if we want the branch to have different characteristics.

### Testing
Tested manually.

### Time estimate
30 minutes. ~This is a draft PR for now, since it relies on code in #286.~

### DIFF
--- a/prereise/gather/griddata/hifld/const.py
+++ b/prereise/gather/griddata/hifld/const.py
@@ -459,6 +459,23 @@ proxy_substations = [
     },
 ]
 
+proxy_lines = [
+    {
+        "SUB_1_ID": 301352,
+        "SUB_2_ID": 308812,
+        "VOLTAGE": 138,
+        "circuits": 2,
+        "conductors": 2,
+    },
+    {
+        "SUB_1_ID": 309418,
+        "SUB_2_ID": 308608,
+        "VOLTAGE": 138,
+        "circuits": 2,
+        "conductors": 1,
+    },
+]
+
 substations_lines_filter_override = {301995}
 
 seams_substations = {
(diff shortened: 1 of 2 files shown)


danielolsen:
#286 is merged, so this is ready for review.

danielolsen:
The commits for this PR are also now part of the branch for #289, so we could either merge this and then rebase #289, or do everything there. **EDIT**: #289 is merged, so closing this one.

PR #299, opened 2022-06-01
fix: negative dwell time calculation in data processing

isond:

### Purpose
To update the dwell time calculation in the original data processing code.

### What the code is doing
How is the purpose executed?
The dwell times are calculated by subtracting the current trip's end time from the next trip's start time. The last trip's dwell time is calculated by subtracting it's end time from the first trip's start time. This updated calculation also handles when start times begin past 12AM for all trips, which the previous version didn't handle and outputted negative times.

### Testing
Since the output should not contain negative values (which the original code produced), some of the trips with the negative values were manually recalculated. Those were then used as test cases to compare against the updated code output.

### Where to look
PreREISE/prereise/gather/demanddata/transportation_electrification/data_process.py

data_process.py

### Usage Example/Visuals

### Time estimate
It should take maybe 15 minutes to 30 minutes to understand.

### DIFF
--- a/prereise/gather/demanddata/transportation_electrification/data_process.py
+++ b/prereise/gather/demanddata/transportation_electrification/data_process.py
@@ -4,7 +4,9 @@ import pandas as pd
 
 
 def calculate_dwell_time(data: pd.DataFrame):
-    """Calculates the dwell time, how long a vehicle has been charging
+    """Calculates the dwell time, how long a vehicle has been charging. For the last
+    trip, dwell time is calculated based on the last trip's end time and the start
+    time of the first trip. This calculation handles when start times begin past 12AM.
 
     :param pandas.DataFrame data: the data to calculate the dwell time from
     :return: (*pandas.Series*) -- list of dwell times
@@ -14,10 +16,10 @@ def calculate_dwell_time(data: pd.DataFrame):
         - data["End time (hour decimal)"].iloc[:-1]
     )
     dwells.loc[data.index[-1]] = (
-        24
+        data["Start time (hour decimal)"].iloc[0]
         - data["End time (hour decimal)"].iloc[-1]
-        + data["Start time (hour decimal)"].iloc[0]
     )
+    dwells[dwells < 0] += 24
 
     return dwells
 


prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -4,19 +4,34 @@
 
 
 def calculate_dwell_time(data: pd.DataFrame):
-    """Calculates the dwell time, how long a vehicle has been charging
+    """Calculates the dwell time, how long a vehicle has been charging. For the last
+    trip, dwell time is calculated based on the last trips's end time and the start
```
BainanXia: `trips's` -> `trip's`?

prereise/gather/demanddata/transportation_electrification/data_process.py
```diff
@@ -4,19 +4,34 @@
...
+
     dwells.loc[data.index[-1]] = (
-        24
+        data["Start time (hour decimal)"].iloc[0]
         - data["End time (hour decimal)"].iloc[-1]
+        if data["Start time (hour decimal)"].iloc[0]
+        > data["End time (hour decimal)"].iloc[-1]
+        else 24
         + data["Start time (hour decimal)"].iloc[0]
+        - data["End time (hour decimal)"].iloc[-1]
```
BainanXia: Maybe we can do following instead
```python
dwells =  (
    data["Start time (hour decimal)"].iloc[1:].values 
    - data["End time (hour decimal)"].iloc[:-1]
)
dwells.loc[data.index[-1]] = (
    data["Start time (hour decimal)"].iloc[0]
    - data["End time (hour decimal)"].iloc[-1]
 )
dwells[dwells < 0] += 24
```

BainanXia:
Not related to this PR: should we regenerate the existing immediate charing profiles in the Dropbox folder with this fix?

isond:
> Not related to this PR: should we regenerate the existing immediate charing profiles in the Dropbox folder with this fix?

I think it would be a good idea to regenerate the existing profiles that way they're up-to-date.

BainanXia:
@isond This is good to go and I will press the button once the linter doesn't complain. Thanks!

BainanXia (approved):
Let's first rebase onto the latest `Breakthrough-Energy:transportation_electrification`, then do a `squash and merge` Thanks 🚀

PR #302, opened 2022-08-26
feat: generate charging profiles for immediate charging HDV

isond:

### Purpose
To add the Python translation of the immediate charging HDV code

### What the code is doing
How is the purpose executed?
Similar to the immediate charging code for LDVs and LDTs, it iterates through the HDV vehicle data and adjusts the charging profiles by applying weight factors. However, this code applies those factors after each vehicle rather than applying it at the end to the final charging profiles.

### Testing
Output from this code was compared against the MATLAB output.

### Where to look
The new code can be found in "PreREISE/prereise/gather/demanddata/transportation_electrification/".

`immediate_charging_HDV.py` contains the main algorithm
`fdata_v10st.mat` which is a new dataset with HDV vehicles

### Usage Example/Visuals

### Time estimate
It can take a few days for the reviewers to go over the code change.


isond:
I just want to point out that `calculate_charging` was refactored because there was an issue with grouping and indexing. Currently, it groups by `Vehicle Number` instead of `trip number`. In the current data, some vehicles have single trips that occur multiple times, but with the previous version of the function, it would group them all together and renumber them as if they were all consecutive trips. 

Also, in the previous version, I think `previous_group["sample vehicle number"].isin(relevant_vehicles)` gets the unique values and so not all of the entries for the vehicles would get considered to update their start SOC. This would affect the indexing to put the new values back into the DataFrame.

I feel like there's a better way to go about the current function, but let me know what you think or if you need more clarification!

BainanXia (approved):
Given the outputs are as expected and the logic is clear, we decide to merge this into the feature branch in PreREISE first.

Issue #308, opened 2022-09-14
LDV: add charging_efficiency condition


dmuldrew:
# :rocket:

- [ x] Is your feature request essential for your project?

### Describe the workflow you want to enable

We would like have charging efficiency vary according to the power parameter input

### Describe your proposed implementation, if applicable

1. Add the following condition:
```
if power > 19.2, charging efficiency=0.95
else charging_efficiency = 0.9
```
2. Make `charging_efficiency` a parameter in relevant functions instead using the value from `const.py`
3. Remove charging efficiency and ER values from `const.py`

Example in existing code can be found here:
https://github.com/isond/PreREISE/blob/fc6ed851fc9fbdc0a65b309ab37aa51bbfe8a84c/prereise/gather/demanddata/transportation_electrification/smart_charging_HDV.py#L261

### Additional context
Add any other context or screenshots in this section, e.g., a plot from an article you
believe would clearly communicate results.

PR #314, opened 2022-10-04
refactor: maximum LDV dwell period to 72 hours, finish charging efficiency implementation 

dmuldrew:

### Purpose

Reconcile LDV and HDV implementations into one.

### What the code is doing

1. Change `why_to` to `dwell_location`
2. Change maximum LDV dwell period to 72 hours like the HDV code
3. Finish implementation of issue:
https://github.com/Breakthrough-Energy/PreREISE/issues/308

### Testing

Automated tests

### Time estimate
~20 min

### DIFF
--- a/prereise/gather/demanddata/transportation_electrification/immediate.py
+++ b/prereise/gather/demanddata/transportation_electrification/immediate.py
@@ -155,7 +155,7 @@ def immediate_charging(
         trips["location_allowed"] = True
     else:
         allowed = allowed_locations_by_strategy[location_strategy]
-        trips["location_allowed"] = trips["why_to"].map(lambda x: x in allowed)
+        trips["location_allowed"] = trips["dwell_location"].map(lambda x: x in allowed)
     # Add booleans for whether the trip_number (compared to total trips) allows charging
     if trip_strategy == 1:
         trips["trip_allowed"] = True
--- a/prereise/gather/demanddata/transportation_electrification/immediate.py
+++ b/prereise/gather/demanddata/transportation_electrification/immediate.py
@@ -11,13 +11,16 @@ allowed_locations_by_strategy = {
 }
 
 
-def calculate_charging(trips, charging_power, battery_capacity, kwhmi):
+def calculate_charging(
+    trips, charging_power, battery_capacity, kwhmi, charging_efficiency
+):
     """Parse travel patterns to estimate charging and state-of-charge after each trip.
 
     :param pandas.DataFrame trips: trip data.
     :param int/float charging_power: charging power (kW).
     :param int/float battery_capacity: battery capacity (kWh).
     :param int/float kwhmi: vehicle electricity consumption (kWh/ mile).
+    :param int/float charging_efficiency: from grid to battery efficiency.
     """
     # Add trip_number entries
     trips["trip_number"] = trips.groupby("vehicle_number").cumcount() + 1
@@ -46,12 +49,12 @@ def calculate_charging(trips, charging_power, battery_capacity, kwhmi):
         # Calculate charging duration/energy for the trips that can charge
         trips.loc[group.index, "full_charge_time"] = (
             battery_capacity - trips["trip end battery charge"]
-        ) / (charging_power * const.charging_efficiency)
+        ) / (charging_power * charging_efficiency)
         trips.loc[group.index, "charging time"] = trips["charging_allowed"] * (
             trips[["full_charge_time", "dwell_time"]].apply(min, axis=1)
         )
         trips.loc[group.index, "charging consumption"] = (
-            trips["charging time"] * charging_power * const.charging_efficiency
+            trips["charging time"] * charging_power * charging_efficiency
         )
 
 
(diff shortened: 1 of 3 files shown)


prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -155,7 +163,7 @@ def immediate_charging(
         trips["location_allowed"] = True
     else:
         allowed = allowed_locations_by_strategy[location_strategy]
-        trips["location_allowed"] = trips["why_to"].map(lambda x: x in allowed)
+        trips["location_allowed"] = trips["dwell_location"].map(lambda x: x in allowed)
```
BainanXia: Is it equivalent to `trips["location_allowed"] = trips["dwell_location"].isin(allowed)`? Same comment below.

prereise/gather/demanddata/transportation_electrification/immediate.py
```diff
@@ -136,6 +139,11 @@ def immediate_charging(
     # updates the weekend and weekday values in the nhts data
     trips = data_helper.update_if_weekend(trips)
 
+    if power > 19.2:
+        charging_efficiency = 0.95
+    else:
+        charging_efficiency = 0.9
```
BainanXia, line 145: Not sure whether we should factor this out given it is used in both immediate charging and smart charging, also it would be great if references of these numbers (from assumptions or from existing studies?) can be added somewhere with the module.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -284,17 +289,31 @@ def smart_charging(
...
+                load_demand[i] + model_year_profile[i]
+                for i in range(day_iter * 24, ((day_iter + 1) * 24) + 48)
+            ]
+
         else:
             adjusted_load = [
                 load_demand[i] + model_year_profile[i]
-                for i in range(day_iter * 24, day_iter * 24 + 48)
+                for i in range(day_iter * 24, day_iter * 24 + 72)
             ]
```
BainanXia: I think we should be able to combine these branches together by following:
```python
day_num
for day_iter in range(day_num):
    adjusted_load = [
        load_demand[i] + model_year_profile[i]
        for i in range(day_iter * 24, day_iter * 24 + min(day_num - day_iter - 1, 3) * 24
    ]
    if 3 - day_num + day_iter > 0:
        adjusted_load += [load_demand[i] + model_year_profile[i] for i in range(24*(3 - day_num + day_iter))]
```

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -470,10 +490,26 @@ def smart_charging(
...
+            )
+            model_year_profile[:24] += (
+                outputelectricload[48:72]
+                / (daily_vmt_total[day_iter] * 1000)
+                * emfacvmt
+            )
+
+        else:
+            # MW
+            model_year_profile[day_iter * 24 : day_iter * 24 + 72] += (
```
BainanXia, line 500: Not sure whether we can simplify the logic in a similar way above.

BainanXia (commented):
This looks good. Changing the maximum dwell period for LDV to 72 hours to make it compatible with HDV makes sense, which I believe won't change the behavior of LDV because no LDV trips in the dataset has dwell period greater than 48 hours. Not sure whether it's worth the effort to make the dwell period a variable (maybe still a multiple of 24) so that if there are more HDV data coming in, corner cases will be handled automatically. Maybe that's overkill though.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: Yeah, I’ve been thinking about what a generalized function for an arbitrary dwell window would look like, though it would likely be later…maybe also some type of detection of how big the dwell window should be.

prereise/gather/demanddata/transportation_electrification/immediate.py
dmuldrew: For now I’d leave it as is, as I’ve been using the diffs between the LDV and HDV immediate/smart charging files. There will be another PR that will just be moving code blocks around…

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: Tried this out in the following branch:
https://github.com/Breakthrough-Energy/PreREISE/tree/load_branching_simplification
however I got the following test error:
Screen Shot 2022-10-05 at 12 16 51 PM

BainanXia: Ah, my bad. I copied wrong line, now the test passes:
```python
    day_num = len(input_day)
    for day_iter in range(day_num):

        adjusted_load = [
            load_demand[i] + model_year_profile[i]
            for i in range(
                day_iter * 24, (day_iter + 1) * 24 + min(day_num - day_iter - 1, 2) * 24
            )
        ]

        if 3 - day_num + day_iter > 0:
            adjusted_load += [
                load_demand[i] + model_year_profile[i]
                for i in range(24 * (3 - day_num + day_iter))
            ]
```

BainanXia (approved):
Given we decided to save some potential refactors into future PRs, this is good to go.

PR #326, opened 2022-12-11
Update dependencies

rouille:

### Purpose
Update `powersimdata` following release to `0.5.5` in **Pipfile** and **requirements.txt**. A new **Pipfile.lock** is also generated

### What the code is doing
N/A

### Testing
Existing unit test. One test failed because the newly `pandas` does not allow to use `set` for columns.

### Where to look
One test has been modified to avoid an error raised by `pandas`

### Usage Example/Visuals
N/A

### Time estimate
2min


BainanXia (approved):
🙏

Issue #317, opened 2022-10-12
Split urban areas crossing multiple states proportional to the population within each state


BainanXia:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
In the current implementation, UA VMT is not split properly when such area lies across state boundaries. The proposed implementation is to split such urban area proportional to the population within each state.

### Describe your proposed implementation, if applicable
A new data source has been found from census data set that gives the population (and land area, etc.) of UA between states.
https://www.census.gov/programs-surveys/geography/technical-documentation/records-layout/2010-urban-lists-record-layout.html 
Scroll down one can find a file titled “ua_st_list_all.xls”.

PR #327, opened 2023-01-09
feat: state demand profile interface function

dmuldrew:

### Purpose

Create an interface function to simplify generating profiles by state. Also, finished merging the LDV/HDV smart charging functions into one common function.

### What the code is doing

This function generates a dictionary of vehicle miles traveled (`bev_vmt`) for a specific geographic regions within a given state. It then outputs a demand profile for each geographic region. The state function input is also mapped to a census region which loads the appropriate vehicle census data.

### Testing

manual testing so far, would like to eventually add an automated integration test

### Time estimate

~30min


prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    model_year,
+    state,
+    load_demand,
+    power=6.6,
+    location_strategy=2,
+    trip_strategy=1,
+):
+    """Generate Battery Electric Vehicle (BEV) profiles
+
+    :param str vehicle_trip_data_filepath: collected trip data from external sources representing driving patterns
```
BainanXia: Maybe we mean `file path of collected trip data ...` here?

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    :param str charging_strategy: establishes whether charging happens immediately or via a smart charging algorithm;
+                current options are "immediate" and "smart"
```
BainanXia: Redundant new line. Also, maybe we can be more specific here `charge immediately between trips or optimize based on external signals, i.e. smart charging`.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    :param str veh_type: vehicle category (LDV, LDT, MDV, HDV)
```
BainanXia: Given this function might be user-facing, let's specify what these terms short for as well, 
`LDV: light duty vehicle, LDT: light duty truck, MDV: medium duty vehicle, HDV: heavy duty vehicle`

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    :param int veh_range: 100, 200, or 300, represents how far vehicle can travel on
+        single charge.
```
BainanXia: Let's include the unit as `miles` here.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    :param int model_year: year that is being modelled/projected to, 2017, 2030, 2040,
+        2050.
```
BainanXia: Maybe we should change it into `projected_year` or `year_projection` to eliminate the confusion since we are talking about vehicles here?

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    :param str state: US state abbreviation
+    :param np.array load_demand: initial load demand (MW for each hour)
```
BainanXia: `np.array` -> `np.ndnarray`. We will need to be more specific for this parameter, since it is only used if `smart charing` is chosen as the charging strategy and it is the external signal that the algorithm is optimizing against, hence it is not necessarily to be demand. Speaking of that, maybe we could rename it as `external_signal`?

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    :param int power: (optional) charger power, EVSE kW; default value: 6.6 kW;
+    :param int location_strategy: (optional) where the vehicle can charge-1, 2, 3, 4, or 5;
+        1-home only, 2-home and work related, 3-anywhere if possibile,
```
BainanXia, line 48: What is the difference between `work related` and `work`?

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+        4-home and school only, 5-home and work and school.
+        default value: 2
+    :param int trip_strategy: (optional) determine to charge after any trip (1) or only after the
+        last trip (2); default value: 1
+    :return: (*numpy.ndarray*) -- yearly charging profiles for all urban areas and the rural area in each state (MW for each hour)
```
BainanXia: Should we return a numpy array or a dataframe? If we mean the former, how multiple time series profiles are represented in one numpy array?

BainanXia: Update: it seems we are returning a dictionary with keys being geographical areas within the state and values being the time series charging profile.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+        veh_type,
+        veh_range,
+        state,
+        filepath=os.path.join(
+            const.data_folder_path,
+            "regional_scaling_factors",
+            "Regional_scaling_factors_RA_",
+        ),
+    )
+    geographic_area_bev_vmt.update({state: rural_bev_vmt})
```
BainanXia: Should we add a suffix to this key to specify "the rural area of the state", say `f"{state}_rural"`?

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+                kwhmi=kwhmi,
+                power=power,
+                location_strategy=location_strategy,
+                veh_type=veh_type,
+                filepath=vehicle_trip_data_filepath,
+                daily_values=daily_values,
+                load_demand=load_demand,
+                bev_vmt=bev_vmt,
+                trip_strategy=trip_strategy,
+            )
```
BainanXia, line 155: Are we still trying to figure out whether to call `adjust_bev` for smart charing?

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,146 @@
...
+    # scaling factors for listed urban areas
+    for urban_area in state_urban_areas:
+        urban_bev_vmt = load_urbanized_scaling_factor(
+            model_year=model_year,
+            veh_type=veh_type,
+            veh_range=veh_range,
+            urbanized_area=urban_area,
+            state=state,
+            filepath=urban_scaling_filepath,
+        )
```
BainanXia: I guess the logic that splits large urban areas that crossing state borders will kick in here, in the downstream function calls once implemented, right?

BainanXia (commented):
Let's be more specific on the doc strings of input parameters given this will be a user-facing function, i.e. which parameter is used for which charing strategy; if a parameter is only used for smart charing, then it should be optional; speaking of that, it would be great if we can add some checks on different combinations of the input parameters based on different charing strategies.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
dmuldrew: Yep, Dan and I decided on the dictionary yesterday and forgot to update this

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: No, that is no longer needed.  The `bev_vmt` scaling is included directly in the smart charging algorithm.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: I believe that logic will actually be part of the pre-algorithm data processing step that we'll finish via #317.  Each state will have a separate entry to include its portion of those urban areas.
Example: from the `ua_st_list_all.xls` file, Dubuque IA--IL has population of 64,767 in IA and 3,051 in IL.  
Currently we've used the `ua_list_all.xls` file, which has Dubuque IA--IL with a population of 67,818.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: Nothing as far as I know.  I'll double-check with Kate.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: Ooo, I like `external_signal`.  Good idea.
And indeed, this is only used for `smart` charging.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: Ha!  Good call.  My vote is `projection_year`

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: Given the discussion about urban areas being in more than one state, I think we should consider adding the `state` name to all of these.
E.g.
`f"{state}_rural"`
`f"{state}_urban_area"`

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,158 @@
...
+    :param str charging_strategy: establishes whether charging happens immediately ("immediate")
+         or optimize based on external signals, i.e. smart charging ("smart")
+    :param str veh_type: vehicle category: LDV: light duty vehicle, LDT: light duty truck,
+        MDV: medium duty vehicle, HDV: heavy duty vehicle
+    :param int veh_range: 100, 200, or 300, represents how far vehicle can travel on
+        single charge in miles.
+    :param int projection_year: year that is being modelled/projected to, 2017, 2030, 2040,
+        2050.
+    :param str state: US state abbreviation
+    :param np.ndnarray (optional) external_signal: initial load demand (MW for each hour)
```
BainanXia: `np.ndnarray` -> `numpy.ndarray`

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
```diff
@@ -0,0 +1,158 @@
...
+    :param int power: (optional) charger power, EVSE kW; default value: 6.6 kW;
+    :param int location_strategy: (optional) where the vehicle can charge-1, 2, 3, 4, or 5;
+        1-home only, 2-home and work related, 3-anywhere if possibile,
+        4-home and school only, 5-home and work and school.
+        default value: 2
+    :param int trip_strategy: (optional) determine to charge after any trip (1) or only after the
+        last trip (2); default value: 1
+    :return: (*pandas.DataFrame*) -- yearly charging profiles for all urban areas and the rural area
+        in each state (MW for each hour)
```
BainanXia, line 54: This function is supposed to return yearly charging profiles for all urban areas and the rural area of the specified `state`. I guess we will need to loop through all the states in the higher level function call to get profiles for each state, no?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -51,16 +71,15 @@ def smart_charging(
     # load NHTS data from function
     if veh_type.lower() == "ldv":
         newdata = data_helper.remove_ldt(data_helper.load_data(census_region, filepath))
+        # updates the weekend and weekday values in the nhts data
+        newdata = data_helper.update_if_weekend(newdata)
```
BainanXia, line 75: Do we need this for `ldt` as well?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -76,12 +95,20 @@ def smart_charging(
     newdata["trip_number"] = newdata.groupby("vehicle_number").cumcount() + 1
 
     input_day = data_helper.get_input_day(data_helper.get_model_year_dti(model_year))
+    load_demand = -min(load_demand) + load_demand
```
BainanXia: Let's use `external_signal` across the board.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: Yes indeed.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
danlivengood: Yes we do

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -11,18 +11,38 @@
...
+
+
+def hdv_use_all_data_rows(x, y):
+    """Helper function to select weekday/weekend data rows
+
+    :param int x: data weekday/weekend value
+    :param int y: model year weekday/weekend value
+    :return: (*bool*) -- always returns true to use all data rows
+    """
+    return True
```
BainanXia: I don't quite follow what we are doing here.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -76,12 +95,20 @@ def smart_charging(
...
     model_year_profile = np.zeros(24 * len(input_day))
-    data_day = data_helper.get_data_day(newdata)
-
-    daily_vmt_total = data_helper.get_total_daily_vmt(newdata, input_day, daily_values)
+    if veh_type.lower() == "ldv":
+        data_day = data_helper.get_data_day(newdata)
+        daily_vmt_total = data_helper.get_total_daily_vmt(
+            newdata, input_day, daily_values
+        )
+    elif veh_type.lower() == "mdv" or veh_type.lower() == "hdv":
```
BainanXia: We could do `veh_type.lower() in {'mdv', 'hdv'}`.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
danlivengood: This looks similar to the `get_data_day` and/or `get_input_day' functions in `data_helper.py`.  Could one of those be used instead of creating a new function?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -11,18 +11,38 @@
 )
 
 
+def ldv_weekend_weekend_check(x, y):
```
BainanXia: `weekend_weekend` -> `weekday_weekend`?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
danlivengood: Bainan - the context is that the LDV and LDT data from NHTS has a differentiation between weekday and weekend trip patterns.  The MDV and HDV data does not have that differentiation

prereise/gather/demanddata/transportation_electrification/tests/test_smart_charging_integration.py
```diff
@@ -86,7 +85,7 @@ def test_smart_charging_hdv():
             const.test_folder_path,
             "hdv_test_data.csv",
         ),
-        initial_load=load_demand,
+        load_demand=load_demand,
```
BainanXia: Same comment above.

prereise/gather/demanddata/transportation_electrification/tests/test_smart_charging_integration.py
```diff
@@ -136,7 +135,7 @@ def test_smart_charging_mdv():
             const.test_folder_path,
             "mdv_test_data.csv",
         ),
-        initial_load=load_demand,
+        load_demand=load_demand,
```
BainanXia: Same comment above.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
```diff
@@ -128,157 +162,144 @@ def smart_charging(
...
-            if data_day[i] == input_day[day_iter]:
-
-                # only home based trips
-                if (
-                    newdata.iloc[i, newdata.columns.get_loc("why_from")]
-                    * newdata.iloc[
-                        i + total_trips - 1, newdata.columns.get_loc("dwell_location")
-                    ]
-                    == 1
-                ):
```
BainanXia, line 140: We no longer need this logic or it has been moved somewhere else?

danlivengood: The `only home based trips` if statement is a filtering process of the input data that we'll be moving to an early point in the code workflow.  It's also one of the sources of the smart charging scaling factor bug.

BainanXia: Good to know!

prereise/gather/demanddata/transportation_electrification/smart_charging.py
BainanXia: I see. Nonetheless, having a function simply returning `x == y` and another one always returning `True` seems to be redundant. I would suggest to use `lambda` if necessary.

prereise/gather/demanddata/transportation_electrification/smart_charging.py
BainanXia: @dmuldrew Given this missing logic, the integration test still passes. Shall we update the test case to include at least one weekend to cover this case?

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: It's an intermediate refactoring step before changing the looping structure...I wanted to use lambda functions instead of function definitions but the linter complained:
https://www.flake8rules.com/rules/E731.html

Ideally I'm going to get rid of the double loop over all the data every day, and switch out weekend and weekday dataframes for the light vehicle case.

The function `use_data_row` determines if a data row is used for a particular day:
``` python
if use_data_row(data_day[i], input_day[day_iter]):
```
and is set based on whether `veh_type` is "ldv"/"ldt" or "mdv"/"hdv"
``` python
  if veh_type.lower() == "ldv" or veh_type.lower() == "ldt":
      use_data_row = ldv_weekend_weekend_check
  elif veh_type.lower() == "mdv" or veh_type.lower() == "hdv":
      use_data_row = hdv_use_all_data_rows
```

prereise/gather/demanddata/transportation_electrification/smart_charging.py
dmuldrew: Maybe, I think the main issue here is we don't have a test case for `ldt`...

prereise/gather/demanddata/transportation_electrification/smart_charging.py
BainanXia: @dmuldrew So eventually we will get rid of this, right?

dmuldrew: Yep, I'm definitely not happy with it long-term

BainanXia: Cool. Let's leave it as it is for now then.

prereise/gather/demanddata/transportation_electrification/generate_BEV_vehicle_profiles.py
danlivengood: Aha!  Reply from Kate: `It is currently labeled work-related, because there are a couple "whyfrom" and "whyto" answers that have work in the description.`  For instance, `04=Work-related meeting / trip`.  This is a nuance that I don't think is worth parsing in the data nor the code.  I'm ok with leaving it as `work related` but also open to changing it.

dmuldrew:
Looks like 
prereise/gather/flexibilitydata/doe/tests/test_bus_data.py::test_get_bus_fips
is failing in PR test workflows

prereise/gather/demanddata/transportation_electrification/smart_charging.py
BainanXia: Should we add it to the `ldt` branch below then?

BainanXia (approved):
Thanks!

PR #345, opened 2023-02-13
Dmuldrew/performance testing

dmuldrew:

### Purpose

1. Fix a bug where a single urban area was not compatible with iteration loop
2. Fix immediate demand scaling for leap years
3. Divide immediate demand by 1000 to get MW
4. Optimize immediate charging code to only calculate demand shape once per state
5. Create script to output US immediate demand pickle files on the dev server

### Testing

manual testing

### Time estimate

~15min


rouille:
In the `generate_BEV_vehicle_profiles` module, I would raise a `ValueError` within a `else` statement after the `elif`:
* `raise ValueError("Invalid veh_type. Choose from ldv | ldt | mdv | hdv")`
* `raise ValueError("Invalid charging_strategy. Choose from immediate | smart")`

prereise/gather/demanddata/transportation_electrification/generate_state_immediate_files.py
```diff
@@ -0,0 +1,53 @@
...
+            power=6.6,
+            location_strategy=2,
+            trip_strategy=1,
+        )
+
+    print(state)
+
+    write_state_demand_files(state_demand_df, state)
+
+    print()
```
rouille: Should the 3 lines above be moved within the `for` loop? Or you just want to print out and write data for the last sate in `abv2state`? Also , can we do `print(f"{state}\n")` in the first `print` and remove the last one?

dmuldrew: Yep, fixed this already, will push the changes

PR #355, opened 2023-04-25
Update dependencies

rouille:

### Purpose
Update dependencies as new packages were introduced with the building electrification project. It was tricky. Also, docstrings have been cleaned up and missing ***\_\_init\_\_.py*** files have been added.

### What the code is doing
N/A

### Testing
Current unit test

### Where to look
* Impossible to generate a ***Pipefile.lock*** file  that work with all 3 Python versions we support. The ***Pipefile.lock*** file in this PR has been generated with Python 3.10.4. For this reason, dependencies for testing are installed using the ***requirements.txt*** file (see ***tox.ini***).@jenhagg, do we want to keep the ***Pipefile.lock*** in version control? [pipenv](https://pipenv-fork.readthedocs.io/en/latest/basics.html#general-recommendations-version-control) recommends not to if multiple version of python are targeted.
* Skip test in `prereise.gather.flexibilitydata.doe.tests.test_bus_data` that most of the time fails to connect to DOE server
* Add missing ***\_\_init\_\_.py*** files

### Usage Example/Visuals
N/A

### Time estimate
15min

### DIFF
--- a/prereise/gather/demanddata/bldg_electrification/floor_area_avg_slopes.py
+++ b/prereise/gather/demanddata/bldg_electrification/floor_area_avg_slopes.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 
+import geopandas  # noqa: F401
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
--- a/prereise/gather/demanddata/bldg_electrification/load_projection.py
+++ b/prereise/gather/demanddata/bldg_electrification/load_projection.py
@@ -18,6 +18,9 @@ from prereise.gather.demanddata.bldg_electrification.helper import (
     read_shapefile,
     zone_shp_overlay,
 )
+from prereise.gather.demanddata.bldg_electrification.load_projection_scenario import (  # noqa: F401
+    LoadProjectionScenario,
+)
 from prereise.gather.demanddata.bldg_electrification.zone_profile_generator import (
     zonal_data,
 )
@@ -32,10 +35,9 @@ def temp_to_energy(temp_series, hourly_fits_df, db_wb_fit, base_scen, hp_heat_co
         coefficients for electricity use equations.
     :param pandas.DataFrame db_wb_fit: least-square estimators of the linear
         relationship between WBT and DBT
-    :param load_projection_scenario.LoadProjectionScenario base_scen: reference
-        scenario instance
-    :param pandas.DataFrame hp_heat_cop: heat pump COP against DBT with
-        a 0.1 degree C interval
+    :param LoadProjectionScenario base_scen: reference scenario instance
+    :param pandas.DataFrame hp_heat_cop: heat pump COP against DBT with a 0.1 degree C
+        interval
     :return: (*list*) -- energy for baseload, heat pump heating, resistance heating,
         and cooling of certain hour
     """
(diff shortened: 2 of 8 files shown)


prereise/gather/demanddata/bldg_electrification/floor_area_avg_slopes.py
```diff
@@ -1,6 +1,7 @@
 import os
 import shutil
 
+import geopandas  # noqa: F401
```
rouille, line 4: Needed for `sphinx`

prereise/gather/demanddata/bldg_electrification/load_projection.py
```diff
@@ -18,6 +18,9 @@
     read_shapefile,
     zone_shp_overlay,
 )
+from prereise.gather.demanddata.bldg_electrification.load_projection_scenario import (  # noqa: F401
+    LoadProjectionScenario,
+)
```
rouille, line 23: Needed for `sphinx`

requirements.txt
```diff
@@ -21,4 +21,8 @@ pygrib~=2.1.4; sys_platform != "win32"
 geopandas
 rtree~=1.0
 shapely~=1.8
-geopy
+geopy~=2.0
```
rouille, line 24: To mirror what is in the ***Pipfile***

jenhagg:
I would keep Pipfile.lock versioned. Maybe also add the python version to Pipfile (see [docs](https://pipenv-fork.readthedocs.io/en/latest/basics.html#specifying-versions-of-python)), as below. Btw, best source I found for `python_full_version` is in [PEP 508](https://peps.python.org/pep-0508).

```
[requires]
python_full_version= "3.10.4"
```

I think testing multiple python versions with requirements.txt is fine. Not 100% reproducible, but don't expect too much noise.

Thinking out loud a bit here, but maybe helpful:

The context for Pipfile.lock is that pipenv and lock file is designed for "deployment environments" meaning an application that will always use exactly one python version. We kinda have a hybrid situation where we publish packages which are libraries - meant to be usable in a variable number of python environments (could be multiple OSes, python versions, etc. whatever we specify is supported). And pipenv says it's not [necessarily] meant for packages. But also have users who work with the packages directly in a local environment, but not e.g. typically included in another application, the way libraries "normally" are (idk, just in my experience). I think this is somewhat a result of splitting up repositories and simplifying internal dev workflow, e.g. by setting powersimdata as a dependency in postreise. But there is also value in making the functionality accessible to anyone generally. 

So anyway, since we more or less want both a reproducible environment and to support package distribution, it maybe makes sense to keep both Pipfile.lock and requirements.txt, respectively. And the CI checks will catch anything that breaks downstream users (fine print: as long as it's covered by tests/linting/etc).

jenhagg (approved):

rouille:
> I would keep Pipfile.lock versioned. Maybe also add the python version to Pipfile (see [docs](https://pipenv-fork.readthedocs.io/en/latest/basics.html#specifying-versions-of-python)), as below. Btw, best source I found for `python_full_version` is in [PEP 508](https://peps.python.org/pep-0508).
> 
> ```
> [requires]
> python_full_version= "3.10.4"
> ```
> 
> I think testing multiple python versions with requirements.txt is fine. Not 100% reproducible, but don't expect too much noise.
> 
> Thinking out loud a bit here, but maybe helpful:
> 
> The context for Pipfile.lock is that pipenv and lock file is designed for "deployment environments" meaning an application that will always use exactly one python version. We kinda have a hybrid situation where we publish packages which are libraries - meant to be usable in a variable number of python environments (could be multiple OSes, python versions, etc. whatever we specify is supported). And pipenv says it's not [necessarily] meant for packages. But also have users who work with the packages directly in a local environment, but not e.g. typically included in another application, the way libraries "normally" are (idk, just in my experience). I think this is somewhat a result of splitting up repositories and simplifying internal dev workflow, e.g. by setting powersimdata as a dependency in postreise. But there is also value in making the functionality accessible to anyone generally.
> 
> So anyway, since we more or less want both a reproducible environment and to support package distribution, it maybe makes sense to keep both Pipfile.lock and requirements.txt, respectively. And the CI checks will catch anything that breaks downstream users (fine print: as long as it's covered by tests/linting/etc).

Sounds good. I have updated the ***Pipfile*** as suggested.

PR #356, opened 2023-05-01
Fix call to function

rouille:

### Purpose

### What the code is doing
Pass a single argument, the path to the folder containing the network data, to the `get_bus_pos` function. Some code responsible for reading files was refactored.

### Testing
An error was thrown when generating the documentation locally with `sphinx`. With the change, it is fixed.

### Where to look
The `get_bus_pos` located in the `prereise.gather.flexibilitydata.doe.bus_data` module takes one argument, the path to a folder containing the ***sub.csv*** and ***bus2sub.csv*** file. The `bus_example` script located in `prereise.gather.flexibilitydata.doe.examples.bus_example` module call the function with two arguments, the path to a folder and a filename.

@BainanXia and @lanesmith, do you know why the `get_bus_pos` reads network data instead of using a `Grid` object? A `Grid` object could be the argument of the `get_bus_pos` function.

### Usage Example/Visuals
N/A

### Time estimate
5min

### DIFF
--- a/prereise/gather/flexibilitydata/doe/bus_data.py
+++ b/prereise/gather/flexibilitydata/doe/bus_data.py
@@ -4,11 +4,11 @@ import pickle as pkl
 import re
 import time
 
-import geopy
 import numpy as np
 import pandas as pd
 import requests
 from geopy.extra.rate_limiter import RateLimiter
+from geopy.geocoders import Nominatim
 from powersimdata.utility.distance import find_closest_neighbor
 from tqdm import tqdm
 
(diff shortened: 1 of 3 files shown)


lanesmith (approved):
Looks good to me!

As for why `get_bus_pos` reads in network data instead of using a `Grid` object, I don't remember. As you said, I don't think there is a good reason that a `Grid` object shouldn't be used.

BainanXia:
If I recall correctly, it took a specific version of the network and some corner cases were resolved manually at the early stage. Later we automated the process which should no longer a problem.

rouille:
> If I recall correctly, it took a specific version of the network and some corner cases were resolved manually at the early stage. Later we automated the process which should no longer a problem.

I will refactor the function in a follow up PR then.

